PR #28, opened 2015-01-02
One Makefile to rule them all.

dkorolev:


cerealize/KnowSheetDir.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+#
+# Goes up the directory tree until "KnowSheet/" sub-directory is found.
+# Prints its path.
+
+dir=$(cd $(dirname $0) && pwd)
+subdir=KnowSheet
+
+while [ ! -d ${dir}/${subdir} ] ; do
+  [ "$dir" == "/" ] && dir="" && break
```
kt97679: dir="" not necessary here since you check if directory exists after loop

dkorolev: Thanks, fixed.

dkorolev:
@deathbaba PTAL.

deathbaba:
Why do you need all these dir print scripts? Can't you do it without copy-paste?

dkorolev:
@deathbaba Please suggest a solution without copy-paste. I'm all for it, but git/GitHub do not support symlinks, and modern build systems are far from perfect.

deathbaba:
git/github support symlinks.

dkorolev:
Just checked, indeed, thanks! Hang on for a few mins for an update to this pull request ...

dkorolev:
@deathbaba : Please take another look.

PR #29, opened 2015-01-03
Covered MakeScope* with tests + minor tweaks.

dkorolev:


util/make_scope_guard.h
```diff
@@ -30,8 +30,15 @@ SOFTWARE.
 
 namespace bricks {
 
-template <typename POINTER, typename DELETER>
-std::unique_ptr<POINTER, DELETER> MakePointerScopeGuard(POINTER* x, DELETER t) {
+struct CPlusPlusStandardDeleter {
```
deathbaba: http://en.cppreference.com/w/cpp/memory/default_delete is already there, why not use it?

dkorolev:
Thanks, fixed, @deathbaba PTAL.

PR #36, opened 2015-01-07
URL Parsing.

dkorolev:


dkorolev:
@deathbaba : Please take a look.
@sompylasar : Could you please check the encode/decode logic and the terminology of URL/URI/path/route?

Thanks!
Dima

net/api/impl/posix.h
```diff
@@ -87,7 +89,7 @@ class HTTPClientPOSIX final {
       // connection.SendEOF();
       message_.reset(new HTTPRedirectableReceivedMessage(connection));
       // TODO(dkorolev): Rename `URL()`, it's only called so now because of HTTP request/response format.
```
deathbaba: old todo?

net/api/impl/posix.h
```diff
@@ -87,7 +89,7 @@ class HTTPClientPOSIX final {
...
-      const int response_code_as_int = atoi(message_->URL().c_str());
+      const int response_code_as_int = atoi(message_->Path().c_str());
```
deathbaba: How does code from Path work?

net/api/test.cc
```diff
@@ -135,13 +136,12 @@ class UseLocalHTTPTestServer {
       HTTPServerConnection connection(socket.Accept());
       const auto& message = connection.Message();
       const string method = message.Method();
-      const string url = message.URL();
+      const URL url = URL(message.Path());
       if (method == "GET") {
-        if (url == "/get") {
-          connection.SendHTTPResponse("DIMA");
-        } else if (url == "/drip?numbytes=7") {
+        if (url.path == "/drip") {
```
deathbaba: path is correct name for a part after host and before query params

net/url/url.h
```diff
@@ -115,20 +118,25 @@ struct URL {
...
-    if (!protocol.empty()) {
-      os << protocol << "://";
-    }
-    os << host;
-    if (port != DefaultPortForProtocol(protocol)) {
-      os << ':' << port;
+    if (!host.empty()) {
+      std::ostringstream os;
+      if (!protocol.empty()) {
+        os << protocol << "://";
```
deathbaba: may be scheme instead of protocol?

net/url/url.h
```diff
@@ -115,20 +118,25 @@ struct URL {
...
+      }
+      os << host;
```
deathbaba: wanna be cool? you do not parse user:password part in scheme://user:password@host:port/path?query

net/url/url.h
```diff
@@ -139,6 +147,109 @@ struct URL {
   static std::string DefaultProtocolForPort(int port) { return port == 80 ? "http" : ""; }
```
deathbaba: you can put https port here

net/api/impl/posix.h
dkorolev: Yes. Renamed ```URL()``` into ```Path()``` in its message.

net/api/impl/posix.h
dkorolev: It just extracts the second part of the first non-empty HTTP line.

Where the request contains ```GET /path HTTP/1.1```, making ```"/path"``` a path, the response contains ```HTTP/1.1 200 OK```. Thus, the ```200``` part in the response is located exactly same way as the ```"/path"``` part is located in the request, thus variable/method name reuse.

I'll clean it up one day.

net/url/url.h
dkorolev: Not now, please :-)

We'll do HTTP via nginx and auth tokens via headers first.

net/url/url.h
dkorolev: Done.

dkorolev:
@deathbaba : Thanks, done everything (but the ```username@password``` part), please take another loop.

net/api/impl/posix.h
```diff
@@ -86,8 +88,8 @@ class HTTPClientPOSIX final {
       // Don't uncomment the next line!
       // connection.SendEOF();
       message_.reset(new HTTPRedirectableReceivedMessage(connection));
-      // TODO(dkorolev): Rename `URL()`, it's only called so now because of HTTP request/response format.
-      const int response_code_as_int = atoi(message_->URL().c_str());
+      // TODO(dkorolev): Rename `Path()`, it's only called so now because of HTTP request/response format.
+      const int response_code_as_int = atoi(message_->Path().c_str());
```
sompylasar: Please add a comment with an example of this similarity (`GET /path HTTP/1.1` and `HTTP/1.1 200 OK`). BTW, the response may omit the text description of the status code, so you'd better make another method for parsing out the response code.

sompylasar: What would be the behavior of this code in case of a malformed response?

dkorolev: Done, thanks for reminding me to do so. It's the numerical code that is parsed by this logic, its string representation is being ignored.

net/api/impl/posix.h
```diff
@@ -86,8 +88,8 @@ class HTTPClientPOSIX final {
...
       response_code_ = static_cast<HTTPResponseCode>(response_code_as_int);
       if (response_code_as_int >= 300 && response_code_as_int <= 399 && !message_->location.empty()) {
```
sompylasar: The value of the `Location` header would more logically be a part of some `headers` hashmap, not a property of the message (though it makes things more complex).

net/api/impl/posix.h
dkorolev: It would store that code as ```int```, cast to ```enum class
HTTPResponseCode```, return as ```"Unknown Code"``` when queries as string.
ᐧ

Thanks,
Dima

net/api/impl/posix.h
dkorolev: It's parsed in ```OnHeader()``` of ```struct HTTPRedirectHelper :
HTTPDefaultHelper```.
ᐧ

Thanks,
Dima

net/api/test.cc
```diff
@@ -135,13 +136,12 @@ class UseLocalHTTPTestServer {
...
+          const size_t numbytes = atoi(url["numbytes"].c_str());
```
sompylasar: The indexer on the `url` looks nice at first sight, but it would be more logical from an API standpoint to put the parsed query parameters into a separate `query` property, e.g. `url.query["numbytes"]`.

net/api/impl/posix.h
dkorolev: (It would actually make things easier, but, for some strange reason, I thought that parsing all the headers into a hash-map would be undesirable for a lightweigh server, especially given that there already is a mechanism to only expose the ones we need.)

net/api/test.cc
dkorolev: Very good point! Let me do this once I am down to the ```express```-style server implemenation.

(It would have ```query``` parameters and the ```URL``` ones separately, ex. ```/user/:uid/followers?time_filter=1day``` should set both some ```.query["time_filter"]``` and ```.params["uid"]```.)

sompylasar: Oh, I thought this PR's aimed towards the `express`-like API, too.

dkorolev: It's going to get much cleaner. I have a plan :)

net/http/impl/server.h
```diff
@@ -47,9 +47,9 @@ typedef std::vector<std::pair<std::string, std::string>> HTTPHeadersType;
 namespace {
 
 const char kCRLF[] = "\r\n";
-const size_t kCRLFLength = CompileTimeStringLength(kCRLF);
+const size_t kCRLFLength = strings::CompileTimeStringLength(kCRLF);
 const char kHeaderKeyValueSeparator[] = ": ";
```
sompylasar: [According to the specs](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), the space after the header separator is optional, and that space may be any linear whitespace character, not only `0x20` space:
> The field value MAY be preceded by any amount of LWS, though a single SP is preferred.
> ...
> `message-header = field-name ":" [ field-value ]`

net/url/test.cc
```diff
@@ -35,37 +35,37 @@ TEST(URLTest, SmokeTest) {
...
   EXPECT_EQ(27960, u.port);
 
   u = URL("meh://www.google.com:27960/bazinga");
   EXPECT_EQ("www.google.com", u.host);
   EXPECT_EQ("/bazinga", u.path);
-  EXPECT_EQ("meh", u.protocol);
+  EXPECT_EQ("meh", u.scheme);
   EXPECT_EQ(27960, u.port);
 
   u = URL("localhost:/test");
```
sompylasar: You should also test for a terminating slash without the following path, like: `URL("localhost/")`.

sompylasar: BTW, does your `URL` support or is going to support browser-side-like schema-less root-pointing syntax: `//host/path`?

dkorolev: Yes, this one is taken care of, check out the unit test for ```url.h```.

Let me do the whitespaces part between ```HTTP/1.1``` and ```200``` real quick ...

net/url/test.cc
```diff
@@ -105,11 +109,73 @@ TEST(URLTest, RedirectPreservesProtocolHostAndPortTest) {
             URL("blah://new_host:6000/foo", URL("meh://localhost:5000")).ComposeURL());
 }
 
+TEST(URLTest, ExtractsURLParameters) {
+  {
+    URL u("www.google.com");
+    EXPECT_EQ("", u.fragment);
+    EXPECT_EQ("", u["key"]);
+    EXPECT_EQ("default_value", u("key", "default_value"));
```
sompylasar: Re: `u("key", "default_value")`
Nice but frustrating to call a URL as a function.

dkorolev: I don't like it that much either. Any good idea on how to make it read better?

net/url/test.cc
```diff
@@ -105,11 +109,73 @@ TEST(URLTest, RedirectPreservesProtocolHostAndPortTest) {
...
+  }
```
sompylasar: Why no `u.ComposeURL()` line here as in the tests below?

net/url/test.cc
```diff
@@ -105,11 +109,73 @@ TEST(URLTest, RedirectPreservesProtocolHostAndPortTest) {
...
+  {
+    URL u("www.google.com/a#fragment");
+    EXPECT_EQ("fragment", u.fragment);
+    EXPECT_EQ("", u["key"]);
+    EXPECT_EQ("default_value", u("key", "default_value"));
+    EXPECT_EQ("http://www.google.com/a#fragment", u.ComposeURL());
+  }
+  {
+    URL u("www.google.com/b#fragment#foo");
```
sompylasar: Please add a test for `?` and `&` inside the fragment identifier, too.

net/url/test.cc
```diff
@@ -105,11 +109,73 @@ TEST(URLTest, RedirectPreservesProtocolHostAndPortTest) {
...
+}
+
 TEST(URLTest, EmptyURLException) {
-  // Empty URL or host should throw.
+  // Empty URL should throw.
   ASSERT_THROW(URL(""), EmptyURLException);
-  ASSERT_THROW(URL("http://"), EmptyURLHostException);
-  ASSERT_THROW(URL("http:///foo"), EmptyURLHostException);
 
   // Empty host is allowed in local links.
```
sompylasar: These links are `relative`, not `local`.

net/url/url.h
```diff
@@ -26,65 +26,68 @@ SOFTWARE.
...
 // URL manages the mapping between the string and parsed representations of the URL. It manages:
 //
-// * host     (string)
-// * path     (string, defaults to "/", never empty).
-// * protocol (defaults to "http", empty only if set explicitly in constructor).
-// * port (defaults to the default port for supported protocols).
+// * host    (string)
+// * path    (string, defaults to "/", never empty).
+// * scheme  (defaults to "http", empty only if set explicitly in constructor).
+// * port    (defaults to the default port for supported schemes).
```
sompylasar: ...and if the scheme is not supported, is the default zero or what?

net/url/url.h
```diff
@@ -26,65 +26,68 @@ SOFTWARE.
...
-      protocol = url.substr(0, i);
-      offset_past_protocol = i + 3;
+      scheme = url.substr(0, i);
+      offset_past_scheme = i + 3;
     }
 
-    const size_t colon = url.find(':', offset_past_protocol);
-    const size_t slash = url.find('/', offset_past_protocol);
-    host = url.substr(offset_past_protocol, std::min(colon, slash) - offset_past_protocol);
+    const size_t colon = url.find(':', offset_past_scheme);
```
sompylasar: This colon could match the username-password separator (`http://user:pass@host:80/`).
You should likely first parse/ignore the authorization part if it is there.

net/url/test.cc
dkorolev: It's tested above, but added anyway.

net/http/impl/server.h
dkorolev: Fixed, see commits being pushed in a minute.

net/url/test.cc
dkorolev: Nice catch! Done.

net/url/test.cc
dkorolev: Fixed.

net/url/url.h
dkorolev: Edited the comment.

net/url/url.h
dkorolev: Added a TODO.

dkorolev:
@sompylasar : Thanks for the **fantastic** review. Fixed most of your comments; added ```TODO```-s where it's a bit too much for now.

Please take another look.

sompylasar:
Sorry for... ehm... well... you might want to take a look [here](http://src.chromium.org/viewvc/chrome/trunk/src/url/third_party/mozilla/url_parse.cc) and maybe [here](http://src.chromium.org/viewvc/chrome/trunk/src/url/third_party/mozilla/url_parse.h) for more insights on correct URL parsing.

net/api/impl/posix.h
```diff
@@ -86,8 +88,13 @@ class HTTPClientPOSIX final {
...
       // connection.SendEOF();
       message_.reset(new HTTPRedirectableReceivedMessage(connection));
-      // TODO(dkorolev): Rename `URL()`, it's only called so now because of HTTP request/response format.
-      const int response_code_as_int = atoi(message_->URL().c_str());
+      // TODO(dkorolev): Rename `Path()`, it's only called so now because of HTTP request/response format.
+      // Elaboration:
+      // HTTP request  message is: `GET /path HTTP/1.1`, "/path" is the second component of it.
+      // HTTP response message is: `HTTP/1.1 200 OK`, "200" is the second component of it.
+      // Thus, since the same code is used for request and response parsing as of now,
+      // the numerical response code "200" can be accessed with the same methos as the "/path".
```
sompylasar: Typo `methos` -> `method`

net/api/test.cc
```diff
@@ -135,65 +141,66 @@ class UseLocalHTTPTestServer {
...
           connection.SendHTTPResponse("", HTTPResponseCode::Found, "text/html", headers);
           serve_more_requests = true;
-        } else if (url == "/redirect-loop-2") {
+        } else if (url.path == "/redirect-loop-2") {
           HTTPHeadersType headers;
           headers.push_back(std::make_pair("Location", "/redirect-loop-3"));
           connection.SendHTTPResponse("", HTTPResponseCode::Found, "text/html", headers);
           serve_more_requests = true;
-        } else if (url == "/redirect-loop-3") {
+        } else if (url.path == "/redirect-loop-3") {
```
sompylasar: How about an infinite redirect loop?

net/url/test.cc
sompylasar: `u.query["key"]`
`u.query.get("key", "default_value")`

net/url/url.h
```diff
@@ -102,43 +106,157 @@ struct URL {
...
+    const size_t pound_sign_index = url.find('#');
+    if (pound_sign_index != std::string::npos) {
+      fragment = url.substr(pound_sign_index + 1);
+      url = url.substr(0, pound_sign_index);
+    }
+    const size_t question_mark_index = url.find('?');
+    if (question_mark_index != std::string::npos) {
+      parameters_vector = strings::SplitIntoKeyValuePairs(url.substr(question_mark_index + 1), '=', '&');
+      for (auto& it : parameters_vector) {
+        it.second = DecodeURIComponent(it.second);
```
sompylasar: The key could also be encoded, like `foo%20bar=value` should give a `"foo bar"` key.

net/url/url.h
```diff
@@ -102,43 +106,157 @@ struct URL {
...
+      }
+      parameters.insert(parameters_vector.begin(), parameters_vector.end());
+      url = url.substr(0, question_mark_index);
+    }
+    url_without_parameters = url;
+  }
+
+  std::string operator()(const std::string& key, const std::string& default_value) const {
```
sompylasar: Discussed before. The `()` operator has always had function-call semantics, please don't abuse it with a non-function-call behavior, this confuses a lot.

net/url/url.h
```diff
@@ -102,43 +106,157 @@ struct URL {
...
+    }
+  }
+
+  std::string operator[](const std::string& key) const { return operator()(key, ""); }
+
+  bool HasParameter(const std::string& key) const { return parameters.find(key) != parameters.end(); }
+
+  static std::string DecodeURIComponent(const std::string& encoded) {
+    std::string decoded;
+    for (size_t i = 0; i < encoded.length(); ++i) {
```
sompylasar: Why do you prefer not to cache constant loop end in a variable and call a `.length()` getter each time?

net/url/url.h
```diff
@@ -102,43 +106,157 @@ struct URL {
...
   }
 
-  static int DefaultPortForProtocol(const std::string& protocol) {
-    // We don't really support HTTPS/SSL or any other protocols yet -- D.K. :-)
-    return protocol == "http" ? 80 : 0;
+  std::string ComposeParameters() const {
+    std::string composed_parameters;
+    if (!parameters_vector.empty()) {
+      for (size_t i = 0; i < parameters_vector.size(); ++i) {
+        composed_parameters += "?&"[i];
```
sompylasar: Wow a nice trick with the index. Wouldn't here be a buffer overrun without clamping `i` to `[0; 1]` range?

net/url/url.h
```diff
@@ -102,43 +106,157 @@ struct URL {
...
   }
 
-  static std::string DefaultProtocolForPort(int port) { return port == 80 ? "http" : ""; }
+  std::vector<std::pair<std::string, std::string>> parameters_vector;
+  std::map<std::string, std::string> parameters;
+  std::string fragment;
+  std::string url_without_parameters;
 };
 
+struct URL : URLParameters, URLWithoutParameters {
```
sompylasar: Prefer composition over inheritance?

strings/test.cc
```diff
@@ -78,3 +89,123 @@ TEST(FixedSizeSerializer, ImplicitSyntax) {
...
+    EXPECT_EQ("one\ntwo\nthree\n", result);
+  }
+  {
+    std::string result;
+    Split("one,two,three", ',', [&result](std::string&& s) { result += s + '\n'; });
+    EXPECT_EQ("one\ntwo\nthree\n", result);
+  }
+  {
+    std::string result;
+    struct Helper {
```
sompylasar: What's this helper for? Definitely needs some comments.

sompylasar: I don't see you use `helper` instead of a lambda. Either I don't see some C++ construct, or this is a copy-paste error. Same for the next test.

net/url/url.h
sompylasar: Logically, URL "is-not-a" URLParameters, it "has-a" URLParameters. The field re-use makes sense.

net/url/test.cc
sompylasar: I meant without the path ("test") but with the slash. Either `URL("localhost/")` or `URL("localhost:/")`, the colon is tested in your test, the path with the only slash isn't.

strings/test.cc
dkorolev: Good eye! :+1: Fixed.

net/url/url.h
dkorolev: ᐧ
Logically yes, but then I would not be able to make use of C++ base classes
initialization order to keep clean.

Renamed the classes to logically read more "is-a"-ish. :)

net/url/test.cc
dkorolev: Added!
ᐧ

Thanks,
Dima

PR #39, opened 2015-01-10
Example HTTP server syntax.

dkorolev:
Hi @sompylasar and @deathbaba,

What do you think of a proposed HTTP server code syntax below?

Thanks,
Dima

### DIFF
--- a/net/api/test_server.cc.h
+++ b/net/api/test_server.cc.h
@@ -9,17 +9,213 @@
 #ifndef BRICKS_NET_API_TEST_SERVER_CC_H
 #define BRICKS_NET_API_TEST_SERVER_CC_H
 
-#include "../../dflags/dflags.h"
+#include <thread>
 
+#include "api.h"
+#include "../url/url.h"
 #include "../http/test_server.cc.h"
+
+#include "../../strings/printf.h"
 
 namespace bricks {
 namespace net {
 namespace api {
 namespace test {
 
-struct TestHTTPServer_APIImpl : bricks::net::http::test::TestHTTPServer_HTTPImpl {
-  // TODO(dkorolev): Real API implementation goes here.
+enum class Method { UNINITIALIZED, GET, POST };
+
+struct Request {
+  const url::URL& url;
+  HTTPServerConnection& connection;
+  const HTTPReceivedMessage& message;
+};
+
+struct Routes {
+  struct Helper {
+    struct Impl {
+      explicit Impl(Routes* routes) : routes_(routes) {}
+      ~Impl() {
+        assert(method_ != Method::UNINITIALIZED);
+        assert(!path_.empty());
+        assert(handler_);
+        routes_->AddRoute(method_, path_, handler_);
+      }
+      void SetHandler(std::function<void(Request&&)>&& f) {
+        assert(!handler_);
+        handler_ = f;
+      }
+      void AddConstraint(Method method) {
+        assert(method != Method::UNINITIALIZED);
+        assert(method_ == Method::UNINITIALIZED);
+        method_ = method;
+      }
+      void AddConstraint(const std::string& path) {
+        assert(!path.empty());
+        assert(path_.empty());
+        path_ = path;
+      }
+
+      Routes* routes_;
+      std::function<void(Request&&)> handler_;
+
+      Method method_ = Method::UNINITIALIZED;
+      std::string path_ = "";
+
+      Impl() = delete;
+      Impl(const Impl&) = delete;
+      Impl(Impl&&) = delete;
+      void operator=(const Impl&) = delete;
+      void operator=(Impl&&) = delete;
+    };
+
+    Helper(Routes* routes) : impl_(new Impl(routes)) {}
+    Helper(Helper&& rhs) { impl_.swap(rhs.impl_); }
+    void operator=(std::function<void(Request&&)>&& f) {
+      assert(impl_);
+      impl_->SetHandler(std::forward<std::function<void(Request && )>>(f));
+    }
+    template <typename T>
+    Helper& operator[](T&& constraint) {
+      assert(impl_);
+      impl_->AddConstraint(std::forward<T>(constraint));
+      return *this;
+    }
+
+    std::unique_ptr<Impl> impl_;
+  };
+
+  template <typename T>
+  Helper operator[](T&& constraint) {
+    Helper helper(this);
+    helper[constraint];
+    return helper;
+  }
+
+  void AddRoute(Method method, const std::string& path, std::function<void(Request&&)> handler) {
+    routes_[std::make_pair(method, path)] = handler;
+  }
+
+  void HandleRequest(HTTPServerConnection& connection) const {
+    const HTTPReceivedMessage& message = connection.Message();
+    const Method method = message.Method() == "GET" ? Method::GET : Method::POST;  // TODO(dkorolev).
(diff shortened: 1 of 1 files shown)


dkorolev:
Please look at the ```routes[Method::GET]["/get"] = [](Request&&) { ... };``` part.

I'll clean up the code and move files around, but would prefer to do it once we all like the general syntax. I feel there is more work to be done there from my side.

Thanks,
Dima

net/api/test_server.cc.h
```diff
@@ -9,17 +9,216 @@
...
+    const HTTPReceivedMessage& message = connection.Message();
+    const Method method = message.Method() == "GET" ? Method::GET : Method::POST;  // TODO(dkorolev).
+    const std::string& path = message.Path();  // TODO(dkorolev): Rename message.Path(), it's incorrect.
+    const url::URL url = url::URL(path);
+    const auto cit = routes_.find(std::make_pair(method, url.path));
+    if (cit != routes_.end()) {
+      Request request{url, connection, message};
+      cit->second(std::forward<Request>(request));
+    } else {
+      ASSERT_TRUE(false) << "Method not implemented.";
```
sompylasar: There should be some default handler giving a valid HTTP response instead.

net/api/test_server.cc.h
```diff
@@ -9,17 +9,216 @@
...
+  void AddRoute(Method method, const std::string& path, std::function<void(Request&&)> handler) {
+    routes_[std::make_pair(method, path)] = handler;
+  }
+
+  void HandleRequest(HTTPServerConnection& connection) const {
+    const HTTPReceivedMessage& message = connection.Message();
+    const Method method = message.Method() == "GET" ? Method::GET : Method::POST;  // TODO(dkorolev).
+    const std::string& path = message.Path();  // TODO(dkorolev): Rename message.Path(), it's incorrect.
+    const url::URL url = url::URL(path);
+    const auto cit = routes_.find(std::make_pair(method, url.path));
```
sompylasar: Here should be some route matcher that understands URL parameters.

net/api/test_server.cc.h
```diff
@@ -9,17 +9,216 @@
...
+class Server {
+ public:
+  Server(int port, bool& time_to_terminate, Routes routes)
+      : port_(port),
+        time_to_terminate_(time_to_terminate),
+        routes_(routes),
+        thread_(&Server::Code, this, Socket(port_)) {}
+
+  ~Server() {
+    HTTP(GET(strings::Printf("localhost:%d/killtestserver", port_)));
```
sompylasar: There should be some internal way to signal the server to stop, not via HTTP.

sompylasar:
Why not `express.js`-like API like we discussed before?

```javascript
app = express();

app.get("/path", commonHandler, onGetPath);
app.post("/something", commonHandler, onPostSomething);

app.listen(port);
```

net/api/test_server.cc.h
```diff
@@ -9,17 +9,216 @@
...
+      impl_->SetHandler(std::forward<std::function<void(Request && )>>(f));
+    }
+    template <typename T>
+    Helper& operator[](T&& constraint) {
+      assert(impl_);
+      impl_->AddConstraint(std::forward<T>(constraint));
+      return *this;
+    }
+
+    std::unique_ptr<Impl> impl_;
```
deathbaba: Why do you need impl in header-only?

net/api/test_server.cc.h
```diff
@@ -9,17 +9,216 @@
...
+    }
+
+    bool time_to_terminate_ = false;
+    Server server_;
+  };
+
+  class UniqueImpl {
+   public:
+    UniqueImpl(int port) : port_(port), impl_(new Impl(port)) {}
+    std::string BaseURL() const { return std::string("http://localhost:") + std::to_string(port_); }
```
deathbaba: If port is default (80 for http) it's better do not return it.

net/api/test_server.cc.h
```diff
@@ -9,17 +9,216 @@
...
-  // TODO(dkorolev): Real API implementation goes here.
+enum class Method { UNINITIALIZED, GET, POST };
+
+struct Request {
+  const url::URL& url;
+  HTTPServerConnection& connection;
+  const HTTPReceivedMessage& message;
+};
+
+struct Routes {
```
deathbaba: Routes is not the best name I suppose. In NGINX they are called locations, in golang - handlers, may be there are better names.

sompylasar:
In C++ the syntax would look like:

```cpp
ExpressApi app;

const auto commonHandler = [](Request& req, Response& res) {
    
};
const auto onGetPath = [](Request& req, Response& res) {
    assert(req.path == "/path");
    res.write("OK");
};

app.get("/path", commonHandler, onGetPath);

```

deathbaba:
This is the same approach as in golang :)

sompylasar:
http://expressjs.com/4x/api.html

net/api/test_server.cc.h
sompylasar: It's better to return or wrap the return value with a URL object which copes with the default port.

sompylasar:
Please note the connect.js/express.js core concept called middleware, their API architecture is built upon this concept.

A middleware is a function that takes the request and response objects and operates on them. There is a chain of middlewares, some of them only trigger on requests with a specific path or a set of paths, others trigger on every request. A middleware may also take a reference to the 'next' function for async continuation; if the function does not take 'next', the next middleware in the chain is called implicitly.

This is a very powerful concept and having first-class functions we could make similar API in C++.

net/api/test_server.cc.h
dkorolev: There will be, this is just for the unit test.

net/api/test_server.cc.h
dkorolev: True indeed, "/user/:uid" is coming in V2 very soon.

net/api/test_server.cc.h
dkorolev: True that, although keep in mind that ```::accept()``` is a blocking call in POSIX, and making another server connection is a much easier way to signal the server that is has to be terminated.

Even if it would be an HTTP route at the end, the user will not see it.

net/api/test_server.cc.h
dkorolev: To deal with move-construction seamlessly.

One ```routes[foo][bar][baz] = meh;``` should result in exactly one ```Impl``` object created, populated with several ```AddConstraint()``` and one ```SetHandler()```, and then call ```routes_->AddRoute()``` once for the whole statement.

Returning from a top-level ```operator[]``` results in  top-level object copied; thus, I made the instance of ```Helper``` to be copied, while the ```Impl``` one stays one.

net/api/test_server.cc.h
dkorolev: Agreed, this code is for the unit test exclusively.

net/api/test_server.cc.h
dkorolev: We'll need to come up with a name :-)

I like neither handler (too broad) nor locations (too web-oriented). Route-based handlers is what sounds best to me so far, but I'll be flexible.

dkorolev:
Thanks @sompylasar and @deathbaba,

As for code-level comments: Most will get resolves once I separate server code and test code. These commits so far are just the RFC; I'll refactor those accordingly once we agree on the user-facing syntax.

As for design-level comments: Replied via e-mail, would appreciate your comments a lot.

Thanks,
Dima

PR #40, opened 2015-01-11
Wrapper for plotutils graph library.

dkorolev:


plotutils/graph.h
```diff
@@ -0,0 +1,205 @@
...
+    return *this;
+  }
+
+  Graph& OutputFormat(const std::string& output_format) {
+    parameters_["-T"] = output_format;
+    return *this;
+  }
+
+  operator std::string() const {
+    const std::string input_file_name = bricks::FileSystem::GenTmpFileName();
```
deathbaba: Why use temporary files instead of stdin and stdout?

plotutils/graph.h
```diff
@@ -0,0 +1,205 @@
...
+// 3: blue, solid
+// 4: magenta, solid
+// 5: cyan, solid
+// Linemodes #6 through #10 use the same five colors, but are dotted; linemodes #11 through #15 are dotdashed;
+// linemodes #16 through #20 are shortdashed; and linemodes #21 through #25 are longdashed.
+// So besides linemode #0, there are a total of 25 distinct colored linemodes
+enum class LineMode : int {
+  None = 0,
+};
+enum class LineColor : int { Red = 1, Green = 2, Blue = 3, Magenta = 4, Cyan = 5 };
```
sompylasar: Why enum, not RGB(A) with some color constants for named aliases?

dkorolev: Those are just ```plotutils```' default colors. It doesn't offer much flexibility.

FTR, this module is not to serve production; it's more for tests, demos, hacks and debugging purposes, to expose an endpoint showing some quick chart.

plotutils/graph.h
dkorolev: If you meant ```popen()```, it's one-directional and Linux-specific. I thought, if at least one temp file is there to create, I might as well create two.

deathbaba: It's ok for small load.

dkorolev:
@deathbaba Polite ping?

plotutils/graph.h
sompylasar: I did not realized first this is for a third-party command line tool. It's OK then.

PR #43, opened 2015-01-17
New HTTP API.

dkorolev:
New HTTP API, based on registering and unregistering individual handlers.
New exceptions with messages.

Simplified HTTP tests by a) getting rid of ```httpbin.org```, and b) separating the ones in ```net/api``` and ```net/http```.

This change is actually negative in the number of lines of code added vs. removed.


net/api/api.h
```diff
@@ -48,53 +48,52 @@ SOFTWARE.
...
-#include "impl/java.h"
-typedef bricks::net::api::HTTPClientImpl<aloha::HTTPClientPlatformWrapper> HTTP_TYPE;
+#include "impl/java_client.h"
+#include "impl/posix_server.h"
+typedef bricks::net::api::aloha::HTTPClientPlatformWrapper HTTP_CLIENT;
 #else
 #error "No implementation for `net/api/api.h` is available for your system."
 #endif
 
+typedef bricks::net::api::HTTPImpl<HTTP_CLIENT, bricks::net::api::HTTPServerPOSIX> HTTP_TYPE;
```
sompylasar: Why `HTTP_TYPE`, not `HTTP_IMPL`?

dkorolev: https://github.com/KnowSheet/Bricks/pull/44 :)

PR #45, opened 2015-01-18
HTTP chunked response wrapper.

dkorolev:


dkorolev:
CC @sompylasar.

dkorolev:
@deathbaba : I have added a few commits into this PR, mostly chunked format fixes.
@sompylasar : And here comes the chunked JSON example code for you!

Thanks and PTAL,
Dima

net/api/impl/posix_server.h
```diff
@@ -126,24 +137,29 @@ class HTTPServerPOSIX final {
...
+        Request request{url, connection, message};
+        {
+          // TODO(dkorolev): Read-write lock for performance?
+          std::lock_guard<std::mutex> lock(mutex_);
+          const auto cit = handlers_.find(url.path);
+          if (cit != handlers_.end()) {
+            // TODO(dkorolev): Properly handle the shutdown case when the handler spawns another thread.
+            cit->second(std::move(request));
+          } else {
+            connection.SendHTTPResponse("", HTTPResponseCode::NotFound);
```
deathbaba: Should else case be under mutex? It's relatively slow and will definitely block other threads.

dkorolev: Fixed in 8e590db.

(The logic is going to change to indicate that a particular handler is now active, so that it's not deregistered while running. Stay tuned.)

PR #47, opened 2015-01-19
Added /meta to the server.

dkorolev:


dkorolev:
@sompylasar : To review after https://github.com/KnowSheet/Bricks/pull/46 is in.
@deathbaba : FYI.

dkorolev:
@deathbaba @sompylasar Status check?

net/http/impl/server.h
```diff
@@ -370,22 +392,31 @@ class HTTPServerConnection {
         }
       }
 
-      // Only support containers of chars and bytes, this does not cover std::string.
+      // The actual implementation of sending HTTP chunk data.
       template <typename T>
-      inline typename std::enable_if<sizeof(typename std::remove_reference<T>::type::value_type) == 1>::type
-      Send(T&& data) {
+      void SendImpl(T&& data) {
```
sompylasar: How about naming it `SendHTTPChunkImpl`?

net/http/impl/server.h
```diff
@@ -370,22 +392,31 @@ class HTTPServerConnection {
...
         connection_.BlockingWrite(strings::Printf("%X", data.size()));
         connection_.BlockingWrite(kCRLF);
         connection_.BlockingWrite(data);
         connection_.BlockingWrite(kCRLF);
       }
 
+      // Only support STL containers of chars and bytes, this does not yet cover std::string.
+      template <typename T>
+      inline typename std::enable_if<sizeof(typename std::remove_reference<T>::type::value_type) == 1>::type
+      Send(T&& data) {
```
sompylasar: How about naming these `SendHTTPChunk`?

net/http/impl/server.h
```diff
@@ -370,22 +392,31 @@ class HTTPServerConnection {
...
-        connection_.BlockingWrite(data);
-        connection_.BlockingWrite(kCRLF);
+      inline void Send(const std::string& data) { SendImpl(data); }
+
+      // Support objects that can be serialized as JSON-s via Cereal.
+      template <class T>
+      inline typename std::enable_if<
+          (cerealize::is_cerealizeable<typename std::remove_reference<T>::type>::value)>::type
+      Send(T&& object) {
+        SendImpl(cerealize::JSON(object) + kCRLF);
```
sompylasar: This JSON payload can be confusing because of `kCRLF` that is also used in the chunked encoding. It's not obvious that `kCRLF` here stands for the "new-line separator" in the "JSON-per-line" format which is a higher-level abstraction ontop of the chunked encoding.

net/http/impl/server.h
```diff
@@ -350,7 +360,19 @@ class HTTPServerConnection {
...
+  // Support objects that can be serialized as JSON-s via Cereal.
+  template <class T>
+  inline typename std::enable_if<
+      (cerealize::is_cerealizeable<typename std::remove_reference<T>::type>::value)>::type
+  SendHTTPResponse(T&& object,
+                   HTTPResponseCode code = HTTPResponseCode::OK,
+                   const std::string& content_type = DefaultContentType(),
+                   const HTTPHeadersType& extra_headers = HTTPHeadersType()) {
+    // TODO(dkorolev): We should probably make this not only correct but also efficient.
+    const std::string s = cerealize::JSON(object) + kCRLF;
```
sompylasar: It's not necessary to put the trailing `kCRLF` here.

net/http/impl/server.h
```diff
@@ -335,14 +336,23 @@ class HTTPServerConnection {
     connection_.BlockingWrite(kCRLF);
```
sompylasar: This `kCRLF` is wrong here. It's not included into the measured `Content-Length` and it's not in the specs either: http://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html

net/http/impl/server.h
sompylasar: Sorry I missed the container name `ChunkedResponseSender`. It's OK then.

net/http/impl/server.h
sompylasar: Sorry I missed the container name `ChunkedResponseSender`. It's OK then.

sompylasar:
LGTM except the trailing CRLF that is not included into the Content-Length and the JSON-per-line separator.

net/http/impl/server.h
dkorolev: Done in 65adb1f56de2d753fb19a2a5e42addaa5eb8dc99

net/http/impl/server.h
dkorolev: Fixed in 65adb1f56de2d753fb19a2a5e42addaa5eb8dc99

net/http/impl/server.h
dkorolev: Done in f416cddeb4286cd6a7b2cfa5fc6d98fde877c4da

dkorolev:
Fixed the CRLF-s.

@sompylasar, please take another look.
@deathbaba : would appreciate it if you could glance through as well.

deathbaba:
Hope it will work with chunked binary data.

PR #51, opened 2015-01-26
HTTP handlers by-ref/by-val/by-ptr semantics + minor changes.

dkorolev:


dkorolev:
@deathbaba : Please review.
@sompylasar : FYI.

Makefile change also in https://github.com/KnowSheet/KnowSheet/pull/2.

dkorolev:
@deathbaba ping?

cerealize/test/impl.cc
```diff
@@ -202,6 +202,9 @@ TEST(Cerealize, ConsumerSupportsPolymorphicTypes) {
 
     // This function will not be called since the `EventAppSuspend` type is not in `DERIVED_TYPE_LIST` above.
     void operator()(const EventAppSuspend& e) { e.AppendTo(os) << " *** NOT CALLED SUSPEND *** " << std::endl; }
+
+    ExampleConsumer() {}
```
deathbaba: = default?

dkorolev: Fixed.

PR #52, opened 2015-01-28
Added gnuplot support.

dkorolev:
Need somewhat fancier graphs for the demo, ```plotutils``` is not enough.


dkorolev:
@deathbaba , poke?

graph/gnuplot.h
```diff
@@ -0,0 +1,188 @@
...
+        f << (i ? ", " : "plot ") << "'-'";
+        plots_[i]->AppendMetadata(f);
+      }
+      f << std::endl;
+      for (const auto& p : plots_) {
+        p->AppendData(f);
+        f << "end" << std::endl;
+      }
+    }
+    ::system(strings::Printf("gnuplot <%s >%s\n", input_file_name.c_str(), output_file_name.c_str()).c_str());
```
deathbaba: What about the simplest error checking?

dkorolev: I'll make a platform-neutral ```::system()``` call and throw an exception -- when merging a few places I use ```::system()``` from as of today.

PR #53, opened 2015-01-29
chunked_demo integration with frontend

sompylasar:
* Added "layout" structures and the route handler.
* Added "Access-Control" headers required to access the HTTP API from
different Origin (running the chunked_demo and the frontend on
different machines).
* Added "Connection: close" headers for immediately closed connections
(see the HTTP Connection headers RFC).
* Changed "Content-Type" to actual.
* Changed data generation to reduce send frequency (the browser got
overwhelmed by the amount of points received to render; frontend
optimization is likely to be required).

### DIFF
--- a/net/api/chunked_demo.cc
+++ b/net/api/chunked_demo.cc
@@ -109,8 +109,38 @@ using namespace bricks::net::api;
 using namespace bricks::cerealize;
 using bricks::time::Now;
 using bricks::strings::Printf;
+using bricks::net::HTTPResponseCode;
 
 DEFINE_int32(port, 8181, "The port to serve chunked response on.");
+
+struct LayoutCell {
+  std::string meta_url = "/meta";
+  
+  template <typename A>
+  void serialize(A& ar) {
+    ar(CEREAL_NVP(meta_url));
+  }
+};
+
+struct LayoutItem {
+  std::vector<LayoutItem> row;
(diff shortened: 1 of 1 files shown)


net/api/chunked_demo.cc
```diff
@@ -109,9 +109,39 @@ using namespace bricks::net::api;
...
+  }
+};
+
+struct LayoutItem {
+  std::vector<LayoutItem> row;
+  std::vector<LayoutItem> col;
+  LayoutCell cell;
+
+  template <typename A>
+  void serialize(A& ar) {
```
dkorolev: This will not work for de-serialization.

Which is totally fine if you never de-serialize struct ```LayoutItem``` but only serialize it -- but please make sure it does not compile when attempting to de-serealize it then.

Should be a matter of either specifying a non-template type for the parameter (to effectively support ```JSON.stringify()``` type of serialization only), or by inserting a ```static_assert``` checking that the ```A``` is a serialization, not de-serialization medium.

sompylasar: Thanks. It's for serialization only; otherwise we'll have to change the layout definition structure. Please help me to write this assert expression, I'm not that confident in C++.

dkorolev: I'll look into it over the weekend. Thanks!

PR #56, opened 2015-02-03
Got rid of '&&' for HTTP requests.

dkorolev:
Performance later, less confusion for the users comes first.


deathbaba:
What was wrong with &&?

dkorolev:
1. I just thought it's a bit too confusing for the developer to keep thinking of it.
2. If ```&&``` is the right implementation, we can always ```typedef``` it later.
3. The most complex usecase to support is when the ```Request``` object is being put into a message queue and then server later on, from a different thread. I realized it is easier to have that working first and worry about syntax nits later.

Thanks,
Dima

PR #58, opened 2015-02-05
Test fixes.

dkorolev:


dkorolev:
Ref.  https://github.com/KnowSheet/Bricks/issues/57

sompylasar:
Thanks,looks good.

sompylasar:
Btw, maybe test at least two headers so that the loop is tested, too?

dkorolev:
Added a test with two headers. Didn't have cycles for anything else today except this change and https://github.com/KnowSheet/Bricks/pull/59. :(

@sompylasar : Please click "Merge" if this one looks good. Alex @deathbaba prefers it this way ("don't merge your own code"), and we're somewhat following this practice so far. Thanks!

PR #62, opened 2015-02-12
Two minor cleanups and localized a Windows test issue.

dkorolev:


deathbaba:
I'm looking to replace cereal with own bicycle. What are server-side requirements to it?

PR #63, opened 2015-02-14
Shortened cereal-ization magic.

dkorolev:


dkorolev:
@sompylasar PTAL
@deathbaba FYI

PR #65, opened 2015-02-16
Shaping Bricks' external interfaces.

dkorolev:
* HTTP(POST(...)) can now be run without body.
* HTTP(POST(...)) can now accept user object directly, it will be JSON-ified automatically.
* Added a somewhat smart ```JSONParse<>```.
* Ported @sompylasar's MIME type detection code into Bricks.
* Added ```Connection: close/keep-alive``` as part of Bricks' ```net/api```.
* A few minor cleanup changes.

### DIFF
--- a/file/file.h
+++ b/file/file.h
@@ -52,7 +52,7 @@ namespace bricks {
 // Platform-indepenent, injection-friendly filesystem wrapper.
 // TODO(dkorolev): Move the above methods under FileSystem.
 struct FileSystem {
-  static std::string GetFileExtension(const std::string& file_name) {
+  static inline std::string GetFileExtension(const std::string& file_name) {
     const static auto get_extension_from_basename = [](const std::string& fn) {
       const size_t i = fn.find_last_of('.');
       return i == std::string::npos ? "" : fn.substr(i + 1);
--- a/file/file.h
+++ b/file/file.h
@@ -52,7 +52,24 @@ namespace bricks {
 // Platform-indepenent, injection-friendly filesystem wrapper.
 // TODO(dkorolev): Move the above methods under FileSystem.
 struct FileSystem {
-  static inline std::string ReadFileAsString(std::string const& file_name) {
+  static std::string GetFileExtension(const std::string& file_name) {
+    const static auto get_extension_from_basename = [](const std::string& fn) {
+      const size_t i = fn.find_last_of('.');
+      return i == std::string::npos ? "" : fn.substr(i + 1);
+    };  // LCOV_EXCL_LINE
+
+    static_assert(std::string::npos + 1 == 0, "This invariant is used in the next line.");
+    const size_t basename_offset = file_name.find_last_of("/\\") + 1;
+
+    if (!basename_offset) {
+      return get_extension_from_basename(file_name);
+    } else {
+      // Unfortunately, `std::string::find_last_of()` accepts index to stop at, not index to start form.
+      return get_extension_from_basename(file_name.substr(basename_offset));
+    }
+  }
+
+  static inline std::string ReadFileAsString(const std::string& file_name) {
     try {
       std::ifstream fi;
       fi.exceptions(std::ifstream::failbit | std::ifstream::badbit);
--- a/file/file.h
+++ b/file/file.h
@@ -53,19 +53,11 @@ namespace bricks {
 // TODO(dkorolev): Move the above methods under FileSystem.
 struct FileSystem {
   static inline std::string GetFileExtension(const std::string& file_name) {
-    const static auto get_extension_from_basename = [](const std::string& fn) {
-      const size_t i = fn.find_last_of('.');
-      return i == std::string::npos ? "" : fn.substr(i + 1);
-    };  // LCOV_EXCL_LINE
-
-    static_assert(std::string::npos + 1 == 0, "This invariant is used in the next line.");
-    const size_t basename_offset = file_name.find_last_of("/\\") + 1;
-
-    if (!basename_offset) {
-      return get_extension_from_basename(file_name);
+    const size_t i = file_name.find_last_of("/\\.");
+    if (i == std::string::npos || file_name[i] != '.') {
+      return "";
     } else {
-      // Unfortunately, `std::string::find_last_of()` accepts index to stop at, not index to start form.
-      return get_extension_from_basename(file_name.substr(basename_offset));
+      return file_name.substr(i + 1);
     }
   }
 
--- a/file/test.cc
+++ b/file/test.cc
@@ -49,6 +49,24 @@ TEST(File, JoinPath) {
   ASSERT_THROW(FileSystem::JoinPath("/foo/", ""), FileException);
 }
 
+TEST(File, GetFileExtension) {
+  EXPECT_EQ("", FileSystem::GetFileExtension(""));
+  EXPECT_EQ("", FileSystem::GetFileExtension("a"));
+  EXPECT_EQ("b", FileSystem::GetFileExtension("a.b"));
+  EXPECT_EQ("c", FileSystem::GetFileExtension("a.b.c"));
+  EXPECT_EQ("a", FileSystem::GetFileExtension(".a"));
+  EXPECT_EQ("a", FileSystem::GetFileExtension("..a"));
+  EXPECT_EQ("", FileSystem::GetFileExtension("a/b"));
+  EXPECT_EQ("b", FileSystem::GetFileExtension("a/.b"));
+  EXPECT_EQ("", FileSystem::GetFileExtension("a/b/c"));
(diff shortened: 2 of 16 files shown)


dkorolev:
@deathbaba PTAL.
@sompylasar PTAL.

Thanks!
Dima

dkorolev:
Added a few small commits, no major changes, ready for your guys' review.

@sompylasar @deathbaba

dkorolev:
Gentlemen, esp. Alex @deathbaba, would you please be so kind to dedicate a couple dozen minutes of your valuable time to this piece of craft being considered under this GitHub link? Much appreciated in advance.

dkorolev:
Ready to look at this one. PTAL at https://github.com/KnowSheet/Bricks/pull/66 afterwards.

file/file.h
```diff
@@ -52,7 +52,24 @@ namespace bricks {
...
 // TODO(dkorolev): Move the above methods under FileSystem.
 struct FileSystem {
-  static inline std::string ReadFileAsString(std::string const& file_name) {
+  static inline std::string GetFileExtension(const std::string& file_name) {
+    const static auto get_extension_from_basename = [](const std::string& fn) {
+      const size_t i = fn.find_last_of('.');
+      return i == std::string::npos ? "" : fn.substr(i + 1);
+    };  // LCOV_EXCL_LINE
+
+    static_assert(std::string::npos + 1 == 0, "This invariant is used in the next line.");
```
deathbaba: Suspicious trick.

deathbaba: But looks like standard has well-defined npos with -1, so it should be safe.

file/test.cc
```diff
@@ -49,6 +49,24 @@ TEST(File, JoinPath) {
   ASSERT_THROW(FileSystem::JoinPath("/foo/", ""), FileException);
 }
 
+TEST(File, GetFileExtension) {
+  EXPECT_EQ("", FileSystem::GetFileExtension(""));
```
deathbaba: Would be easier to read in a char array form + for()

net/api/test.cc
```diff
@@ -58,6 +59,17 @@ DEFINE_int32(net_api_test_port,
              "lifetime of the binary.");
 DEFINE_string(net_api_test_tmpdir, ".noshit", "Local path for the test to create temporary files in.");
 
+struct HTTPAPITestObject {
+  int number;
+  std::string text;
+  std::vector<int> array;  // Visual C++ does not support the `= { 1, 2, 3 };` non-static member initialization.
```
deathbaba: Code compiles well on VS2015.
```
struct A {
  int a = 4;
  std::vector<int> v = { 1, 2, 3 };
};


int _tmain(int argc, _TCHAR* argv[])
{
  A a;
	return a.a + a.v.size();
}
```

file/file.h
dkorolev: Yes, it's guaranteed to be the case. The ```static_assert``` is more of a comment for the reader of the code.

file/test.cc
dkorolev: No. If a test fails, the user should immediately see which one did, by the line number.

(This is one of many small things I leaned in Google.)

net/api/test.cc
msft-dkorolev: Not for me. Visual Studio Preview 2015.

```
1>------ Build started: Project: net_http, Configuration: Debug Win32 ------
1>  test.cc
1>test.cc(59): error C2797: 'HTTPTestObject::array': list initialization inside member initializer list or non-static data member initializer is not implemented
1>  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\utility(138): note: while compiling class template member function 'std::pair<const _Kty,_Ty>::pair<int,const char(&)[9],void>(_Other1 &&,_Other2) noexcept(false)'
1>          with
1>          [
1>              _Kty=int
1>  ,            _Ty=std::string
1>  ,            _Other1=int
1>  ,            _Other2=const char (&)[9]
1>          ]
```

```

struct HTTPTestObject {
  int number = 42;
  std::string text = "text";
  std::vector<int> array = { 1, 2, 3 };  // Visual C++ does not support the `= { 1, 2, 3 };` non-static member initialization.
  // HTTPTestObject() : number(42), text("text"), array({1, 2, 3}) {}
  template <typename A>
  void serialize(A& ar) {
    ar(CEREAL_NVP(number), CEREAL_NVP(text), CEREAL_NVP(array));
  }
};
```

dkorolev:
@deathbaba Thanks, and how about a real review? :-)

dkorolev:
@deathbaba @sompylasar PTAL.

file/test.cc
```diff
@@ -49,6 +49,24 @@ TEST(File, JoinPath) {
...
+  EXPECT_EQ("", FileSystem::GetFileExtension("a"));
+  EXPECT_EQ("b", FileSystem::GetFileExtension("a.b"));
+  EXPECT_EQ("c", FileSystem::GetFileExtension("a.b.c"));
+  EXPECT_EQ("a", FileSystem::GetFileExtension(".a"));
+  EXPECT_EQ("a", FileSystem::GetFileExtension("..a"));
+  EXPECT_EQ("", FileSystem::GetFileExtension("a/b"));
+  EXPECT_EQ("b", FileSystem::GetFileExtension("a/.b"));
+  EXPECT_EQ("", FileSystem::GetFileExtension("a/b/c"));
+  EXPECT_EQ("c", FileSystem::GetFileExtension("a/b/.c"));
```
sompylasar: We maybe should handle linux-style-hidden files without extension differently? E.g. `.htaccess`, `.gitignore` etc.

sompylasar: Forget it. `pathinfo` does the same thing as we do: http://www.tools4noobs.com/online_php_functions/pathinfo/

net/api/test.cc
```diff
@@ -237,13 +278,108 @@ TEST(HTTPAPI, PostFromBufferToBuffer) {
...
+            HTTP(POST(Printf("localhost:%d/post_const_char_ptr", FLAGS_net_api_test_port),
+                      static_cast<const char*>("const char*"),
+                      "text/plain")).body);
+}
+
+TEST(HTTPAPI, RespondWithStringAsString) {
+  HTTP(FLAGS_net_api_test_port).ResetAllHandlers();
+  HTTP(FLAGS_net_api_test_port).Register("/respond_with_std_string", [](Request r) {
+    ASSERT_FALSE(r.http.HasBody());
+    r.connection.SendHTTPResponse(std::string("std::string"), HTTPResponseCode::InternalServerError);
```
sompylasar: Why `InternalServerError`?

net/http/mime_type.h
```diff
@@ -0,0 +1,69 @@
...
+      {"json", "application/json"},
+      {"css", "text/css"},
+      {"html", "text/html"},
+      {"htm", "text/html"},
+      {"txt", "text/plain"},
+      {"png", "image/png"},
+      {"jpg", "image/jpeg"},
+      {"jpeg", "image/jpeg"},
+      {"gif", "image/gif"},
+      {"svg", "image/svg+xml"}  // TODO(sompylasar): Is this the right MIME for svg?
```
sompylasar: Yes, it is.

net/http/mime_type.h
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  std::string extension = bricks::FileSystem::GetFileExtension(file_name);
+  std::transform(extension.begin(), extension.end(), extension.begin(), ::tolower);
+  const auto cit = file_extension_to_mime_type_map.find(extension);
+  if (cit != file_extension_to_mime_type_map.end()) {
+    return cit->second;
+  } else {
+    return "text/plain";
```
sompylasar: I think we should provide this as an argument `default_mime_type`. Sometimes the caller may want to distinguish known types from unknown, and do not give out files with unknown types or make them directly downloadable via `application/octet-stream` plus a set of headers.

dkorolev: Done in the other pull request.

file/test.cc
dkorolev: PHP? :-)

net/api/test.cc
dkorolev: Lack of sleep is a terrible thing, you know? Ah, wait, never mind, you do :( Fixed!

net/api/test.cc
deathbaba: May be it is not the latest ctp 5 build, strange.

dkorolev:
@sompylasar : Fixed your comments, thank you!

@deathbaba , @sompylasar : I'm working on two more pieces right now, and they both depend on this pull request. Please give it some love. Thank you!

dkorolev:
That was quick thank you!

PR #69, opened 2015-02-20
Polishing the API and the docu. Dot notation and a couple getters.

dkorolev:


dkorolev:
@sompylasar @deathbaba PTAL.

README.md
```diff
@@ -242,39 +188,96 @@ struct PennyInput {
   std::vector<int> x;
   template <typename A> void serialize(A& ar) {
     ar(CEREAL_NVP(op), CEREAL_NVP(x));
-  }   
-  // TODO(dkorolev): Safe mode wrt parsing malformed JSON.
+  }
```
sompylasar: GitHub shows something red here and below (a space?).

README.md
```diff
@@ -242,39 +188,96 @@ struct PennyInput {
   std::vector<int> x;
   template <typename A> void serialize(A& ar) {
     ar(CEREAL_NVP(op), CEREAL_NVP(x));
-  }   
-  // TODO(dkorolev): Safe mode wrt parsing malformed JSON.
+  } 
+  void FromInvalidJSON(const std::string& input_json) {
+    op = "JSON parse error: " + input_json;
```
sompylasar: Is this OK to put an error message into the "operation" field?

dkorolev:
Fixed the whitespaces. Some are necessary for ```README.md``` indent, but certainly not all.

And you're right about the error message, let me fix that one too in a sec.

dkorolev:
Fixed, PTAL.

README.md
```diff
@@ -240,41 +183,100 @@ HTTP(port).Register("/found", [](Request r) {
...
+      for (const auto v : input.x) {
+        result += v;
+      }
+    } else if (input.op == "mul") {
+      result = 1;
+      for (const auto v : input.x) {
+        result *= v;
+      }
     }
+    r(PennyOutput{"", result});
```
sompylasar: Was there some sort of unknown operation error few minutes ago?

dkorolev: There was no, but I just added it while you've been looking at this code :)

dkorolev: Just added a test. On the other hand, we also need a "not enough parameters" case, give me a sec.

net/http/impl/server.h
```diff
@@ -505,9 +523,10 @@ class HTTPServerConnection final {
     std::unique_ptr<Impl> impl_;
   };
 
-  inline ChunkedResponseSender SendChunkedHTTPResponse(HTTPResponseCode code = HTTPResponseCode::OK,
-                                                       const std::string& content_type = DefaultContentType(),
-                                                       const HTTPHeaders& extra_headers = HTTPHeaders()) {
+  inline ChunkedResponseSender SendChunkedHTTPResponse(
+      HTTPResponseCodeValue code = HTTPResponseCodeValue::OK,
```
sompylasar: `HTTPResponseCodeValue::OK` -> `HTTPResponseCode.OK` (like everywhere else)

dkorolev:
Just pushed Penny's logic to be real, with the tests. Thanks for pointing it out, PTAL.

dkorolev:
@sompylasar : Good eye! That was the very only last one left :-) Fixed it too.

dkorolev:
(Also, you should really go to bed ...)

sompylasar:
Yep... The birds are singing...

Right, this one is good to merge if the tests are passing.

dkorolev:
Of course, they are, with 100% coverage :) Thanks for a late night / early morning review, really appreciate it.

PR #83, opened 2015-02-23
More exception types for ScanDir().

dkorolev:
Fixing https://github.com/KnowSheet/Bricks/issues/72


dkorolev:
@sompylasar @deathbaba : PTAL.

PR #102, opened 2015-02-27
Finalizing net/{api+tcp} and locking down net/http. Several simplifications.

dkorolev:
Hi!

First off, I'm officially an underqualified network programmer. The problem was solved by telling `::send()` that more data is there to come, by passing `MSG_MORE` as `flags` when there is more data to come. I have made the necessary changes; please take a look on whether I missed any of those and/or did them wrong.

Also, `MSG_WAITALL` for `::recv()` does (most of) the trick with synchronous receive. I've made use of it too. So, two network programming blunders. So much for good architecture. :-(

Now, state of the art.

`net/tcp` and `net/api` are stable.

I've run them with `--gtest_repeat=10000 --gtest_break_on_failure` both on Linux (`-O3` and `-g`) and on Windows (`x64/Debug` and `x64/Release).

(Although with the most recent change, `--gtest_break_on_failure` defaults to `true`, and is thus optional.)

I even ran them with `--gtest_filter=*Smoke*` / `--gtest_filter=*Smoke*`, to only run the simplest and fastest tests (fast cycles have been showing random issues before on 1'000+ iterations. Btw, on Linux Bricks is showing >1K QPS on basic `tcp/http/api` tests, even on the ones parsing the URL.)

The `het/http` test is stable on Linux, with all the above statements about stress testing being true.

On Windows though, the `net/http` test can be made fail. It happens **only** in `Release` mode, and only when running the binary outside the IDE; in other words, it's not debuggable. :-1: It is, however, consistently repeatable; the following command does it: `Bricks\VisualStudioSolution\x64\Release> .\net_http.exe --gtest_repeat=1000 --gtest_filter=*Smoke`. On my desktop this command fails within a second; sometimes on the 2nd or 3rd, sometimes on 100++-th iteration, but still within a second.

I have narrowed the issue to the client thread not being able to read the expected response from the socket. The exception is either `SocketReadException` or `ConnectionResetByPeer`, with no apparent pattern.

With respect to this problem, I am close to giving up since a) it is low-pri, b) it is a bit too deep, and c) interestingly, all `net/api` tests, which are what our users would use, pass like a Swiss watch.

One reasonable explanation may be that `net/api` opens the listener socket on the specified port once per binary, while `net/http` tests re-open it for each test. Perhaps some form of `flush()` w.r.t. waiting for internal buffers per port to get released, might do this trick. Another point to support this theory it that `net/api` perform much better QPS-wise than the ones in `net/tcp` or `net/http`, despite the latter ones being more trivial.

(A relatively simple test might me to run `WSAStartup()/WSAShutdown()` per each test, not once per binary. But it's 3am.)

@deathbaba : Please review. Thanks!

@mzhurovich : Thanks for volunteering to look into! An extra pair of `C++`-savvy eyes certainly won't hurt :+1:

@sompylasar : Sorry man, I've changed `Blocking{Read/Write}` once again. Hope it doesn't affect your REPL code much; in my defense, the changes are probably either irrelevant to your code or making it simpler.

### DIFF
--- a/net/api/impl/posix_client.h
+++ b/net/api/impl/posix_client.h
@@ -70,21 +70,22 @@ class HTTPClientPOSIX final {
       }
       all_urls.insert(composed_url);
       Connection connection(Connection(ClientSocket(parsed_url.host, parsed_url.port)));
-      connection.BlockingWrite(request_method_ + ' ' + parsed_url.path + parsed_url.ComposeParameters() +
-                               " HTTP/1.1\r\n");
-      connection.BlockingWrite("Host: " + parsed_url.host + "\r\n");
+      connection.BlockingWrite(
+          request_method_ + ' ' + parsed_url.path + parsed_url.ComposeParameters() + " HTTP/1.1\r\n", true);
+      connection.BlockingWrite("Host: " + parsed_url.host + "\r\n", true);
       if (!request_user_agent_.empty()) {
-        connection.BlockingWrite("User-Agent: " + request_user_agent_ + "\r\n");
+        connection.BlockingWrite("User-Agent: " + request_user_agent_ + "\r\n", true);
       }
       if (!request_body_content_type_.empty()) {
-        connection.BlockingWrite("Content-Type: " + request_body_content_type_ + "\r\n");
+        connection.BlockingWrite("Content-Type: " + request_body_content_type_ + "\r\n", true);
       }
       if (request_has_body_) {
-        connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n");
-        connection.BlockingWrite("\r\n");
-        connection.BlockingWrite(request_body_contents_);
+        connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n",
+                                 true);
+        connection.BlockingWrite("\r\n", true);
+        connection.BlockingWrite(request_body_contents_, false);
       } else {
-        connection.BlockingWrite("\r\n");
+        connection.BlockingWrite("\r\n", false);
       }
       // Attention! Achtung! Увага! Внимание!
       // Calling SendEOF() (which is ::shutdown(socket, SHUT_WR);) results in slowly sent data
--- a/net/api/impl/posix_client.h
+++ b/net/api/impl/posix_client.h
@@ -87,11 +87,6 @@ class HTTPClientPOSIX final {
       } else {
         connection.BlockingWrite("\r\n", false);
       }
-      // Attention! Achtung! Увага! Внимание!
-      // Calling SendEOF() (which is ::shutdown(socket, SHUT_WR);) results in slowly sent data
-      // not being received. Tested on local and remote data with "chunked" transfer encoding.
-      // Don't uncomment the next line!
-      // connection.SendEOF();
       http_request_.reset(new HTTPRedirectableRequestData(connection));
       // TODO(dkorolev): Rename `Path()`, it's only called so now because of HTTP request/response format.
       // Elaboration:
--- a/net/tcp/test.cc
+++ b/net/tcp/test.cc
@@ -32,6 +32,7 @@ SOFTWARE.
 #include "../../dflags/dflags.h"
 
 #include "../../strings/printf.h"
+#include "../../util/singleton.h"
 
 #include "../../3party/gtest/gtest-main-with-dflags.h"
 
@@ -47,6 +48,7 @@ using std::this_thread::sleep_for;
 using std::chrono::milliseconds;
 
 using bricks::strings::Printf;
+using bricks::Singleton;
 using bricks::net::Socket;
 using bricks::net::Connection;
 using bricks::net::ClientSocket;
@@ -223,8 +225,17 @@ TEST(TCPTest, EchoLongMessageTestsDynamicBufferGrowth) {
   ExpectFromSocket("ECHO: " + message, server_thread, message);
 }
 
+// Don't run this ~1s test more than once in a loop.
+struct OnlyCheckForUnresolvableURLOnceSingleton {
+  bool done = false;
+};
+
 TEST(TCPTest, ResolveAddressException) {
-  ASSERT_THROW(Connection(ClientSocket("999.999.999.999", 80)), SocketResolveAddressException);
+  bool& done = Singleton<OnlyCheckForUnresolvableURLOnceSingleton>().done;
+  if (!done) {
+    ASSERT_THROW(Connection(ClientSocket("999.999.999.999", 80)), SocketResolveAddressException);
+    done = true;
+  }
 }
 
 #ifndef BRICKS_WINDOWS
--- a/net/tcp/test.cc
+++ b/net/tcp/test.cc
@@ -78,7 +78,7 @@ static void ExpectFromSocket(const std::string& golden,
                              const string& message_to_send_from_client = "") {
   ExpectFromSocket(golden, server_thread, host, port, [&message_to_send_from_client](Connection& connection) {
     if (!message_to_send_from_client.empty()) {
-      connection.BlockingWrite(message_to_send_from_client);
+      connection.BlockingWrite(message_to_send_from_client, true);
     }
   });
 }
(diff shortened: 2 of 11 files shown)


deathbaba:
Fast ports opening/closing can lead to "no more free ports available" situation. Which iteration throws? Is it >64000?

net/api/impl/posix_client.h
```diff
@@ -70,21 +70,22 @@ class HTTPClientPOSIX final {
...
+        connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n",
+                                 true);
+        connection.BlockingWrite("\r\n", true);
+        connection.BlockingWrite(request_body_contents_, false);
       } else {
-        connection.BlockingWrite("\r\n");
+        connection.BlockingWrite("\r\n", false);
       }
       // Attention! Achtung! Увага! Внимание!
       // Calling SendEOF() (which is ::shutdown(socket, SHUT_WR);) results in slowly sent data
```
sompylasar: Haven't you removed `SendEOF` completely?

net/tcp/test.cc
```diff
@@ -96,34 +98,75 @@ static void ExpectFromSocket(const std::string& golden,
...
+  server.join();
+}
+
 TEST(TCPTest, CanNotUseMovedAwayConnection) {
   thread server([](Socket socket) {
                   Connection connection = socket.Accept();
-                  char buffer[3];  // Wait for three incoming bytes before sending data out.
-                  connection.BlockingRead(buffer, 3, Connection::FillFullBuffer);
-                  connection.BlockingWrite("OK");
+                  char buffer[2];  // Wait for three incoming bytes before sending data out.
```
sompylasar: There are two bytes now, not three.

sompylasar:
@dkorolev No problem. The previous version of the code was more suitable for the JS rewrite because it managed the buffers explicitly, not delegated to system calls, so I will keep it (works already).

Questions arised: 
* What will happen if the user of the API won't call `BlockingWrite(..., true)` by the end of his handler?
* What will happen if called `BlockingWrite(..., true)` and then `BlockingWrite(..., false)`?

The `more` flag passed to `BlockingWrite` looks very sensitive from the API point of view. From my Node.js experience, the response could be either written to (`write`) or ended (`end`) with optional writing before ending. Maybe have two methods, `BlockingWrite` and `End`?
```cpp
r.BlockingWrite("ABC"); //< Always expecting more data.
r.BlockingWrite("ABC");
r.End(); //< or r.End("Final words.");
```

dkorolev:
No, often times within first ten.

On Friday, February 27, 2015, Alex Zolotarev <notifications@github.com>
wrote:

> Fast ports opening/closing can lead to "no more free ports available"
> situation. Which iteration throws? Is it >64000?
>
>
>
> > On 27.2.2015, at 13.55, Dima <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
> >
> > Hi!
> >
> > First off, I'm officially an underqualified network programmer. The
> problem was solved by telling ::send() that more data is there to come, by
> passing MSG_MORE as flags when there is more data to come. I have made the
> necessary changes; please take a look on whether I missed any of those
> and/or did them wrong.
> >
> > Also, MSG_WAITALL for ::recv() does (most of) the trick with synchronous
> receive. I've made use of it too. So, two network programming blunders. So
> much for good architecture. :-(
> >
> > Now, state of the art.
> >
> > net/tcp and net/api are stable.
> >
> > I've run them with --gtest_repeat=10000 --gtest_break_on_failure both on
> Linux (-O3 and -g) and on Windows (x64/Debug and `x64/Release).
> >
> > (Although with the most recent change, --gtest_break_on_failure defaults
> to true, and is thus optional.)
> >
> > I even ran them with --gtest_filter=*Smoke* / --gtest_filter=*Smoke*, to
> only run the simplest and fastest tests (fast cycles have been showing
> random issues before on 1'000+ iterations. Btw, on Linux Bricks is showing
> >1K QPS on basic tcp/http/api tests, even on the ones parsing the URL.)
> >
> > The het/http test is stable on Linux, with all the above statements
> about stress testing being true.
> >
> > On Windows though, the net/http test can be made fail. It happens only
> in Release mode, and only when running the binary outside the IDE; in other
> words, it's not debuggable. It is, however, consistently repeatable; the
> following command does it: Bricks\VisualStudioSolution\x64\Release>
> .\net_http.exe --gtest_repeat=1000 --gtest_filter=*Smoke. On my desktop
> this command fails within a second; sometimes on the 2nd or 3rd, sometimes
> on 100++-th iteration, but still within a second.
> >
> > I have narrowed the issue to the client thread not being able to read
> the expected response from the socket. The exception is either
> SocketReadException or ConnectionResetByPeer, with no apparent pattern.
> >
> > With respect to this problem, I am close to giving up since a) it is
> low-pri, b) it is a bit too deep, and c) interestingly, all net/api tests,
> which are what our users would use, pass like a Swiss watch.
> >
> > One reasonable explanation may be that net/api opens the listener socket
> on the specified port once per binary, while net/http tests re-open it for
> each test. Perhaps some form of flush() w.r.t. waiting for internal buffers
> per port to get released, might do this trick. Another point to support
> this theory it that net/api perform much better QPS-wise than the ones in
> net/tcp or net/http, despite the latter ones being more trivial.
> >
> > (A relatively simple test might me to run WSAStartup()/WSAShutdown() per
> each test, not once per binary. But it's 3am.)
> >
> > @deathbaba : Please review. Thanks!
> >
> > @mzhurovich : Thanks for volunteering to look into! An extra pair of
> C++-savvy eyes certainly won't hurt
> >
> > @sompylasar : Sorry man, I've changed Blocking{Read/Write} once again.
> Hope it doesn't affect your REPL code much; in my defense, the changes are
> probably either irrelevant to your code or making it simpler.
> >
> > You can view, comment on, or merge this pull request online at:
> >
> > https://github.com/KnowSheet/Bricks/pull/102
> >
> > Commit Summary
> >
> > Fighting failing HTTP test on Windows.
> > rm zzz_full_test/coverage0.info
> > net/{tcp+api} OK on Windows, net/http still flaky.
> > make indent from Linux.
> > Diagnosing the test that fails on Windows.
> > Got rid of 'atomic_bool' signaling in the test.
> > Merge remote-tracking branch 'origin/t1' into t1
> > make indent
> > Minor beautification.
> > Minor style nit.
> > Dummy merge due to GitHub down for a sec.
> > 100% test coverage.
> > Fixed messed up true/false in more/done.
> > Resolved 200ms+ TODO-s.
> > Defaulting --gtest_break_on_failure to 'true'.
> > File Changes
> >
> > M 3party/gtest/gtest-main-with-dflags.h (3)
> > M 3party/gtest/gtest-main.h (3)
> > M net/api/impl/posix_client.h (19)
> > M net/api/impl/posix_server.h (2)
> > M net/api/test.cc (31)
> > M net/debug_log.h (2)
> > M net/http/impl/server.h (23)
> > M net/http/test.cc (419)
> > M net/tcp/impl/posix.h (149)
> > M net/tcp/test.cc (94)
> > Patch Links:
> >
> > https://github.com/KnowSheet/Bricks/pull/102.patch
> > https://github.com/KnowSheet/Bricks/pull/102.diff
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> <https://github.com/KnowSheet/Bricks/pull/102#issuecomment-76395275>.
>


-- 
Thanks,
Dima

dkorolev:
@sompylasar : Good catch, fixed the mentions of `EOF`.

Regarding the `more` flag (I know it's confusing; I got it wrong myself in `net/http/test.cc`, and you also confused `true` and `false` in your comment.

1. As @mzhurovich pointed out, Berkeley sockets don't have that flag, and Mac uses Berkeley sockets. Still investigating.
2. The `MSG_MORE` part is very interesting.

I googled it around. Page 1260 of http://personal.denison.edu/~bressoud/cs375-s13/supplements/linux_sockets_supplement.pdf indicates that no data will be sent if MSG_MORE is set, with the data being preserved until the consecutive call. Essentially, it's a per-call `TCP_CORK` functionality (`TCP_NOPUSH` in FreeBSD).

This, actually, makes it a mistake to use `MSG_MORE` in the middle of a chunked response.

More info (http://yarchive.net/comp/linux/sendfile.html):

```
Date: 	Thu, 18 Jan 2001 08:49:38 -0800 (PST)

MSG_MORE is very different from TCP_NOPUSH, which is very different from
TCP_CORK.

First off, the interfaces are very different. MSG_MORE is a "this write
will be followed by more writes", and only works on programs that know
that they are writing to a socket.

That has its advantages: it's a very local thing, and doesn't need any
state. However, the fact is that you _need_ the persistency of a socket
option if you want to take advantage of external programs etc getting good
behaviour without having to know that they are talking to a socket. 

Remember the UNIX philosophy: everything is a file. MSG_MORE completely
breaks that, because the only way to use it is with send[msg](). It's
absolutely unusable with something like a traditional UNIX "anonymous"
application that doesn't know or care that it's writing to the network.

So while MSG_MORE has uses, it's absolutely and utterly wrong to say that
it is equivalent to either TCP_NOPUSH or TCP_CORK.

Now, I'll agree that TCP_NOPUSH actually has the same _logic_ as MSG_MORE:
you can basically say that the two are more or less equivalent by a source
transformation (ie send(MSG_MORE) => "set TCP_NOPUSH + send() + clear
TCP_NOPUSH". Both of them are really fairly "local", but TCP_NOPUSH has a
_notion_ of persistency that is entirely lacking in MSG_MORE.

In contrast, TCP_CORK has an interface much like TCP_NOPUSH, along with
the notion of persistency. The difference between those two is that
TCP_CORK really took the notion of persistency to the end, and made
uncorking actually say "Ok, no more packets". You can't do that with
TCP_NOPUSH: with TCP_NOPUSH you basically have to know what your last
write is, and clear the bit _before_ that write if you want to avoid bad
latencies (alternatively, you can just close the socket, which works
equally well, and was probably the designed interface for the thing. That
has the disadvantage of, well, closing the socket - so it doesn't work if
you don't _know_ whether you'd write more or not).

So the three are absolutely not equivalent. I personally think that
TCP_NOPUSH is always the wrong thing - it has the persistency without the
ability to shut it off gracefully after the fact. In contrast, both
MSG_MORE and TCP_CORK have well-defined behaviour but they have very
different uses.

		Linus

```

**NONETHELESS**: I just tested on Linux that removing the `MSG_MORE` flag **does** make the issue re-appear:

```
dima@dima-i9:~/github/dkorolev/Bricks/net/http$ git diff ../tcp/impl/posix.h
...
@@ -178,7 +178,7 @@ class Connection : public SocketHandle {
-        static_cast<int>(::send(socket, buffer, write_length, MSG_NOSIGNAL | (more ? MSG_MORE : 0)));
+        static_cast<int>(::send(socket, buffer, write_length, MSG_NOSIGNAL)); // | (more ? MSG_MORE : 0)));
...
dima@dima-i9:~/github/dkorolev/Bricks/net/http$ make -B .noshit/test && time ./.noshit/test --gtest_repeat=10000 --gtest_filter=*Smoke 
mkdir -p .noshit
g++ -std=c++11 -Wall -W -g -o .noshit/test test.cc -pthread
In file included from impl/../../tcp/tcp.h:31:0,
                 from impl/server.h:40,
                 from http.h:35,
                 from test.cc:29:
impl/../../tcp/impl/posix.h:276:22: warning: unused parameter ‘more’ [-Wunused-parameter]
                      ^

Repeating all tests (iteration 1) . . .

Note: Google Test filter = *Smoke
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from PosixHTTPServerTest
[ RUN      ] PosixHTTPServerTest.Smoke
[       OK ] PosixHTTPServerTest.Smoke (2 ms)
[----------] 1 test from PosixHTTPServerTest (2 ms total)

[==========] 1 test from 1 test case ran. (2 ms total)
[  PASSED  ] 1 test.

Repeating all tests (iteration 2) . . .

Note: Google Test filter = *Smoke
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from PosixHTTPServerTest
[ RUN      ] PosixHTTPServerTest.Smoke
[       OK ] PosixHTTPServerTest.Smoke (0 ms)
[----------] 1 test from PosixHTTPServerTest (0 ms total)

[==========] 1 test from 1 test case ran. (1 ms total)
[  PASSED  ] 1 test.

Repeating all tests (iteration 3) . . .

Note: Google Test filter = *Smoke
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from PosixHTTPServerTest
[ RUN      ] PosixHTTPServerTest.Smoke
[       OK ] PosixHTTPServerTest.Smoke (0 ms)
[----------] 1 test from PosixHTTPServerTest (0 ms total)

[==========] 1 test from 1 test case ran. (0 ms total)
[  PASSED  ] 1 test.

Repeating all tests (iteration 4) . . .

Note: Google Test filter = *Smoke
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from PosixHTTPServerTest
[ RUN      ] PosixHTTPServerTest.Smoke
test.cc:67: Failure
Value of: false
  Actual: false
Expected: true
impl/../../tcp/impl/posix.h:256	SocketReadException()	
Segmentation fault (core dumped)

real	0m0.088s
user	0m0.008s
sys	0m0.004s
```

@sompylasar : Fixed your comments.
Everyone: PTAL.

Thanks,
Dima

dkorolev:
@sompylasar : And, on a final note: While the idea with `.End()` looks cute, as I understand `C`-level API, it won't work as simply here as in `node.js`. Sorry.

Guys, let's keep this issue in the state we understand it now and move on. Just need to make sure it runs on Mac. As long as `net/tcp` and `net/api` tests are solidly passing, I don't see a reason for this to block us from moving forward.

dkorolev:
To clarify: By no means I claim that `MSG_MORE` fixes the problem; rather, it masks it on the `net/http` side.

The problem, as I understand it, is that no data is being sent in response from the server to the client. While I do not have a plausible explanation, it does not mean we should keep `MSG_MORE` in the code; I'd rather `DISABLE_*` the `net/http` tests and move on without this "fix".

Thanks,
Dima

dkorolev:
Alright, I'm on the phone with @mzhurovich, will `#ifdef` out `MSG_MORE` on Apple and merge this change in since otherwise it would start blocking us.

dkorolev:
Merged. @deathbaba : Please post-review.

deathbaba:
LGTM if not count strange Windows behavior.

PR #106, opened 2015-03-03
Minor warnings cleanup on Apple.

dkorolev:

PR #108, opened 2015-03-03
Mac full build

dkorolev:


dkorolev:
CC @mzhurovich @deathbaba : Mac and Ubuntu pass `make check` and (almost) pass `make` from the top-level repository directory.

PTAL.

KnowSheet/scripts/check-headers.sh
```diff
@@ -10,9 +10,16 @@ set -u -e
...
 
+if [ $(uname) = "Darwin" ] ; then
+  CPPFLAGS+=" -stdlib=libc++ -x objective-c++ -fobjc-arc"
+  LDFLAGS+=" -framework Foundation"
+fi
+
 # NOTE: TMP_DIR must be resolved from the current working directory.
 
 TMP_DIR_NAME=".noshit"
+TMP_STDOUT=$TMP_DIR_NAME/stdout.log
```
sompylasar: Quotes please.

dkorolev: Done.

dkorolev:
Added the quotes, thanks @sompylasar. Green light to merge from my side.

sompylasar:
@dkorolev Just ran this on Mac OS X 10.9.2. All tests pass, the output is clean except the thread_local warning. Got a question from the OS to allow accepting network connections for the `test` app, but this didn't influence the test results.

time/chrono.h
```diff
@@ -65,7 +65,8 @@ inline EPOCH_MILLISECONDS Now() {
 }
 
 #else
-#warning 'thread_local' is not supported by compiler
+
+#warning "Falling back to a naive `Now()`, since `thread_local is not supported by the compiler."
```
sompylasar: Just noticed missing backtick after `thread_local`

sompylasar:
And what to do with #106 ?

mzhurovich:
@dkorolev LGTM. Btw, `docu_04graph_05gnuplot.cc` fails on my side due to the difference in gnuplot version. Maybe temporary disable this test until we find a proper way to implement in a version-independent way?

KnowSheet/scripts/check-headers.sh
```diff
@@ -10,9 +10,16 @@ set -u -e
 CPPFLAGS="-std=c++11 -g -Wall -W -DBRICKS_CHECK_HEADERS_MODE"
 LDFLAGS="-pthread"
 
+if [ $(uname) = "Darwin" ] ; then
+  CPPFLAGS+=" -stdlib=libc++ -x objective-c++ -fobjc-arc"
```
deathbaba: Is -stdlib=libc++ a linker flag?
Why do you need it at all?

graph/docu/docu_04graph_02plotutils.cc
```diff
@@ -67,7 +67,9 @@ for (size_t e = 0; e < 2; ++e) {
     .OutputFormat("svg");
 #endif
 if (FLAGS_regenerate_golden_graphs) bricks::FileSystem::WriteStringToFile(result, (std::string("golden/love.") + extensions[e]).c_str());
+#ifndef BRICKS_APPLE // TODO(dkorolev): Figure out how to run this test on Apple.
 if (!e) ASSERT_EQ(result, bricks::FileSystem::ReadFileAsString(std::string("golden/love.") + extensions[e]));
```
deathbaba: What is wrong with ReadFileAsString on Apple?

deathbaba: Aha, you need some console tool installed, right?
Which exactly binary should be called? I suppose it can be easily installed via mac port or brew.

net/tcp/impl/posix.h
```diff
@@ -274,6 +275,9 @@ class Connection : public SocketHandle {
   }
 
   inline Connection& BlockingWrite(const void* buffer, size_t write_length, bool more) {
+#ifdef BRICKS_APPLE
+    static_cast<void>(more);  // Supress the 'unused parameter' warning on Apple.
```
deathbaba: I would make it without any ifdefs - looks like it is not used not only in BRICKS_APPLE case.

KnowSheet/scripts/check-headers.sh
mzhurovich: To maximize support of c++11 on Mac. In particular, `thread_local` specifier can be used if clang is installed from macports (unfortunately default XCode's clang still doesn't support it)
http://stackoverflow.com/a/23850891

deathbaba: This is bad idea, as mobile clang compilers for iOS and Android still does not support it.
I would prefer to avoid using thread_local for at least a year.

graph/docu/docu_04graph_02plotutils.cc
sompylasar: Easier to not depend on any tool that "can be easily installed via ...".

deathbaba: If I understand correctly, it does not make any sense to use this code at all without this tool installed. Much better would be change #ifdef to instructions how to install it.

dkorolev: It's not about installing the tool or its version, it's about the inconsistency of its output binary-wise.

If only we had a simple way to do a visual diff, it would be resolved. But I am not aware of such a way for Linux+Mac+Windows.

We'll likely go with an SVG with non-image parts `grep`-ped out. Stay tuned.

And I'd say the code is fine for now. The images library is not critical for Bricks.

dkorolev:
Guys, let's merge this one and https://github.com/KnowSheet/Bricks/pull/106. They certainly improve the state of the art Linux<=>Mac compatibility wise, and are very unlikely to be introducing new bugs.

CC @deathbaba @sompylasar @mzhurovich.

Thanks,
Dima

dkorolev:
Correction: I've closed https://github.com/KnowSheet/Bricks/pull/106 without merging since this PR contains all the changes from it.

Thus, only this PR is now pending now. And I'd love to have it shipped to `upstream/master` before making further changes.

Thanks,
Dima

KnowSheet/scripts/check-headers.sh
dkorolev: @deathbaba , we are avoiding using `thread_local` as necessary.

Just printing a `#warning` in case it can be supported on Apple, but is not. We figured with Max that it would be the safest route for now.

graph/docu/docu_04graph_02plotutils.cc
dkorolev: Another solution might be to break `Bricks` into separate repositories, with `graph` not being a part of the main one. Certainly not now though.

net/tcp/impl/posix.h
dkorolev: It is used for Linux.

sompylasar:
@dkorolev There is an issue @mzhurovich pointed out: #109 , should it be taken care of?

dkorolev:
Re. https://github.com/KnowSheet/Bricks/issues/109: Yes, it's an issue to be taken care of.

It's a bit non-trivial, and, while affecting `KnowSheet`, I don't think it affects `Bricks` directly.

I'd suggest we go ahead and merge this pull request.

Thanks,
Dima

dkorolev:
And I'm merging it. Accepting comments regardless, preferably in the form of GitHub issues. Thanks! :-)

PR #118, opened 2015-03-07
A `WaitableAtomic` primitive.

dkorolev:


dkorolev:
Hi @t3nsor,

May I please ask you to take a look at this piece?

CC @sompylasar @mzhurovich @deathbaba @themoriarty.

Thanks,
Dima

waitable_atomic/waitable_atomic.h
```diff
@@ -0,0 +1,324 @@
...
+        POINTER* parent_;
+        bool mark_as_unmodified_;
+      };
+      typedef typename std::conditional<std::is_const<POINTER>::value,
+                                        ImmutableAccessorDoesNotNotify,
+                                        MutableAccessorDoesNotify>::type type;
+    };
+
+    template <class PARENT>
+    class TemplatedScopedAccessor : private ScopedUniqueLock, public NotifyIfMutable<PARENT>::type {
```
themoriarty: the class (and especially all the places where it's used) is hard to follow, since it directly access some parents fields. if it accepted not only ```parent```, but ```mutex``` and ```pdata``` as constructor arguments, it'd be easier to read

waitable_atomic/waitable_atomic.h
```diff
@@ -0,0 +1,324 @@
...
+      auto scope = ImmutableScopedAccessor();
+      f(*scope);
+    }
+
+    inline void MutableUse(std::function<void(T_DATA&)> f) {
+      auto scope = MutableScopedAccessor();
+      f(*scope);
+    }
+
+    inline void PotentiallyMutableUse(std::function<bool(T_DATA&)> f) {
```
themoriarty: huh, that's an easy way out and easy way to shoot yourself. never trust your fellow programmer - something along the following lines would be safer:
```c++
void PotentiallyMutableUse(std::function<void(const T_DATA&, std::function<MutableAccessor()>)>){};
blah.PotentiallyMutableUse([](data, upgradeFn) {
   if (data.isStale()){
     auto& mutableData = *(upgradeFn());
     mutableData.moveYourAss();
   }
})
```

waitable_atomic/waitable_atomic.h
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    virtual bool IsNotDestructing() const override {
+      std::lock_guard<std::mutex> guard(BasicImpl::data_mutex_);
+      return not_destructing_;
+    }
+
+   protected:
+    size_t ref_count_ = 0;
+    bool not_destructing_ = true;
```
themoriarty: minor minor: ```destructing``` is easier to follow than ```not_destructing```.

waitable_atomic/waitable_atomic.h
```diff
@@ -0,0 +1,324 @@
...
+        BasicImpl::Notify();
+      }
+      RefCounterDecrease();
+      if (destructor_ptr_) {
+        destructor_ptr_->WaitableAtomicDestructing();
+      }
+      ref_count_is_nonzero_mutex_.lock();
+    }
+
+    inline bool Wait(std::function<bool(const T_DATA&)> predicate) const {
```
themoriarty: unsafe function. by checking ```if (!not_destructing_)``` you imply that it's possible that the main thread that owns AtomicWaitable attempts to destroy that AtomicWaitable, while thread B is doing Wait() on it. but I don't see anything stopping the main thread from destroying that main object. I should only be able to call Wait() after I registered a scoped client (move it to the client?)

themoriarty:
it **seems** ok although there's a lot of multithreading code and it has a tendency to bite you from totally unexpected places. 
wondering would it be easier if you had standard ```intrusive_ptr<>```. you could have very basic MutableAtomic, intursive_ptr<> would replace your clients, and all that would be left is business logic (Wait(), ImmutableUse, MutableUse).

waitable_atomic/waitable_atomic.h
dkorolev: I started re-implementing it only to realize that it makes `MutableUse()` and `ImmutableUse()` unnecessarily more complex.

So I renamed it into `*Impl` instead. :-)

waitable_atomic/waitable_atomic.h
dkorolev: Could you please elaborate?

waitable_atomic/waitable_atomic.h
dkorolev: Fixed!

waitable_atomic/waitable_atomic.h
dkorolev: Locking the main `data_mutex_` and checking for `!destructing_` as the first condition of the predicate should take care of that, shoudn't they?

Could you describe the case in more detail, please?

dkorolev:
@themoriarty @t3nsor : Thank you, could you please take another look?

@themoriarty, I gave `intrusive_ptr<>` a big thought a while back, and came to the conclusion that the intent of `WaitableAtomic` is eternally different from the intent of `intrusive_ptr<>`: We focus on `Wait()` being the main operation, with reference counting being a small addition.

In theory, we could make the *intrusiveness* conditional on the type of the input object (whether it itself is an `intrusive_ptr<>`). In practice though I would like the user to control whether each particular instance of `WaitableAtomic` allows intrusive access or no; not to mention that this approach is semantically cleaner for the end user.

Thanks,
Dima

dkorolev:
Alright, I'll merge it for the sake of unblocking the progress. Thanks!

PR #119, opened 2015-03-07
An `OptionallyOwned` primitive.

dkorolev:


dkorolev:
Hi @themoriarty,

May I please ask you to take a look at this piece?

CC @sompylasar @mzhurovich @t3nsor @deathbaba.

Thanks,
Dima

optionally_owned/optionally_owned.h
```diff
@@ -0,0 +1,121 @@
...
+  T* operator->() {
+    if (!valid_) {
+      throw std::logic_error("Attempted to `operator->()` an invalid OptionallyOwned.");
+    }
+    return &instance_;
+  }
+
+  bool IsValid() const { return valid_; }
+
+  bool IsDetachable() const { return unique_ptr_.get() != nullptr; }
```
themoriarty: what's the usecase for this method?
also, isn't a bit dangerous to work differently with different objects? If I want to wrap ```unique_ptr<>``` with OptionallyOwned, it's perfectly safe to let the original ```unique_ptr<>``` go out of scope (imagine a function that cooks some data, starts the thread and exits immediately). But it's not OK with stack-allocated objects, and OptionallyOwned kinda erases the difference between them for me - I might take the function from the example above, add something benign like stack-allocated std::string, and pass it to the thread as I do with every other arguments.

themoriarty:
:+1: nice abstraction, but I think it's unsafe

optionally_owned/optionally_owned.h
dkorolev: When an `OptionallyOwned<T>` is used by a thread (`std::move`-d into it), then `OptionallyOwned::IsDetachable()` indicates whether detaching this thread with respect to the lifetime scope of the "owned" object.

If the `unique_ptr` was passed in, and thus its ownership has been transferred to the `OptionallyOwned` primitive, the thread can be `.detach()`-ed, since the ending of a more global scope will not destruct the object.

If an object was passed in by a reference, then the thread can not be `.detach()`-ed and must be `.join()`-ed instead, otherwise data corruption will occur.

Perhaps some `DoesOwn()` might be a better name?

dkorolev:
@themoriarty : PTAL.

dkorolev:
I'm closing this pull request. Will either add a SFINAE-friendly version instead, or totally deprecate it.

PR #122, opened 2015-03-07
Localized a concurrency issue in strings::Printf().

dkorolev:


dkorolev:
CC @deathbaba @mzhurovich : This was a nasty bug. Alex, please update.

Issue #126, opened 2015-03-10
Consider renaming `EPOCH_MILLISECONDS` into `EpochMilliseconds`.


dkorolev:

dkorolev:
Deeply obsolete, using C++ native `std::chrono:::{micro/milli}seconds` now.

PR #127, opened 2015-03-10
Modifications for demo v1.0.

dkorolev:


dkorolev:
Assigned @deathbaba 
CC @sompylasar @mzhurovich

sompylasar:
LGTM. How about #126 ?

mq/inmemory/mq.h
```diff
@@ -25,7 +25,7 @@ class MMQ final {
   typedef MESSAGE T_MESSAGE;
 
   // Type of the processor of the entries.
-  // It should expose one method, void OnMessage(const T_MESSAGE&, size_t number_of_dropped_messages_if_any);
+  // It should expose one method, void OnMessage(T_MESSAGE&, size_t number_of_dropped_messages_if_any);
```
deathbaba: Why by non-const reference?

port.h
```diff
@@ -114,4 +115,13 @@ SOFTWARE.
 #define BRICKS_HAS_THREAD_LOCAL
 #endif
 
+// `std::make_unique` exists in C++14, but for Bricks we'd like to see it "supported" in C++11. -- D.K.
```
deathbaba: You can make clean #ifdef and define std::make_unique
```
alexz@cdn3:~$ clang++ -dM -E -x c++ /dev/null | grep __cplusplus
#define __cplusplus 199711L
alexz@cdn3:~$ clang++ -std=c++11 -dM -E -x c++ /dev/null | grep __cplusplus
#define __cplusplus 201103L
alexz@cdn3:~$ clang++ -std=c++14 -dM -E -x c++ /dev/null | grep __cplusplus
#define __cplusplus 201402L
```

time/chrono.h
```diff
@@ -28,20 +28,52 @@ SOFTWARE.
...
-
 namespace time {
 
 enum class EPOCH_MILLISECONDS : uint64_t {};
 enum class MILLISECONDS_INTERVAL : uint64_t {};
 
+}  // namespace bricks::time
+}  // namespace bricks
+
+inline bricks::time::MILLISECONDS_INTERVAL operator-(bricks::time::EPOCH_MILLISECONDS lhs,
```
deathbaba: Why do you need all this code? Why not use simple uint64_t (or it's typedef)?

mq/inmemory/mq.h
dkorolev: There is a longer comment in Sherlock.

Gist: To allow passing data between message-queue-s of polymorphic types by doing `queue2.emplace_back(unique_ptr_from_queue.release())`.

If immutable objects are passed, the job of cloning the instance of a possibly polymorphic type becomes the job of the client, which is not good.

Also, by design, the element is only added to the message queue once and only removed from the message queue once. Thus, non-immutable accessor makes more sense performance-wise.

port.h
dkorolev: True, I suggest we revisit it once C++14 starts taking over. Doesn't look important or urgent as of now.

time/chrono.h
dkorolev: Because I want to strictly separate `unit64_t`-s from `EPOCH_MILLISECOND`-s.

(Also, this pull request just rearranges sections, this is not new code.)

dkorolev:
@deathbaba, do I get it right that this pull request is good to go?

Thanks,
Dima

time/chrono.h
deathbaba: I see that it is not a new code. This question worries me till last year: why do you want to strictly separate uint64_t from EPOCH_MILLISECOND if it is still exactly the same arithmetics and context? Your approach is bloating the code and makes it hardly understandable than simple typedef uint64_t EPOCH_MILLISECONDS;

PR #141, opened 2015-03-30
Added pthread-based ThreadLocalSingleton

mzhurovich:
@dkorolev PTAL.

### DIFF
--- a/util/singleton.h
+++ b/util/singleton.h
@@ -31,7 +31,7 @@ SOFTWARE.
 #ifndef BRICKS_HAS_THREAD_LOCAL
 #include <cstdlib>
 #include <iostream>
-#include <pthread.h> // To emulate `thread_local` via `pthread_*`.
+#include <pthread.h>  // To emulate `thread_local` via `pthread_*`.
 #endif
 
 namespace bricks {
@@ -61,12 +61,12 @@ class ThreadLocalSingletonInternals final {
     static pthread_once_t key_is_initialized = PTHREAD_ONCE_INIT;
     return key_is_initialized;
   }
- 
+
   static void Deleter(void* ptr) { delete reinterpret_cast<T*>(ptr); }
 
   static void CreateKey() {
     if (pthread_key_create(&(ThreadLocalSingletonInternals::KeyStaticStorage()),
-                            ThreadLocalSingletonInternals::Deleter)) {
+                           ThreadLocalSingletonInternals::Deleter)) {
       std::cerr << "Error in `pthread_key_create()`. Terminating." << std::endl;
       exit(-1);
     }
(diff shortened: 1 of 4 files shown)


dkorolev:
The code is fantastic! LGTM++ :+1: 

@mzhurovich -- I'm tempted to just click "Merge", and will do so in the next several hours -- but if you happen to have ~10 minutes at the keyboard let me leave a few style nit comments.

Thanks,
Dima

util/singleton.h
```diff
@@ -41,6 +48,55 @@ inline T& ThreadLocalSingleton() {
   thread_local static T instance;
   return instance;
 }
+#else
+template <typename T>
+class ThreadLocalSingletonInternals final {
```
dkorolev: `ThreadLocalSingletonImpl` would be a more natural name here.

util/singleton.h
```diff
@@ -41,6 +48,55 @@ inline T& ThreadLocalSingleton() {
...
+ private:
+  static pthread_key_t& KeyStaticStorage() {
+    static pthread_key_t key;
+    return key;
+  }
+
+  static pthread_once_t& KeyIsInitialized() {
```
dkorolev: The name of this function reads as if it returns a `bool`.

I gave it a though as a reader of this code -- why not call this function some `pthread_once_t_static_storage()` and the above one `pthread_key_t_static_storage()`?

(Personally, I would be fine with either underscores of CamelCase here, it's just to avoid the "*Is*" function **ВНЕЗАПНО** returning a reference of a weird type.)

util/singleton.h
```diff
@@ -41,6 +48,55 @@ inline T& ThreadLocalSingleton() {
...
+    }
+    return ThreadLocalSingletonInternals::KeyStaticStorage();
+  }
+
+ public:
+  static T& GetInstance() {
+    pthread_key_t& key = ThreadLocalSingletonInternals::GetKey();
+    T* ptr = reinterpret_cast<T*>(pthread_getspecific(key));
+    if (!ptr) {
+      ptr = new T;
```
dkorolev: `new T()` would be a tiny bit safer.

(Although the only real usecase I can think of is that with `new T()` a call to `ThreadLocalSingleton<int>` would initialize the `int` with zero, while `new T` leaves it uninitialized.)

util/test.cc
```diff
@@ -207,6 +209,21 @@ TEST(Util, Singleton) {
   bricks::Singleton<Foo>().reset();
 }
 
+TEST(Util, ThreadLocalSingleton) {
+  struct Foo {
+    size_t bar = 0u;
+    void baz() { ++bar; }
+  };
+  const auto add = [](size_t n) {
+    for (size_t i = 0; i < n; ++i) bricks::ThreadLocalSingleton<Foo>().baz();
```
dkorolev: Could we please follow the --

```
for (...) {
  ...
}
```.

-- convention? Thanks!

File: util/test.cc [hunk 5]
```diff
@@ -207,6 +209,21 @@ TEST(Util, Singleton) {
...
+    EXPECT_EQ(n, bricks::ThreadLocalSingleton<Foo>().bar);
+  };
+  std::thread t1(add, 50000);
+  std::thread t2(add, 10);
+  t1.join();
+  t2.join();
```
dkorolev: :+1: love how neat this test is.

PR #145, opened 2015-04-07
Minor extensions to net/.

dkorolev:
Rename `"value0"` into `"data"` by default.
Exposed URL query parameters as a `map<string, string>`.

This PR goes together with https://github.com/KnowSheet/SimpleServer/pull/1.

### DIFF
--- a/net/url/url.h
+++ b/net/url/url.h
@@ -247,8 +247,11 @@ struct URLParametersExtractor {
         return default_value;
       }
     }
+    const std::map<std::string, std::string>& AsImmutableMap() const { return parameters_; }
     std::map<std::string, std::string> parameters_;
   };
+
+  const std::map<std::string, std::string>& AllQueryParameters() const { return query.AsImmutableMap(); }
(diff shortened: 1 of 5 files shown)


net/url/url.h
```diff
@@ -247,9 +247,12 @@ struct URLParametersExtractor {
         return default_value;
       }
     }
+    const std::map<std::string, std::string>& AsImmutableMap() const { return parameters_; }
     std::map<std::string, std::string> parameters_;
   };
 
+  const std::map<std::string, std::string>& AllQueryParameters() const { return query.AsImmutableMap(); }
```
mzhurovich: why not just `const_cast`?

dkorolev: `const_cast` is not ever require here.

Why not just `{ return query.parameters_; }` is a good question. Guess I wanted to play a good citizen and not access potentially private parts :)

mzhurovich:
Looks good. Merging.

PR #150, opened 2015-04-11
Fixed Cerealize file scoping + indent + minor.

dkorolev:
@mzhurovich : PTAL.

@deathbaba : FYI + warning/note: Concurrent file opening is indeed a problem. For instance, in the case fixed by this pull request, was unseen on Mac, yet showed up on Ubuntu.

Thanks,
Dima

### DIFF
--- a/cerealize/cerealize.h
+++ b/cerealize/cerealize.h
@@ -149,13 +149,11 @@ enum class CerealFormat { Default = 0, Binary = 0, JSON };
 class CerealFileAppenderBase {
  public:
   explicit CerealFileAppenderBase(const std::string& filename, bool append = true)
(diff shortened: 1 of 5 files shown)


mzhurovich:
@dkorolev Sorry that I didn't check my lastest changes in `net/` on Ubuntu, my bad :(
Merging.

dkorolev:
No worries at all!

A much more interesting question is how on Earth did concurrently open for append, and not closed, files passed the test :-) Guess we really need to invest into concurrency issue some day soon -- or make sure our modules work around it.

mzhurovich:
Why not? It's the question of the moment when stream is flushed.
But anyway, you are 100% right that we need some sustainable solution for this issue.

deathbaba:
@dkorolev concurrent file opening is a general problem, but how is it related to your code and Alohalytics?

PR #151, opened 2015-04-11
Fixed chunked transfer in HTTP server.

mzhurovich:
Small but important fix to make server more standard-compliant and to allow Mac client properly parse chunked responses (Sherlock's test failed before that).
@dkorolev PTAL


dkorolev:
LGTM++, great catch! :+1:

PR #154, opened 2015-05-03
Levenshtein distance computation.

dkorolev:
Hi @mzhurovich ,

I coded this a while back in a clean branch, hoping the trick would help making ASCII-art-based `gnuplot` tests pass.

Propose we merge it into Bricks.

PTAL.

Thanks,
Dima


strings/distance.h
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    size_t Compute() const { return Compute(la, lb); }
+  };
+  SlowImpl impl(a, b);
+  return impl.Compute();
+}
+
+// Returns the edit distance between two strings, assuming it is forbidden
+// to 'skew' the strings by more than `max_offset` apart while mathing them letter-by-letter.
```
mzhurovich: s/mathing/matching/

strings/distance.h
```diff
@@ -0,0 +1,127 @@
...
+        return static_cast<size_t>(-1);
+      }
+    }
+    size_t Compute() const { return Compute(la, lb); }
+  };
+  FastImpl impl(a, b, max_offset);
+  return impl.Compute();
+}
+
+inline size_t FastEditDistance(const std::string& a, const std::string& b, int max_offset) {
```
mzhurovich: Do we really want to expose this kind of signature in this particular way?
Accidentally passed `max_offset = -1` will result in a really huge vector.

strings/distance.h
dkorolev: :+1: fixed.

strings/distance.h
dkorolev: Good point. Really huge vector is not a problem (`std::bad_alloc` will get thrown instantly), but why not train users to use Bricks like they use STL :-) fixed.

dkorolev:
Thanks for the review, @mzhurovich, PTAL.

(Sorry, my OCD kicked in and I added a few coverage-related comments and did `make indent` in MMQ as well :-) )

Thanks!
Dima

mzhurovich:
@dkorolev LGTM :+1:
Merging.

PR #155, opened 2015-05-04
MMQ with drop-or-wait policy on overflow + tests

mzhurovich:
@dkorolev Looks like I more-or-less finished it, PTAL

CC @deathbaba: Dima said, it could be useful for you - we fixed a bug with dropping MMQ and added an option to make it non-dropping

### DIFF
--- a/mq/inmemory/mq.h
+++ b/mq/inmemory/mq.h
@@ -34,6 +34,7 @@ class MMQ final {
       : consumer_(consumer),
         circular_buffer_size_(buffer_size),
         circular_buffer_(circular_buffer_size_),
+        dropped_messages_count_(0u),
         consumer_thread_(&MMQ::ConsumerThread, this) {}
 
   // Destructor waits for the consumer thread to terminate, which implies committing all the queued messages.
@@ -51,21 +52,27 @@ class MMQ final {
   // THREAD SAFE. Blocks the calling thread for as short period of time as possible.
   void PushMessage(const T_MESSAGE& message) {
     const size_t index = PushMessageAllocate();
-    circular_buffer_[index].message_body = message;
-    PushMessageCommit(index);
+    if (index != static_cast<size_t>(-1)) {
+      circular_buffer_[index].message_body = message;
+      PushMessageCommit(index);
+    }
   }
 
   void PushMessage(T_MESSAGE&& message) {
     const size_t index = PushMessageAllocate();
-    circular_buffer_[index].message_body = std::move(message);
-    PushMessageCommit(index);
+    if (index != static_cast<size_t>(-1)) {
+      circular_buffer_[index].message_body = std::move(message);
+      PushMessageCommit(index);
+    }
   }
 
   template <typename... ARGS>
   void EmplaceMessage(ARGS&&... args) {
     const size_t index = PushMessageAllocate();
-    circular_buffer_[index].message_body = T_MESSAGE(args...);
-    PushMessageCommit(index);
+    if (index != static_cast<size_t>(-1)) {
+      circular_buffer_[index].message_body = T_MESSAGE(args...);
+      PushMessageCommit(index);
+    }
   }
 
  private:
@@ -79,38 +86,48 @@ class MMQ final {
 
   // The thread which extracts fully populated messages from the tail of the buffer and exports them.
   void ConsumerThread() {
+    // The `tail` pointer is local to the procesing thread.
+    size_t tail = 0u;
+
+    // The counter of dropped messages is updated within the first, mutex-locked, section,
+    // and then used in the second, mutex-free, one.
+    size_t actually_dropped_messages;
+
     while (true) {
-      size_t index;
-      size_t this_time_dropped_messages;
       {
         // First, get an message to export. Wait until it's finalized and ready to be exported.
         // MUTEX-LOCKED, except for the conditional variable part.
-        index = tail_;
         std::unique_lock<std::mutex> lock(mutex_);
-        if (head_ready_ == tail_) {
-          if (destructing_) {
-            return;  // LCOV_EXCL_LINE
-          }
-          condition_variable_.wait(lock, [this] { return head_ready_ != tail_ || destructing_; });
+        while (circular_buffer_[tail].status != Entry::READY) {
           if (destructing_) {
             return;
           }
+          condition_variable_.wait(
+              lock, [this, tail] { return (circular_buffer_[tail].status == Entry::READY) || destructing_; });
         }
-        this_time_dropped_messages = number_of_dropped_messages_;
-        number_of_dropped_messages_ = 0;
+        if (destructing_) {
+          return;
+        }
+        circular_buffer_[tail].status = Entry::BEING_EXPORTED;
+
+        actually_dropped_messages = dropped_messages_count_;
+        dropped_messages_count_ = 0u;
       }
 
       {
         // Then, export the message.
         // NO MUTEX REQUIRED.
-        consumer_.OnMessage(std::ref(circular_buffer_[index].message_body), this_time_dropped_messages);
+        consumer_.OnMessage(std::ref(circular_buffer_[tail].message_body), actually_dropped_messages);
       }
 
       {
         // Finally, mark the message as successfully exported.
         // MUTEX-LOCKED.
-        std::lock_guard<std::mutex> lock(mutex_);
-        Increment(tail_);
+        {
+          std::lock_guard<std::mutex> lock(mutex_);
+          circular_buffer_[tail].status = Entry::FREE;
+        }
+        Increment(tail);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
 #define BRICKS_MQ_INMEMORY_MQ_H
 
 // MMQ is an efficient in-memory FIFO buffer.
+// One of the objectives of MMQ is to minimize the time for which the message pushing thread is blocked for.
 //
-// Messages can be pushed into it via `PushMessage()`.
+// Messages can be pushed into it via thread-safe methods `PushMessage()` or `EmplaceMessage()`.
 // The consumer is run in a separate thread, and is fed one message at a time via `OnMessage()`.
-// The order of messages is preserved.
```
dkorolev: The order is still preserved for the messages we don't drop, right?

mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
...
 //
-// One of the objectives of MMQ is to minimize the time for which the message pushing thread is blocked for.
+// The buffer size, i.e. the number of the messages MMQ can hold, is defined by
+// the constructor argument `buffer_size`. For usability reasons the default
+// value for it can be set via `DEFAULT_BUFFER_SIZE` template argument.
```
dkorolev: :+1:

mq/inmemory/mq.h
dkorolev: See it down there, makes sense. Still needs a clarification comment though, here or below, that except for the multithreading lock case, the order of messages is preserved, and, specifically, the order of messages inserted from the same thread is always kept.

dkorolev: Which is all great already in the comment below. You rock! :+1:

mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
...
+//      the subseqent calls of the consumer may be passed as a second argument of `OnMessage()`.
+//   2) Block the pushing thread and wait for the next message to be consumed
+//      and free the space in the buffer.
+//      IMPORTANT NOTE: if there are several threads waiting to push the
+//      message, MMQ DOES NOT guarantee that the messages will be added in the
+//      order in which the functions were called. However, for any particular
+//      thread MMQ DOES GUARANTEE the order of the messages for the subsequent
+//      requests to push the message.
+//  This behavior of MMQ can be controlled via `DROP_ON_OVERFLOW` template
+//  argument.
```
dkorolev: Nit: A lot of these comments seem to be wrapped at 80 characters line width, is it intentional?

mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
...
 
 #include <condition_variable>
 #include <mutex>
 #include <string>
 #include <thread>
 #include <vector>
+#include <iostream>
```
dkorolev: Do we need this guy?

mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
...
+
+template <typename T_CONSUMER, typename T_MESSAGE>
```
dkorolev: No `namespace bricks {` ?

(Strangely, I see it was not here before.)

mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
...
+  return false;
+}
+
+template <typename T_CONSUMER, typename T_MESSAGE>
+constexpr auto HasSimpleOnMessage(int)
+    -> decltype(std::declval<T_CONSUMER>().OnMessage(std::move(std::declval<T_MESSAGE>())), bool()) {
+  return true;
+}
+
+template <typename T_CONSUMER, typename T_MESSAGE>
```
dkorolev: Note: Unused template parameters can have their names omitted too, ex. `template <typename, typename>`. I too prefer to keep them for readability reasons though.

mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
...
+struct ExportMessageImpl<T_CONSUMER, T_MESSAGE, false> {
+  static void OnMessage(T_CONSUMER& consumer, T_MESSAGE&& message, size_t dropped_count) {
+    consumer.OnMessage(std::move(message), dropped_count);
+  }
+};
+
+template <typename T_CONSUMER, typename T_MESSAGE>
+void ExportMessage(T_CONSUMER& consumer, T_MESSAGE&& message, size_t dropped_count) {
+  static_assert(
+      !(HasSimpleOnMessage<T_CONSUMER, T_MESSAGE>(0) && HasExtendedOnMessage<T_CONSUMER, T_MESSAGE>(0)),
```
dkorolev: Would `a != b` be a better condition than `!(a && b)`?

mq/inmemory/mq.h
```diff
@@ -2,30 +2,104 @@
...
 class MMQ final {
  public:
   // Type of entries to store, defaults to `std::string`.
   typedef MESSAGE T_MESSAGE;
 
   // Type of the processor of the entries.
-  // It should expose one method, void OnMessage(T_MESSAGE&, size_t number_of_dropped_messages_if_any);
+  // It should expose one of the methods, accepting message by const or rvalue reference:
+  //
+  //   void OnMessage(T_MESSAGE&[&], size_t number_of_dropped_messages_if_any);
```
dkorolev: I know it looks redundant, but `[const] ` would also fit the prototype. Why not show the `T_MESSAGE&&` syntax, and on the next line put `(Alternatively, void OnMessage(const T_MESSAGE&, size_t)` ?

mq/inmemory/mq.h
```diff
@@ -77,72 +164,103 @@ class MMQ final {
...
+    // and then used in the second, mutex-free, one.
+    size_t actually_dropped_messages;
+
     while (true) {
-      size_t index;
-      size_t this_time_dropped_messages;
       {
-        // First, get an message to export. Wait until it's finalized and ready to be exported.
+        // Get the next message, which is `READY` to be exported.
         // MUTEX-LOCKED, except for the conditional variable part.
```
dkorolev: My old typo: `s/conditional/condition/`. Sorry and thanks!

mq/inmemory/mq.h
```diff
@@ -77,72 +164,103 @@ class MMQ final {
...
       }
     }
   }
 
-  size_t PushMessageAllocate() {
-    // First, allocate room in the buffer for this message.
-    // Overwrite the oldest message if have to.
+  template <bool DROP = DROP_ON_OVERFLOW>
+  typename std::enable_if<DROP, size_t>::type PushMessageAllocate() {
+    // Implementation that discards message if the queue is full.
```
dkorolev: `the message` or `the incoming message`.

mq/inmemory/mq.h
dkorolev: `via` -> `via the` (says my intuition).

mq/inmemory/mq.h
```diff
@@ -77,72 +164,103 @@ class MMQ final {
...
+      return index;
+    } else {
+      // Overflow. Discarding the message.
+      return static_cast<size_t>(-1);
+    }
+  }
+
+  template <bool DROP = DROP_ON_OVERFLOW>
+  typename std::enable_if<!DROP, size_t>::type PushMessageAllocate() {
+    // Implementation that waits for empty space if the queue is full.
```
dkorolev: `for an empty space`, and I'd note that this call blocks the calling thread, potentially indefinitely.

mq/inmemory/mq.h
```diff
@@ -150,41 +268,23 @@ class MMQ final {
...
-  // Messages beyond it will be dropped (the earlier messages will be overwritten).
-  // TODO(dkorolev): Consider adding dropping policy. Often times block-waiting until the queue
-  // can accept new messages, or simply returning `false` on a new message is a safer way.
+  // Messages beyond it will be dropped.
   const size_t circular_buffer_size_;
 
-  // The `Entry` struct keeps the entries along with the flag describing whether the message is done being
-  // populated and thus is ready to be exported. The flag is neccesary, since the message at index `i+1` might
-  // chronologically get finalized before the message at index `i` does.
+  // The `Entry` struct keeps the entries along with the status of
```
dkorolev: Dangling comment, `along with the status of`. Along with its completion status?

mq/inmemory/test.cc
```diff
@@ -1,3 +1,7 @@
+/*
```
dkorolev: This header is safe to remove.

mq/inmemory/test.cc
```diff
@@ -56,40 +60,74 @@ TEST(InMemoryMQ, SmokeTest) {
...
-    }
-  };
+struct SlowConsumer {
+  std::vector<std::string> messages_;
+  std::atomic_size_t dropped_messages_;
+  std::atomic_size_t processed_messages_;
+  size_t processing_time_ms_;
+  SlowConsumer(size_t processing_time_ms = 20u)
+      : dropped_messages_(0u), processed_messages_(0u), processing_time_ms_(processing_time_ms) {}
+  // Uncomment the line below to see that it doesn't compile.
```
dkorolev: :+1:

mq/inmemory/test.cc
```diff
@@ -56,40 +60,74 @@ TEST(InMemoryMQ, SmokeTest) {
...
+  // void OnMessage(const std::string& s) {}
+  void OnMessage(const std::string& s, size_t dropped_messages) {
+    messages_.push_back(s);
+    dropped_messages_ += dropped_messages;
+    // Simulate message processing time of 20 ms.
```
dkorolev: `of 20 ms` might not be true, perhaps just `Simulate processing delay.`

mq/inmemory/test.cc
```diff
@@ -56,40 +60,74 @@ TEST(InMemoryMQ, SmokeTest) {
...
   }
-  // Confirm that some messages were indeed dropped.
-  EXPECT_GT(c.dropped_messages_, 0u);
-  EXPECT_LT(c.dropped_messages_, 50u);
 
-  // Without the next line the test will run for 10+ms, since the remaining queue
-  // is being processed in full in the destructor of MMQ.
-  c.processing_time_ms_ = 0u;
+  // Confirm that exactly 10 messages were dropped due to a very slow consumer.
+  EXPECT_EQ(c.dropped_messages_, 10u);
```
dkorolev: I'd add a note that this condition, despite being flaky, can only fail if it takes more than 20ms to insert the messages -- which is a huge time interval.

mq/inmemory/test.cc
```diff
@@ -56,40 +60,74 @@ TEST(InMemoryMQ, SmokeTest) {
...
+}
+
+TEST(InMemoryMQ, WaitOnOverflowTest) {
+  SlowConsumer c(1u);
+
+  // Queue with 10 events in the buffer.
+  MMQ<SlowConsumer, std::string, 10, false> mmq(c);
+
+  auto producer = [&](char prefix, size_t count) {
+     for (size_t i = 0; i < count; ++i) {
```
dkorolev: indent

mq/inmemory/test.cc
```diff
@@ -56,40 +60,74 @@ TEST(InMemoryMQ, SmokeTest) {
...
+  EXPECT_EQ(c.dropped_messages_, 10u);
+}
+
+TEST(InMemoryMQ, WaitOnOverflowTest) {
+  SlowConsumer c(1u);
+
+  // Queue with 10 events in the buffer.
+  MMQ<SlowConsumer, std::string, 10, false> mmq(c);
+
+  auto producer = [&](char prefix, size_t count) {
```
dkorolev: const auto?

mq/inmemory/test.cc
```diff
@@ -56,40 +60,74 @@ TEST(InMemoryMQ, SmokeTest) {
...
+  }
+
+  for (auto& p : producers) {
+    p.join();
+  }
+
+  // Wait for 30 ms - the queue should be processed by this time.
+  std::this_thread::sleep_for(std::chrono::milliseconds(30));
+
+  // Confirm that exactly 100 messages was processed.
```
dkorolev: `s/was/were/`.

But I'd rather add an `ASSERT` that at least a few messages did get processed, then spin lock until they all did, and then confirm that the set of processed messages matches what is expected. A test that sleeps for 30ms is not good to have for many reasons.

dkorolev:
@mzhurovich : :+1:

A few minor comments, mostly on the comments and about the `sleep` in the test.

Thanks a lot, have a great night and talk soon!
Dima

mzhurovich:
@dkorolev Thanks for the comments! PTAL.

dkorolev:
Looks good to me! :+1:

mzhurovich:
@dkorolev Thanks, fixed!

dkorolev:
:+1: And this one goes live too!

PR #158, opened 2015-05-07
Added an explanatory comment.

dkorolev:
@mzhurovich, your change pinpointed an inconsistency in my previous code. Well done!

(Context: https://github.com/dkorolev/Sherlock/commit/9bff4252038473e36f20cacf315a8ebae26a2902)

### DIFF
--- a/mq/inmemory/mq.h
+++ b/mq/inmemory/mq.h
@@ -101,6 +101,13 @@ struct ExportMessageImpl<T_CONSUMER, T_MESSAGE, false> {
 
 template <typename T_CONSUMER, typename T_MESSAGE>
 void ExportMessage(T_CONSUMER& consumer, T_MESSAGE&& message, size_t dropped_count) {
+  // Note that the allowed syntax for `OnMessage()` is:
+  //   either `OnMessage(const T_MESSAGE& message [, size_t dropped_count]);`
+  //   or     `OnMessage(T_MESSAGE&& message [, size_t dropped_count]);`
+  // If your class defines `OnMessage()` differently, for example, accepting a non-const non-rvalue reference
+  // as the first parameter, the `static_assert()` on the next line will fail.
+  static_assert(HasSimpleOnMessage<T_CONSUMER, T_MESSAGE>(0) || HasExtendedOnMessage<T_CONSUMER, T_MESSAGE>(0),
+                "There must be at least one implementation of `OnMessage()` defined in the consumer.");
   static_assert(HasSimpleOnMessage<T_CONSUMER, T_MESSAGE>(0) != HasExtendedOnMessage<T_CONSUMER, T_MESSAGE>(0),
                 "There must be exactly one implementation of `OnMessage()` defined in the consumer.");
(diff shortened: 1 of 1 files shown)


mq/inmemory/mq.h
```diff
@@ -101,6 +101,13 @@ struct ExportMessageImpl<T_CONSUMER, T_MESSAGE, false> {
 
 template <typename T_CONSUMER, typename T_MESSAGE>
 void ExportMessage(T_CONSUMER& consumer, T_MESSAGE&& message, size_t dropped_count) {
+  // Note that the allowed syntax for `OnMessage()` is:
+  //   either `OnMessage(const T_MESSAGE& message [, size_t dropped_count]);`
+  //   or     `OnMessage(T_MESSAGE&& message [, size_t dropped_count]);`
```
mzhurovich: I'm still not sure about mentioning rvalue-reference argument for user function. Since we are `std:move`'ing, pass-by-value looks more natural. What do you think?

dkorolev: Accepting an object by value, when passed in by `std::move()`, would not eliminate the copy, as far as I know.

mzhurovich: Simple test:

```cpp
struct A {
  int x;
  A(const int x) : x(x) {
    std::cout << "Create A. x = " << x << std::endl;
  }

  A(const A& a) : x(a.x) {
    std::cout << "Copy" << std::endl;
  }

  A(A&& a) : x(std::move(a.x)) {
    std::cout << "Move" << std::endl;
  }
};

void foo(const A& a) {
  std::cout << "x = " << a.x << std::endl;
}

void bar(A a) {
  std::cout << "x = " << a.x << std::endl;
}

int main() {
  A a1(1);
  foo(a1);
  bar(a1);
  bar(std::move(a1));
}
```

dkorolev: It still calls the constructor. Extension :)

```c++
#include <iostream>

struct A {
  int x;
  A(const int x) : x(x) {
    std::cout << "Create A. x = " << x << std::endl;
  }

  A(const A& a) : x(a.x) {
    std::cout << "Copy" << std::endl;
  }

  A(A&& a) : x(std::move(a.x)) {
    std::cout << "Move" << std::endl;
  }
};

void foo(const A& a) {
  std::cout << "x = " << a.x << std::endl;
}

void bar(A a) {
  std::cout << "x = " << a.x << std::endl;
}

void baz(A&& a) {
  std::cout << "x = " << a.x << std::endl;
}

int main() {
  A a1(1);
  foo(a1);
  bar(a1);
  // Constructor called.
  bar(std::move(a1));
  // Constructor not called.
  baz(std::move(a1));
}
```
Create A. x = 1
x = 1
Copy
x = 1
Move
x = 1
x = 1
```

```

mzhurovich: Yep, sorry, you are right.

dkorolev: It's not me, it's C++ :+1:

PR #187, opened 2015-05-26
Added metronome to `SimpleServer`.

mzhurovich:
@dkorolev PTAL.

### DIFF
--- a/SimpleServer/log_collector.h
+++ b/SimpleServer/log_collector.h
@@ -25,9 +25,10 @@ SOFTWARE.
 #ifndef LOG_COLLECTOR_H
 #define LOG_COLLECTOR_H
 
-#include <string>
 #include <map>
 #include <ostream>
+#include <string>
+#include <thread>
 
 #include "../../Current/Bricks/net/api/api.h"
 #include "../../Current/Bricks/time/chrono.h"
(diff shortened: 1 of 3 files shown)


SimpleServer/log_collector.h
```diff
@@ -60,32 +61,70 @@ class LogCollectorHTTPServer {
...
+      const uint64_t now = static_cast<uint64_t>(bricks::time::Now());
+      const uint64_t dt = now - last_event_t_;
+      if (dt >= tick_interval_ms_) {
+        std::lock_guard<std::mutex> lock(mutex_);
+        LogEntry entry;
+        entry.t = now;
+        ostream_ << JSON(entry, "log_entry") << std::endl;
+        last_event_t_ = entry.t;
+      } else {
+        std::this_thread::sleep_for(std::chrono::milliseconds(tick_interval_ms_ - dt));
```
dkorolev: `(tick_interval_ms_ - dt + 1)`
а то может быть "дребезг", что неприятно

SimpleServer/log_collector.h
```diff
@@ -60,32 +61,70 @@ class LogCollectorHTTPServer {
...
-      entry.q = r.url.AllQueryParameters();
-      entry.b = r.body;
-      entry.f = r.url.fragment;
-      // TODO(dkorolev): HTTP headers come here.
-      // entry.h = { ... }
-      ostream_ << JSON(entry, "log_entry") << std::endl;
-      r(response_text_);
-    });
+                         const std::string& response_text = "OK\n",
+                         const size_t tick_interval_ms = 1000)  // `0` to disable tick events.
```
dkorolev: Why not flip these two parameters, so that `"OK"` can be kept default?

SimpleServer/log_collector.h
```diff
@@ -60,32 +61,70 @@ class LogCollectorHTTPServer {
...
+        ostream_(ostream),
+        route_(route),
+        response_text_(response_text),
+        tick_interval_ms_(tick_interval_ms),
+        send_ticks_(tick_interval_ms > 0),
+        timer_thread_(&LogCollectorHTTPServer::TimerThreadFunction, this) {
+    HTTP(http_port_)
+        .Register(route_, [this](Request r) {
+          LogEntry entry;
+          entry.t = static_cast<uint64_t>(bricks::time::Now());
```
dkorolev: I think the whole part between invoking `bricks::time::Now()` and doing `ostream_ <<` should be under a mutex.

Otherwise there's a nonzero chance for time going back in the output, which should not happen.

SimpleServer/log_collector.h
```diff
@@ -60,32 +61,70 @@ class LogCollectorHTTPServer {
...
+  }
 
   void Join() { HTTP(http_port_).Join(); }
 
+  void TimerThreadFunction() {
+    while (send_ticks_) {
+      const uint64_t now = static_cast<uint64_t>(bricks::time::Now());
+      const uint64_t dt = now - last_event_t_;
+      if (dt >= tick_interval_ms_) {
+        std::lock_guard<std::mutex> lock(mutex_);
```
dkorolev: Here too: pls move `bricks::time::Now()` into under the mutex.

SimpleServer/test.cc
```diff
@@ -40,27 +41,37 @@ using bricks::strings::Printf;
...
 
-  bricks::time::SetNow(static_cast<bricks::time::EPOCH_MILLISECONDS>(1234));
+  bricks::time::SetNow(static_cast<bricks::time::EPOCH_MILLISECONDS>(12));
   const auto get_response = HTTP(GET(Printf("http://localhost:%d/log", FLAGS_log_collector_test_port)));
   EXPECT_EQ(200, static_cast<int>(get_response.code));
   EXPECT_EQ("OK\n", get_response.body);
 
-  bricks::time::SetNow(static_cast<bricks::time::EPOCH_MILLISECONDS>(5678));
+  bricks::time::SetNow(static_cast<bricks::time::EPOCH_MILLISECONDS>(112));
+  std::this_thread::sleep_for(std::chrono::milliseconds(150));
```
dkorolev: I'm fine keeping this, but it's scary. Please add a comment at least.

(I'd add an `atomic_size_t` for the total number of events published into `ostream_`, and wait on its value. `#ifdef`-guarded to be test-only, if/when performance is a concern.

SimpleServer/test.cc
```diff
@@ -40,27 +41,37 @@ using bricks::strings::Printf;
...
   EXPECT_EQ(200, static_cast<int>(post_response.code));
   EXPECT_EQ("OK\n", post_response.body);
+
+  bricks::time::SetNow(static_cast<bricks::time::EPOCH_MILLISECONDS>(278));
+  std::this_thread::sleep_for(std::chrono::milliseconds(150));
+
   EXPECT_EQ(
-      "{\"log_entry\":{\"t\":1234,\"m\":\"GET\",\"u\":\"/log\",\"q\":[],\"b\":\"\",\"f\":\"\"}}\n"
-      "{\"log_entry\":{\"t\":5678,\"m\":\"POST\",\"u\":\"/log\",\"q\":[],\"b\":\"meh\",\"f\":\"\"}}\n",
+      "{\"log_entry\":{\"t\":0,\"m\":\"\",\"u\":\"\",\"q\":[],\"b\":\"\",\"f\":\"\"}}\n"
```
dkorolev: HEY! Let's have ticks use their own method, ex. `"TICK"` ? :-)

SimpleServer/log_collector.h
dkorolev: I'd go as far here as saying that `1000` should not be the default -- let's force the users to always set this value. It's performance-critical (RAM- and disk-critical) in certain occasions, so no surprises is a good thing.

And we have `bricks::time::MILLISECONDS_INTERVAL` to make it look substantially scary :-) Thanks!

dkorolev:
LGTM! :+1:

PR #189, opened 2015-05-31
Supporting `size_t` in dflags.

dkorolev:
CC @sompylasar.

Thanks,
Dima


Bricks/dflags/test.cc
```diff
@@ -135,17 +135,20 @@ TEST(DFlags, ParsesMultipleFlags) {
   DEFINE_bool(flag_bool2, true, "");
   DEFINE_int64(flag_int64, 0, "");
   DEFINE_uint64(flag_uint64, 0, "");
```
mzhurovich: `0u` here as well?

mzhurovich:
@dkorolev LGTM!

Bricks/dflags/test.cc
dkorolev: :+1: taken care of in c6e551b. PTAL!

PR #206, opened 2015-06-13
MMQ: flipped default behavior to non-dropping.

mzhurovich:
@dkorolev PTAL.


dkorolev:
:+1: Yay Max, we need more of these targeted changes from good to better! Thanks!!!

Merging.

PR #207, opened 2015-06-13
Bricks fixes for Windows.

dkorolev:
Tweaks:

* `condition_variable` instead of a mutex that was incorrectly locked and unlocked from different threads.
* Windows-friendly `strings::Printf()` and its proper usage (in two tests we were overflowing it).
* Tweak for correct file length detection, which was broken on Windows.
* Pleased MSVS by using `template<typename T>` instead of `template<typename... TS>` for `using`-s (was breaking `is_unique_ptr<>`, because of `decay<>`, because of `rmconst<>` and `rmref<>`).
* MSVC `#pragma` syntax for warnings.

@mzhurovich : It's a big enough change, let's look through it first before I start moving files around and running a Windows test outside `Bricks`.

Thanks,
Dima

### DIFF
--- a/3rdparty/gtest/docu/docu_96gtest.md
+++ b/3rdparty/gtest/docu/docu_96gtest.md
@@ -1,3 +1,5 @@
+`TODO(dkorolev): Include this file some documentation -- likely a separate .md for "3rdparty".`
+
 ## Unit Testing: `gtest`
 
 Bricks contains a header-only port of Google's [`GoogleTest`](http://code.google.com/p/googletest/): a great C++ unit testing library open-sourced by [**Google**](https://www.google.com/finance?q=GOOG).
--- a/3rdparty/gtest/docu/docu_96gtest.md
+++ b/3rdparty/gtest/docu/docu_96gtest.md
@@ -1,4 +1,4 @@
-`TODO(dkorolev): Include this file some documentation -- likely a separate .md for "3rdparty".`
+`TODO(dkorolev): Include this file in some documentation -- likely a separate .md for "3rdparty".`
 
 ## Unit Testing: `gtest`
 
--- a/Bricks/cerealize/cerealize.h
+++ b/Bricks/cerealize/cerealize.h
@@ -100,12 +100,18 @@ enum class CerealFormat { Default = 0, Binary = 0, JSON };
 // TODO(dkorolev): Think of platform-independent layer for filesystem locks.
 class CerealFileAppenderBase {
  public:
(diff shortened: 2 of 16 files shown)


3rdparty/gtest/docu/docu_96gtest.md
```diff
@@ -1,3 +1,5 @@
+`TODO(dkorolev): Include this file some documentation -- likely a separate .md for "3rdparty".`
```
mzhurovich: `Include this file some documentation` - preposition is missing?

dkorolev: :+1: se 44ca3a9

Bricks/template/decay.h
```diff
@@ -72,8 +72,8 @@ struct rmref_impl<std::tuple<TS...>&&> {
   typedef metaprogramming::map<impl, std::tuple<TS...>> type;
 };
 
-template <typename... TS>
-using rmref = typename rmref_impl<TS...>::type;
+template <typename T>
+using rmref = typename rmref_impl<T>::type;
```
mzhurovich: I'm not sure, but does variadic implementation above make sense with this change?

dkorolev: I believe so -- it only uses `<TS...>` as parts of tuples in specializations. And top-level `make test` and `make check` passed on Linux.

Bricks/cerealize/cerealize.h
```diff
@@ -100,12 +100,18 @@ enum class CerealFormat { Default = 0, Binary = 0, JSON };
...
   explicit CerealFileAppenderBase(const std::string& filename, bool append = true)
       : fo_(filename, (append ? std::ofstream::app : std::ofstream::trunc) | std::ofstream::binary),
-        initial_stream_position_(fo_.tellp()) {}
+        initial_stream_position_(static_cast<size_t>((fo_.seekp(0, std::ios_base::end), fo_.tellp()))) {}
 
   inline size_t EntriesAppended() const { return entries_appended_; }
-  inline size_t BytesAppended() const { return current_stream_position() - initial_stream_position_; }
+  inline size_t BytesAppended() const {
+    const size_t current = current_stream_position();
+    assert(current >= initial_stream_position_);  // TFU is it's the case.
```
dkorolev: This is already taken care of in 5abacfc.

Bricks/template/decay.h
mzhurovich: Yep, you're right, everything is good here.

mzhurovich:
@dkorolev LGTM!

PR #214, opened 2015-06-16
Persistence work in progress.

dkorolev:
Hi @mzhurovich ,

This is not to merge, but it should be 75+% code complete and 95%+ design complete for our persistence layer.

Comments are very welcome.

Thanks!
Dima


Bricks/mq/persistence/persistence.h
```diff
@@ -0,0 +1,152 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+// TODO(dkorolev): Sync up with Max on code locations, names, and relative paths.
+//                 I suggest we promote Yoda, MMQ (under a different name),
+//                 and Persistence (under a different name) to the top level directory.
+
+#ifndef BRICKS_MQ_INTERFACE_PERSISTENCE_H
```
sompylasar: `BRICKS_MQ_INTERFACE_PERSISTENCE_H` -> `BRICKS_MQ_PERSISTENCE_H` according to the file path.

dkorolev: :+1: fixed in 1cfa725.

Although this header will probably move around.

dkorolev:
Hi @mzhurovich 

Baby steps -- this one does persistence on the lower level. PTAL!

Thanks
Dima

Bricks/cerealize/cerealize.h
```diff
@@ -212,7 +204,7 @@ class CerealBinaryFileParser {
     if (fi_.peek() != std::char_traits<char>::eof()) {  // This is safe with any next byte in file.
       std::unique_ptr<T_ENTRY> entry;
       si_(entry);
-      processor(*entry.get());
+      processor(std::move(*entry.get()));
```
mzhurovich: I wonder, does it make sense? Trying to `std::move` const pointer.

Bricks/mq/interface/interface.h
```diff
@@ -264,6 +266,94 @@ inline bool DispatchEntryByRValue(F&& f, E&& e, size_t index, size_t total) {
...
+constexpr bool HasTerminateMethod(char) {
+  return false;
+}
+
+template <typename T>
+constexpr auto HasTerminateMethod(int) -> decltype(std::declval<T>().Terminate(), bool()) {
+  return true;
+}
+
+template <typename T, bool>
```
mzhurovich: I vote for named template parameters to increase readability of the code.

Bricks/cerealize/cerealize.h
dkorolev: It should -- the pointer is not `const`. It just makes sure if `processor()` accepts an `E&&`, then a move constructor is used (and the contents of the object stored inside the `unique_ptr<>` here is invalidated.)

Bricks/mq/interface/interface.h
dkorolev: Right, here 73b5433 is it!

Bricks/cerealize/cerealize.h
mzhurovich:
```
Notice that a call to this function does not make unique_ptr release ownership of the pointer 
(i.e., it is still responsible for deleting the managed data at some point).
Therefore, the value returned by this function shall not be used to construct a new managed pointer.
```

Bricks/mq/interface/interface.h
```diff
@@ -264,6 +266,94 @@ inline bool DispatchEntryByRValue(F&& f, E&& e, size_t index, size_t total) {
...
+struct CallTerminateImpl {
+  static bool DoIt(T) { return true; }
+};
+
+template <typename T>
+struct CallTerminateImpl<T, true> {
+  static decltype(std::declval<T>().Terminate()) DoIt(T&& ref) { return ref.Terminate(); }
+};
+
+template <typename T, bool>
```
mzhurovich: Here too, pls.

Bricks/mq/interface/interface.h
```diff
@@ -264,6 +266,94 @@ inline bool DispatchEntryByRValue(F&& f, E&& e, size_t index, size_t total) {
...
+constexpr bool HasReplayDoneMethod(char) {
+  return false;
+}
+
+template <typename T>
+constexpr auto HasReplayDoneMethod(int) -> decltype(std::declval<T>().ReplayDone(), bool()) {
+  return true;
+}
+
+template <typename T, bool>
```
mzhurovich: And here.

Bricks/mq/interface/interface.h
```diff
@@ -264,6 +266,94 @@ inline bool DispatchEntryByRValue(F&& f, E&& e, size_t index, size_t total) {
...
+
+template <typename T>
+bool CallTerminate(T&& ref) {
+  return impl::CallTerminateAndReturnBoolImpl<
+      T,
+      std::is_same<void,
+                   decltype(impl::CallTerminateImpl<T, impl::HasTerminateMethod<T>(0)>::DoIt(
+                       std::declval<T>()))>::value>::DoIt(std::forward<T>(ref));
+}
+
```
mzhurovich: You can return `void`, so, how about shortening the code above to this:
```
template <typename T, bool HAS_TERMINATE_METHOD>
struct CallTerminateImpl {
  static bool DoIt(T) { return true; }
};

template <typename T>
struct CallTerminateImpl<T, true> {
  static auto DoIt(T&& ref) -> decltype(std::declval<T>().Terminate()) { return ref.Terminate(); }
};

template <typename T>
auto CallTerminate(T&& ref)
    -> decltype(impl::CallTerminateImpl<T, impl::HasTerminateMethod<T>(0)>::DoIt(std::declval<T>())) {
  return impl::CallTerminateImpl<T, impl::HasTerminateMethod<T>(0)>::DoIt(std::forward<T>(ref));
}
```

mzhurovich: Ah, damn, you can't use it further in conditional statements :(
Disregard this comment.

Bricks/util/clone.h
```diff
@@ -0,0 +1,114 @@
...
+
+#include <functional>
+
+#include "../cerealize/cerealize.h"
+
+namespace bricks {
+
+namespace impl {
+
+template <typename T, bool HAS_CLONEBY_REF, bool HAS_CLONE_BY_PTR, bool CAN_BE_COPIED>
```
mzhurovich: s/HAS_CLONEBY_REF/HAS_CLONE_BY_REF/

Bricks/util/clone.h
```diff
@@ -0,0 +1,114 @@
...
+  return false;
+}
+
+template <typename T>
+constexpr auto HasCloneByPtr(int) -> decltype(std::declval<const T&>() -> Clone(), bool()) {
+  return true;
+}
+
+template <typename T>
+constexpr bool CanBeCopied(char) {
```
mzhurovich: Maybe `CanBeCopied` -> `HasCopyC[onstruc]tor`?

Bricks/mq/persistence/persistence.h
```diff
@@ -0,0 +1,205 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+// TODO(dkorolev): Sync up with Max on code locations, names, and relative paths.
+//                 I suggest we promote Yoda, MMQ (under a different name),
+//                 and Persistence (under a different name) to the top level directory.
```
mzhurovich: As this implementation combines in-memory and file storage approaches, we definitely need to rename it.
Let's discuss it by phone?

Bricks/mq/persistence/persistence.h
```diff
@@ -0,0 +1,205 @@
...
+        }
+        next.total = exclusively_accessed_list.size();
+        next.at_end = (next.iterator == exclusively_accessed_list.end());
+        return next;
+      }
+    };
+    Cursor current;
+
+    const size_t size_at_start = [this]() {
+      // LOCKED: Move the cursor forward.
```
mzhurovich: Does this lambda really move it?

Bricks/cerealize/cerealize.h
dkorolev: True -- but it's still not `const`.

`std::move()` here is an indicator that the client is free to invalidate the object passed to it.

Taking away ownership is not the same as destructing. For instance, passing a `unique_ptr` to a polymorphic type by an rvalue reference enables `std::move()`-ing it further without copying. The destructor of the original `unique_ptr<>` defined two lines above will still be called, as it should to ensure no memory is leaked.

Bricks/mq/interface/interface.h
dkorolev: Good catch, 5f21cde.

dkorolev: Good catch, fixed in 5f21cde22a8e1402ce477629e381465c250e28de.

Bricks/mq/interface/interface.h
dkorolev: Done in 5f21cde22a8e1402ce477629e381465c250e28de.

Bricks/mq/persistence/persistence.h
dkorolev: Yes! File naming, as well as directories, is the part I have not touched yet.

Bricks/mq/persistence/persistence.h
dkorolev: My bad, outdated comment. Fixed in bd0dd56.

Bricks/util/clone.h
dkorolev: Good catch! :+1: bd0dd56

Bricks/util/clone.h
dkorolev: Yep, great suggestion. Fixed in bd0dd56

dkorolev:
@mzhurovich : Thanks for looking through. Fixed your comments. Should be ready to discuss design and naming in 1+ hours.

Bricks/cerealize/cerealize.h
mzhurovich: Ok, my bad, it's not `const`.
But this still looks like hack for me, I understand `std::move`'ing `unique_ptr`'s, but not their internal pointers.

dkorolev: Yes, it doesn't look perfect to me either. But it's legal, valid C++, and I don't know of a way that does the same while ensuring no memory will be leaked in any case.

Bricks/template/is_unique_ptr.h
```diff
@@ -56,6 +56,41 @@ static_assert(std::is_same<int, typename is_unique_ptr<int>::underlying_type>::v
...
 
+template <typename B, typename E>
+struct can_be_stored_in_unique_ptr {
+  static constexpr bool value = false;
+};
+
+// Note: No custom deleters here [yet]. TODO(dkorolev): Make sure they are here as we need them.
+template <typename B, typename E>
+struct can_be_stored_in_unique_ptr<std::unique_ptr<B>, E> {
+  static constexpr bool value = std::is_same<B, E>::value || std::is_base_of<B, E>::value;
```
mzhurovich: `std::is_base`: If Derived is derived from Base or if both are the same non-union class, provides the member constant value equal to true. Otherwise value is false.
So, `is_same` is redundant.

dkorolev:
Fixed in 5e9a930.

dkorolev:
@mzhurovich -- I've added another commit to add `Response` into this branch.

# And `CTFO_BE` tests pass!

The issue was in `Sherlock`'s listener code, that currently -- in order to not create a copy of the event -- blocks the whole collection (which is an `std::vector`) there for the period of processing the entry through a listener.

Here's the code that is now live in `C5T`: https://github.com/c5t/Current/blob/77f12ad835e517bd2a683ac29a00d297703d912b/Sherlock/sherlock.h#L460-L474

```c++
          blob->data.ImmutableUse([&blob, &cursor, &user_initiated_terminate](const std::vector<T>& data) {

            assert(cursor < data.size());

            // The below implementation for entry cloning is imperfect, but it serves the purpose semantically.
            // TODO(dkorolev): Fix it.
            const std::function<T(const T&)> clone = [](const T& e) { return ParseJSON<T>(JSON(e)); };

            using bricks::mq::DispatchEntryByConstReference;
            if (!DispatchEntryByConstReference(*blob->listener, data[cursor], cursor, data.size(), clone)) {
              user_initiated_terminate = true;
            }

            ++cursor;
          });
```

Note that `DispatchEntryByConstReference()` is called from within `ImmutableUse()`, this keeping the stream locked for the whole time one listener is processing the entry.

# Now, good news!

This is exactly the issue I fixed first when working on persistence and unifying the publisher/listener interfaces. The new stream is using `std::list<>` internally, which does not invalidate iterators, thus allowing to both eliminate the copy/clone of the entry to be processed by a particular listener thread and not lock the whole stream in the meantime.

The new code is in this pull request, which we decided not to merge. I'll make sure to factor it out and send a polished pull request later today -- ensuring `CTFO_BE` tests pass in the meantime.

Perfect timing :+1: !

(Yes, `std::list<>` is not the most efficient solution per se, since it's a linked list with one pointer per entry. We might want to optimize it later to some list that stores blobs of pre-allocated entry buffers. But it's certainly not an action item for now.)

Thus, I just added `Response` to this `persistence_snapshot` branch to confirm `CTFO_BE` passes the test.

```bash

dima@Users-MacBook-Pro /Users/dima/github/mzhurovich $ unlink Current
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich $ ln -sf ../C5T/Current .
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich $ cd CTFO_BE/
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich/CTFO_BE (master) $ make clean test
rm -rf .current core impl/.current impl/core
mkdir -p .current
g++ -std=c++11 -Wall -W -stdlib=libc++ -x objective-c++ -fobjc-arc -o .current/test test.cc -pthread -framework Foundation
...
<< HANGING >>>

dima@Users-MacBook-Pro /Users/dima/github/mzhurovich/CTFO_BE (master) $ cd ..
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich $ ls -las
total 8
0 drwxr-xr-x   4 dima  staff  136 Jun 19 10:21 .
0 drwxr-xr-x  10 dima  staff  340 Jun 19 08:31 ..
0 drwxr-xr-x  17 dima  staff  578 Jun 19 10:21 CTFO_BE
8 lrwxr-xr-x   1 dima  staff   14 Jun 19 10:21 Current -> ../C5T/Current
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich $ unlink Current
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich $ ln -sf ../dkorolev/Current/ .
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich $ cd -
/Users/dima/github/mzhurovich/CTFO_BE
dima@Users-MacBook-Pro /Users/dima/github/mzhurovich/CTFO_BE (master) $ make clean test
rm -rf .current core impl/.current impl/core
mkdir -p .current
g++ -std=c++11 -Wall -W  -ftemplate-backtrace-limit=0 -stdlib=libc++ -x objective-c++ -fobjc-arc -o .current/test test.cc -pthread -framework Foundation
.current/test --bricks_runtime_arch=Darwin
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from CTFO
[ RUN      ] CTFO.SmokeTest
[       OK ] CTFO.SmokeTest (107 ms)
[----------] 1 test from CTFO (107 ms total)

[==========] 1 test from 1 test case ran. (107 ms total)
[  PASSED  ] 1 test.
```

We nailed it. Flying high.

* * *

As discussed by the phone, this pull request is not to be merged. Let me close it with this comment, although I might still add a few commits if that proves to be the easiest way for me to sync between Ubuntu and MacOS while testing.

dkorolev:
@mzhurovich And https://github.com/c5t/Current/pull/218 finalizes the deprecation of this pull request.

PR #216, opened 2015-06-18
Added `Response`.

dkorolev:
Hi @mzhurovich,

Here's a helper class to forward HTTP responses to the 2nd parameter to `Yoda` transactions, that own the HTTP request object.

PTAL.

Thanks,
Dima

### DIFF
--- a/Bricks/net/api/request.h
+++ b/Bricks/net/api/request.h
@@ -35,6 +35,17 @@ namespace bricks {
 namespace net {
 namespace api {
 
+template <typename T>
+constexpr static bool HasRespondViaHTTP(char) {
+  return false;
+}
+
+template <typename T>
+constexpr static auto HasRespondViaHTTP(int)
+    -> decltype(std::declval<T>().RespondViaHTTP(std::declval<struct Request>()), bool()) {
+  return true;
+}
+
 // The only parameter to be passed to HTTP handlers.
 struct Request final {
   std::unique_ptr<HTTPServerConnection> unique_connection;
@@ -67,20 +78,7 @@ struct Request final {
 
   // Support objects with user-defined HTTP response handlers.
   template <typename T>
-  struct HasRespondViaHTTP {
-    typedef char one;
-    typedef long two;
-
-    template <typename C>
-    static one test(decltype(&C::RespondViaHTTP));
-    template <typename C>
-    static two test(...);
-
-    constexpr static bool value = (sizeof(test<T>(0)) == sizeof(one));
-  };
-
-  template <typename T>
-  inline typename std::enable_if<HasRespondViaHTTP<bricks::decay<T>>::value>::type operator()(
+  inline typename std::enable_if<HasRespondViaHTTP<bricks::decay<T>>(0)>::type operator()(
       T&& that_dude_over_there) {
     that_dude_over_there.RespondViaHTTP(std::move(*this));
   }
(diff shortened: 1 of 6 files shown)


Bricks/net/api/response.h
```diff
@@ -0,0 +1,155 @@
...
+            const std::string& object_name,
+            HTTPResponseCodeValue code = HTTPResponseCode.OK,
+            const HTTPHeadersType& extra_headers = HTTPHeadersType()) {
+    this->body = cerealize::JSON(std::forward<T>(object), object_name) + '\n';
+    this->code = code;
+    this->content_type = HTTPServerConnection::DefaultJSONContentType();
+    this->extra_headers = extra_headers;
+  }
+
+  Response& Body(const std::string& s) {
```
sompylasar: Is there a way to construct an empty `Response` object so that its every field is filled with these chainable methods? I don't see such a way now.

mzhurovich: As I see, it's explicitly prohibited by deleting a default constructor.
Why do you think it could be useful to create `Response` without initialization?

Bricks/net/api/request.h
```diff
@@ -0,0 +1,108 @@
...
+    typedef char one;
+    typedef long two;
+
+    template <typename C>
+    static one test(decltype(&C::RespondViaHTTP));
+    template <typename C>
+    static two test(...);
+
+    constexpr static bool value = (sizeof(test<T>(0)) == sizeof(one));
+  };
```
mzhurovich: Maybe rewrite it using our usual approach?

Bricks/net/api/response.h
sompylasar: With empty (default) initialization, not without initialization. Otherwise, these setters look redundant to me as we've got every field initialized via the constructing functions above.

mzhurovich: Yep, the setters seem redundant for me too in this case.
Dima?

Bricks/net/api/request.h
dkorolev: :+1: done in 1b108e3.

Bricks/net/api/response.h
dkorolev: They were not redundant, they were just in case the user wants to change them.

Added an empty/default constructor in 1b108e3. Assignment still disabled.

dkorolev:
Great comments, @mzhurovich, @sompylasar!

Another look? :-)

Thanks,
Dima

Bricks/net/api/response.h
```diff
@@ -0,0 +1,188 @@
...
+#include "../http/http.h"
+#include "../../strings/is_string_type.h"
+#include "../../cerealize/cerealize.h"
+
+namespace bricks {
+namespace net {
+namespace api {
+
+static_assert(strings::is_string_type<const char*>::value, "");
+static_assert(!strings::is_string_type<int>::value, "");
```
mzhurovich: Move to `is_string_type.h`?

dkorolev: They are already there.

And here they are only because I was testing something. Removing.

mzhurovich:
@dkorolev LGTM!
@sompylasar Any comments before merge?

sompylasar:
@mzhurovich no more comments, LGTM.

PR #223, opened 2015-07-02
Yoda bugfix.

mzhurovich:
@dkorolev PTAL


dkorolev:
That's where it was! Thanks Max! :+1:

PR #225, opened 2015-07-22
Proper JSON top-level object naming.

dkorolev:
Let the coding begin!

### DIFF
--- a/Blocks/HTTP/test.cc
+++ b/Blocks/HTTP/test.cc
@@ -419,6 +419,15 @@ struct SerializableObject {
   }
 };
 
+struct NamedSerializableObject {
+  int y = 100;
+  template <typename A>
+  void serialize(A& ar) {
+    ar(CEREAL_NVP(y));
+  }
+  static const char* JSONEntryName() { return "custom_name"; }
+};
+
 TEST(HTTPAPI, PostCerealizableObject) {
   HTTP(FLAGS_net_api_test_port).ResetAllHandlers();
   HTTP(FLAGS_net_api_test_port).Register("/post", [](Request r) {
@@ -427,6 +436,10 @@ TEST(HTTPAPI, PostCerealizableObject) {
   });
   EXPECT_EQ("Data: {\"data\":{\"x\":42,\"s\":\"foo\"}}",
             HTTP(POST(Printf("http://localhost:%d/post", FLAGS_net_api_test_port), SerializableObject())).body);
+
+  EXPECT_EQ(
+      "Data: {\"custom_name\":{\"y\":100}}",
+      HTTP(POST(Printf("http://localhost:%d/post", FLAGS_net_api_test_port), NamedSerializableObject())).body);
 }
 
 TEST(HTTPAPI, PostCerealizableObjectAndParseJSON) {
--- a/Blocks/HTTP/types.h
+++ b/Blocks/HTTP/types.h
@@ -69,7 +69,7 @@ struct GET : HTTPRequestBase<GET> {
   explicit GET(const std::string& url) : HTTPRequestBase(url) {}
 };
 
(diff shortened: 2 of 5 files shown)


Blocks/HTTP/types.h
```diff
@@ -70,7 +70,7 @@ struct GET : HTTPRequestBase<GET> {
 };
 
 // A helper class to fill in request body as either plain text of JSON-ified object.
```
mzhurovich: s/of/or/

Blocks/HTTP/test.cc
```diff
@@ -427,6 +436,10 @@ TEST(HTTPAPI, PostCerealizableObject) {
   });
   EXPECT_EQ("Data: {\"data\":{\"x\":42,\"s\":\"foo\"}}",
             HTTP(POST(Printf("http://localhost:%d/post", FLAGS_net_api_test_port), SerializableObject())).body);
+
+  EXPECT_EQ(
+      "Data: {\"custom_name\":{\"y\":100}}",
```
sompylasar: The object represented by `NamedSerializableObject` is `{"y":100}`, why it's responsible for the outer name?

dkorolev: By design `NamedSerializableObject` has `JSONEntryName() { return "custom_name"; }`.

So that we can finally have proper top-level JSON response object names when we want to -- like I just did yesterday with the profiler.

Blocks/HTTP/types.h
dkorolev: Fixed in d62e87d.

Blocks/HTTP/test.cc
sompylasar: I see this is by design. I think it's not a good design to always force a specific name for an instance of some type. This way the type becomes non-reusable. Imagine I've got a `Slice` and it has a `JSONEntryName` that returns "slice". If I wanted two API endpoints return different kinds of slices, I would make one endpoint name the response instance a "first_endpoint_slice", the other "second_endpoint_slice", both return instances of `Slice`. With the new architecture it looks like a wrong way to go, so I should make a new (maybe derived) type to name my responses differently.

dkorolev: Ah, no, you can always do 'response(my_object, "name")', or 'JSON(my_object,
"custom_name")', or a couple other things. This change is to merely make it
easier to change the default "data".

On Wednesday, July 22, 2015, John Babak <notifications@github.com> wrote:

> In Blocks/HTTP/test.cc
> <https://github.com/c5t/Current/pull/225#discussion_r35267119>:
>
> > @@ -427,6 +436,10 @@ TEST(HTTPAPI, PostCerealizableObject) {
> >    });
> >    EXPECT_EQ("Data: {\"data\":{\"x\":42,\"s\":\"foo\"}}",
> >              HTTP(POST(Printf("http://localhost:%d/post", FLAGS_net_api_test_port), SerializableObject())).body);
> > +
> > +  EXPECT_EQ(
> > +      "Data: {\"custom_name\":{\"y\":100}}",
>
> I see this is by design. I think it's not a good design to always force a
> specific name for an instance of some type. This way the type becomes
> non-reusable. Imagine I've got a Slice and it has a JSONEntryName that
> returns "slice". If I wanted two API endpoints return different kinds of
> slices, I would make one endpoint name the response instance a
> "first_endpoint_slice", the other "second_endpoint_slice", both return
> instances of Slice. With the new architecture it looks like a wrong way
> to go, so I should make a new (maybe derived) type to name my responses
> differently.
>
> —
> <https://github.com/c5t/Current/pull/225/files#r35267119>.
>


-- 
Dima

sompylasar: OK then.

mzhurovich:
LGTM, merging.

PR #226, opened 2015-07-25
`std::forward_list<>`, three-stage locking, and `Profiler`.

dkorolev:
Hi @mzhurovich !

Sorry for three in one; good thing we looked at two of them already, albeit outside pull requests.

Resuming with respect to GirHub pace.

Thanks!
Dima

### DIFF
--- a/Blocks/Persistence/persistence.h
+++ b/Blocks/Persistence/persistence.h
@@ -27,8 +27,8 @@ SOFTWARE.
 
 #include <cassert>
 #include <chrono>
+#include <forward_list>
 #include <functional>
-#include <list>
 #include <mutex>
 #include <thread>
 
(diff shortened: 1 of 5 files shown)


Blocks/Persistence/persistence.h
```diff
@@ -44,13 +44,54 @@ namespace persistence {
...
+    }
+    ~ThreeStagesScopedLock() {
+      assert(stage_ == 3);
+      parent_.stage3_.unlock();
+    }
+    int stage_ = 1;
+    ThreeStageMutex& parent_;
+  };
+
+  struct InnerLevelScopedUniqueLock {
```
mzhurovich: `final` as well?
As I understand, this one and `TopLevelScopedLock` are per-stage scoped locks. Maybe name them `StageThreeScopedLock` and `StageOneScopedLock` respectively?

dkorolev: Yes! 5889d8a

The difference is more `Guard` vs `UniqueLock` than stage 1 vs. 3, since the latter lock is used in `condition_variable::wait(...)`.

Renamed them and added `final`, how's it now?

mzhurovich:
@dkorolev LGTM!

PR #230, opened 2015-08-22
Moved lazy instantiation into Bricks/util.

dkorolev:
Hi @mzhurovich 

More baby steps with Current while my heart is with (at?) CTFO.

Thanks!
Dima

### DIFF
--- a/Bricks/util/lazy_instantiation.h
+++ b/Bricks/util/lazy_instantiation.h
@@ -39,19 +39,11 @@ struct LazyInstantiatorAbstract {
   virtual std::shared_ptr<T> Instantiate() const = 0;
 };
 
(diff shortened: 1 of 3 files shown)


dkorolev:
@mzhurovich -- investigating the `std::forward_as_tuple` issue I took a fresh look and realized that perfect forwarding does not work well with delayed instantiation.

Result:

* Perfect forwarding is not used now.
* The user should explicitly specify `std::ref()` for both from-arguments-list and from-tuple initialization.
* Thus, two forms have been unified and collapsed into one.

The code is actually cleaner this way. :-)

Thanks!
Dima

mzhurovich:
@dkorolev Ok, thanks! LGTM :)

PR #234, opened 2015-09-04
Headers support in HTTP clients and EventCollector.

mzhurovich:


dkorolev:
@mzhurovich Will look tonight. Did you test it in production? :-)

mzhurovich:
@dkorolev Switched to the new `LogEventWithHeaders` structure, keeping the old version. All our parsers should be updated as well.

PR #238, opened 2015-09-18
Using `TypeList` in metaprogramming.

dkorolev:
Hi @mzhurovich ,

`TypeList` is coming, as I'm gradually finding time to go live with `RipCurrent`.

Some of the `TEST`-s are now  just a bunch of `static_assert`-s, which I'm happy with as long as you are.

The rest ... PTAL!

Thanks,
Dima


mzhurovich:
LGTM!

PR #239, opened 2015-09-18
Work in progress

dkorolev:
Hi @mzhurovich 

If you have time after https://github.com/c5t/Current/pull/238.

This PR should be safe to merge in, although it's still work in progress from my end.

* Moved to `clang-format-3.6` (to support `// clang-format off`).
* Fixed `make check`.
* Added a few static tests into `ripcurrent.h`
* On it to have typed message dispatching working.

Thanks!
Dima

### DIFF
--- a/RipCurrent/ripcurrent.h
+++ b/RipCurrent/ripcurrent.h
@@ -30,23 +30,18 @@ SOFTWARE.
 // or dismissed.
 // End-to-end blocks can be run with `(...).RipCurrent().Sync()`. TODO(dkorolev): Not only `.Sync()`.
 //
-// TOP PRIORITY (To use in pilot data crunching very soon, and in CTFO later):
-// TODO(dkorolev): Entry typing and RTTI instead of plain `int`-s.
-//
-// MID PRIORITY:
+// HI-PRI:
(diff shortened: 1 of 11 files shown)


RipCurrent/ripcurrent.h
```diff
@@ -398,12 +404,22 @@ class NextHandlerInitializer {
 };
 
 // Base classes for user-defined code, for `is_base_of<>` `static_assert()`-s.
+template <OutputPolicy, typename EMITTABLE_TYPES_AS_TYPELIST>
+struct ConfirmRHSDoesNotEmit;
```
mzhurovich: How is it used?

mzhurovich:
So we switch to `clang-format-3.6` as this PR is merged, right?

RipCurrent/ripcurrent.h
dkorolev: It won't compile if `CURRENT_RHS` defines any type it emits.

dkorolev:
Yes, I suggest we switch to `clang-format-3.6`. Or `alias` the 3.5 one into it, and not run `make indent` in `RipCurrent/` :)

PR #240, opened 2015-09-20
Further RipCurrent metaprogramming.

dkorolev:
Hi @mzhurovich !

Getting ready to the final step of RTTI-based dispatching.
In this work-in-progress: made the list of emitted types part of RipCurrent subflow node type signature.

A bit of a design doc/idea:

* When it comes to accepting incoming entries, we rely on C++ type dispatching. For example, the user could define a handler for `const BaseClass&`, which will capture all the derived types too. Thus, no need to explicitly make a type list of the accepted types.
* When it comes to emitting, eventually, there will be multithreading, we will use a message queue, and will have to do proper dispatching per type. Thus, the type list is necessary there. There will be a `static_assert` on an attempt to emit an instance of a type that is not listed as emittable. And our `TypeList` can create unions of other `TypeList`-s out of the box, once we'll be be adding `operator+` for RipCurrent-s.

I don't plan on touching Current further today, so please PTAL on your own time.

Thanks!
Dima

### DIFF
--- a/RipCurrent/ripcurrent.h
+++ b/RipCurrent/ripcurrent.h
@@ -111,7 +111,6 @@ class RipCurrentMockableErrorHandler {
 
 // Human-readable symbols for template parameters for one of four open-vs-closed-ended blocks.
 enum class InputPolicy { DoesNotAccept = 0, Accepts = 1 };
-enum class OutputPolicy { DoesNotEmit = 0, Emits = 1 };
 
 // The wrapper to store `__FILE__` and `__LINE__` for human-readable verbose explanations.
 struct FileLine {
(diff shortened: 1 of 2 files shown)


RipCurrent/ripcurrent.h
```diff
@@ -601,30 +661,47 @@ class GenericCurrentSequence : public AbstractCurrentSequence<INPUT, OUTPUT> {
...
-  std::shared_ptr<InstanceBeingRun<INPUT, OUTPUT>> SpawnAndRun(
+  std::shared_ptr<InstanceBeingRun<INPUT, OUTPUT_TYPE_LIST>> SpawnAndRun(
       std::shared_ptr<EntriesConsumer<NEXT_INPUT>> next) const override {
     return std::make_shared<Instance>(this, next);
   }
 };
 
-template <InputPolicy INPUT, OutputPolicy OUTPUT>
-class SubflowSequence : public SharedCurrent<INPUT, OUTPUT> {
+// `SubflowSequence` requres `INTERMEDIATE_TYPE_LIST` to not be an empty type list.
```
mzhurovich: 1) requ**i**res
2) maybe change to `to be nonempty type list`?

RipCurrent/test.cc
```diff
@@ -298,8 +298,8 @@ TEST(RipCurrent, TypeSystemGuarantees) {
   using namespace ripcurrent_unittest;
 
   EXPECT_EQ("Foo", Foo::UnitTestClassName());
-  EXPECT_TRUE(Foo::INPUT_POLICY == InputPolicy::DoesNotAccept);
-  EXPECT_TRUE(Foo::OUTPUT_POLICY == OutputPolicy::Emits);
+  static_assert(Foo::INPUT_POLICY == InputPolicy::DoesNotAccept, "");
```
mzhurovich: Why `EXPECT`'s are bad here?

dkorolev: The `TypeList` part of the `OUTPUT` has to be a `static_assert`, and I just changed the `INPUT` to look the same way.

Overall, if it can fail at compile time, why not let it? :-)

RipCurrent/ripcurrent.h
dkorolev: :+1: 2f9b204

mzhurovich:
@dkorolev LGTM!

PR #241, opened 2015-10-11
Double rounding.

dkorolev:
CC @sompylasar

### DIFF
--- a/Bricks/strings/rounding.h
+++ b/Bricks/strings/rounding.h
@@ -1,7 +1,7 @@
 /*******************************************************************************
 The MIT License (MIT)
 
-Copyright (c) 2014 Dmitry "Dima" Korolev <dmitry.korolev@gmail.com>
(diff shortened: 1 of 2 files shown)


Bricks/strings/rounding.h
```diff
@@ -0,0 +1,54 @@
...
+#define BRICKS_STRINGS_ROUNDING_H
+
+#include <cstring>
+#include <string>
+
+namespace bricks {
+namespace strings {
+
+// Rounds the number to have `n_digits` significant digits.
+std::string RoundDoubleToString(double value, size_t n_digits) {
```
sompylasar: This operation is generalized to "number formatting" and has numerous options, not mentioning locale support, for example, https://github.com/jquery/globalize/blob/master/doc/api/number/number-formatter.md

dkorolev: Not exactly, but very close. We'll get to using them once this part of the product reaches the frontend stage!

PR #242, opened 2015-10-14
Recursive and scoped RmDir.

mzhurovich:
@dkorolev PTAL!

### DIFF
--- a/Bricks/file/file.h
+++ b/Bricks/file/file.h
@@ -197,7 +197,7 @@ struct FileSystem {
   typedef std::ofstream OutputFile;
 
   template <typename F>
-  static inline void ScanDirUntil(const std::string& directory, F&& f) {
+  static inline void ScanDirUntil(const std::string& directory, F&& f, bool include_dirs = false) {
 #ifdef BRICKS_WINDOWS
     WIN32_FIND_DATAA find_data;
     HANDLE handle = ::FindFirstFileA((directory + "\\*.*").c_str(), &find_data);
@@ -211,7 +211,7 @@ struct FileSystem {
       };
       const ScopedCloseFindFileHandle closer(handle);
       do {
-        if (!(find_data.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)) {
+        if (include_dirs || !(find_data.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)) {
           if (!f(find_data.cFileName)) {
             return;
           }
@@ -232,7 +232,7 @@ struct FileSystem {
           // Proved to be required on Ubuntu running in Parallels on a Mac,
           // with Bricks' directory mounted from Mac's filesystem.
           // `entry->d_type` is always zero there, see http://comments.gmane.org/gmane.comp.lib.libcg.devel/4236
-          if (!IsDir(JoinPath(directory, filename))) {
+          if (include_dirs || !IsDir(JoinPath(directory, filename))) {
             if (!f(filename)) {
               return;
             }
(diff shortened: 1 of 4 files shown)


dkorolev:
LGTM!

Why not use an `enum class` instead of that `bool`, like we did for other parameters, with the default of "no"? Then, perhaps, the `*Recursive()` part of `RmDir*()` could use it too, and we'll have a backwards-compatible functionality extension that does not create any new names under `FileSystem::`?

Thanks,
Dima

mzhurovich:
@dkorolev Like this?

Bricks/file/file.h
```diff
@@ -252,11 +256,11 @@ struct FileSystem {
   }
 
   template <typename F>
-  static inline void ScanDir(const std::string& directory, F&& f) {
+  static inline void ScanDir(const std::string& directory, F&& f, bool include_dirs = false) {
```
dkorolev: Here too, `s/bool/ScanRecursive::Yes/`. (I can't think of a better name now. :-( ideally, it's the same symbol that `RmDir` uses, with a more neutral name.)

Bricks/file/file.h
```diff
@@ -284,21 +288,67 @@ struct FileSystem {
...
+          if (IsDir(full_name)) {
+            RmDir(full_name, parameters, RmDirRecursive::Yes);
+          } else {
+            RmFile(full_name, (parameters == RmDirParameters::ThrowExceptionOnError)
+                                  ? RmFileParameters::ThrowExceptionOnError
+                                  : RmFileParameters::Silent);
+          }
+        }, true);
+        RmDir(directory, parameters, RmDirRecursive::No);
+      } catch (...) {
```
dkorolev: Sorry, but `catch (...)` is unacceptable in production code. I know right? ;-)

dkorolev:
LGTM++

Please get rid of `catch(...)`, and sorry you had to suffer through the `clang-format` mess.

mzhurovich:
@dkorolev PTAL once more, pls :)

Bricks/file/test.cc
```diff
@@ -271,3 +273,58 @@ TEST(File, ScanDir) {
...
+  FileSystem::WriteStringToFile("1", f1.c_str());
+  FileSystem::MkDir(dir_y);
+  FileSystem::WriteStringToFile("2", f2.c_str());
+  FileSystem::MkDir(dir_z);
+  FileSystem::WriteStringToFile("3", f3.c_str());
+  FileSystem::WriteStringToFile("4", f4.c_str());
+
+  FileSystem::RmDir(dir_x, FileSystem::RmDirParameters::ThrowExceptionOnError, FileSystem::RmDirRecursive::Yes);
+
+  ASSERT_THROW(FileSystem::WriteStringToFile("test", f3.c_str()), FileException);
```
dkorolev: Any reason the order is `{ f3, f2, f1 }`? :-)

Bricks/file/test.cc
```diff
@@ -271,3 +273,58 @@ TEST(File, ScanDir) {
...
+  {
+    const auto dir_remover = FileSystem::ScopedRmDir(dir_x);
+    FileSystem::MkDir(dir_x);
+    FileSystem::MkDir(dir_y);
+    FileSystem::MkDir(dir_z);
+    FileSystem::WriteStringToFile("PASSED", fn.c_str());
+    EXPECT_EQ("PASSED", FileSystem::ReadFileAsString(fn));
+  }
+  ASSERT_THROW(FileSystem::ReadFileAsString(fn), FileException);
+  ASSERT_THROW(FileSystem::RmDir(dir_z), DirDoesNotExistException);
```
dkorolev: Any reason the order is `{ dir_z, dir_y, dir_x }`? :-)

dkorolev:
**LGTM++** : Final minor comments, feel free to merge this PR yourself after ignoring or taking care of them :-)

Thanks! :+1:

mzhurovich:
@dkorolev Checks are in reverse order, in case something left inside the dirs during the process of recursive removal.

dkorolev:
ACK!

PR #245, opened 2015-11-01
Yoda delete persistence.

dkorolev:


Yoda/test.cc
```diff
@@ -59,8 +64,15 @@ TEST(Yoda, WritesToFile) {
   while (bricks::FileSystem::GetFileSize(persistence_file_name) != yoda_golden_data.size()) {
     ;  // Spin lock.
   }
-
   EXPECT_EQ(yoda_golden_data, bricks::FileSystem::ReadFileAsString(persistence_file_name));
+
+  api.Transaction([](PersistingAPI::T_DATA data) {
+    Dictionary<YodaEntryToPersist>::Mutator(data).Delete("one");
+  }).Wait();
+  while (bricks::FileSystem::GetFileSize(persistence_file_name) != yoda_golden_data_after_delete.size()) {
```
sompylasar: What does `Wait()` do if it needs a spin lock next to it?

sompylasar: P.S. If the persistence is broken (actual does not match golden), the test will hang indefinitely. Maybe compare `> 0`?

dkorolev: The `Wait()` is for the transaction itself (yes, unnecessary), and the spin lock is for the persistence layer / file system (yes, flaky).

Both are going away soon!

dkorolev:
Amended in favor of https://github.com/c5t/Current/pull/246.

PR #253, opened 2015-11-05
JSON serialization for `CURRENT_STRUCT`-s.

dkorolev:
Hi @mzhurovich ,

I've заботал RapidJSON and implemented non-polymorphic serialization with it. It supports primitive types (`uint64_t`), strings, vectors, and nested `CURRENT_STRUCT`-s.

Also, factored out the list of primitive types into a dedicated DSL header, adding more should be one-liners now.

Hope you could jump in and move the JSON logic away from the `test.cc` file -- I've just prototyped it there for now.

Thanks!
Dima

### DIFF
--- a/Reflection/test.cc
+++ b/Reflection/test.cc
@@ -170,12 +170,9 @@ CURRENT_STRUCT(Serializable) {
 CURRENT_STRUCT(ComplexSerializable) {
   CURRENT_FIELD(j, uint64_t);
   CURRENT_FIELD(q, std::string);
-  // TODO(dkorolev): Support `std::vector` during JSON serialization.
-  // CURRENT_FIELD(v, std::vector<std::string>);
+  CURRENT_FIELD(v, std::vector<std::string>);
   CURRENT_FIELD(z, Serializable);
 };
-
-// TODO(dkorolev): When doing serialization, iterate over the fields of the base class too.
 
 struct CollectFieldNames {
   std::vector<std::string>& output_;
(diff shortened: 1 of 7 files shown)


Reflection/test.cc
```diff
@@ -150,3 +159,297 @@ TEST(Reflection, CurrentStructInternals) {
...
+  EXPECT_EQ(
+      "struct Serializable {\n"
+      "  uint64_t i;\n"
+      "  std::string s;\n"
+      "};\n",
+      Reflector().DescribeCppStruct<Serializable>());
+  {
+    std::vector<std::string> result;
+    CollectFieldNames names{result};
+    current::reflection::VisitAllFields<Serializable, current::reflection::FieldTypeAndName>::WithoutObject(
```
mzhurovich: As for now, it could be just `current::reflection::FieldName`.
What do you think, is it worth to keep all the functions or just leave one - `FieldTypeAndName`?

dkorolev: Just `FieldTypeAndName` is probably sufficient.

Also, this test, as well as `RapidJSON` ones, are not there to be kept forever; they just are the reference points.

mzhurovich: Btw, what about index? :)

dkorolev: Didn't need it for this "playground" test code snippet.

mzhurovich:
@dkorolev Cool! Merging :)

dkorolev:
:+1: Yay!

PR #255, opened 2015-11-07
JSON Deserialization.

dkorolev:
Hi @mzhurovich 

And here comes JSON deserialization.

I've added sophisticated error description generation until finally discovering I'm parsing the complex type  giving it a JSON representation of a simple one. :-(

PTAL.

Thanks,
Dima

### DIFF
--- a/Serialization/json.h
+++ b/Serialization/json.h
@@ -61,9 +61,9 @@ void AssignToRapidJSONValue(rapidjson::Value& destination, const T& value) {
 #define CURRENT_DECLARE_PRIMITIVE_TYPE(unused_typeid_index, cpp_type, unused_current_type) \
   template <>                                                                              \
   struct SaveIntoJSONImpl<cpp_type> {                                                      \
-    static void Go(rapidjson::Value& destination,                                          \
-                   rapidjson::Document::AllocatorType&,                                    \
-                   const cpp_type& value) {                                                \
+    static void Save(rapidjson::Value& destination,                                        \
+                     rapidjson::Document::AllocatorType&,                                  \
+                     const cpp_type& value) {                                              \
       AssignToRapidJSONValue(destination, value);                                          \
     }                                                                                      \
   };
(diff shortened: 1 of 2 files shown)


Serialization/json.h
```diff
@@ -128,10 +131,130 @@ std::string JSON(const T& value) {
...
+    LoadFieldVisitor visitor(source, path);
+    current::reflection::VisitAllFields<bricks::decay<T>,
+                                        current::reflection::FieldNameAndMutableValue>::WithObject(destination,
+                                                                                                   visitor);
+  }
+};
+template <>
+struct LoadFromJSONImpl<uint64_t> {
+  static void Load(rapidjson::Value& source, uint64_t& destination, const std::string& path) {
+    if (!source.IsNumber()) {
```
mzhurovich: `.IsUInt64()`?
Btw, they only distinguish 64-bit ints. Should we add additional check for the size of the number, if we deserialize into  8 and 16 bit ints? Same with `float` vs `double` :-/

Serialization/json.h
```diff
@@ -128,10 +131,130 @@ std::string JSON(const T& value) {
...
+      return "The `" + path + "` field could not be parsed.";
+    }
+  }
+  virtual const char* what() const throw() { return ("Expected " + expected_ + ", got: " + value_).c_str(); }
+};
+
+struct InvalidJSONException : ParseJSONException {
+  const std::string erroneus_json_;
+  explicit InvalidJSONException(const std::string& json) : erroneus_json_(json) {}
+  virtual const char* what() const throw() { return erroneus_json_.c_str(); }
```
mzhurovich: Isn't `throw()` specifier deprecated since C++11?

sompylasar:
const std::string expected_;
const std::string value_;

s/value_/actual_/ ?

sompylasar:
```
+ try {
+ ParseJSON<Serializable>("{}");
+ ASSERT_TRUE(false);
+ } catch (const JSONSchemaException& e) {
+ EXPECT_EQ(std::string("Expected number, got: {\"i\":null}"), e.what());
+ }
```

Strictly speaking, a missing field is not the same as a present field with a `null` in it.

sompylasar:
The nested field error is even more confusing: {\"z.i\":\"error\"}".
Better put the path separately from the value in the error message, not try to render it as JSON.

Serialization/json.h
dkorolev: Very good point. I digged further, into `rapidjson/document.h`:

```
// By using proper binary layout, retrieval of different integer types do not need conversions.
union Number {
...
```

I read it that `IsNumber()` is the right check, and `GetUint64()` is the right getter for `uint64_t`.
Re, `float` and `double`, there is no `float` in RapidJSON.

Thus, it appears the code in place now does the right thing.

Serialization/json.h
dkorolev: Removed.

dkorolev:
@sompylasar agree, and it's a compromise. Unfortunately, to dump a part of the JSON object that triggered the error, I need to use RapidJSON's output logic, which -- surprise! -- only supports objects and arrays as top-level nodes.

Thus, I figured the easiest would be to wrap the logic into an object. In which case it looks slightly less confusing to use the "path" to that node as the name of its value in the object -- my first attempt was to call the key `"error"`, and put the node name of that erroneous value in front of it.

Re. `null` being a dedicated value -- perhaps add a dedicated test for it? Let me see...

dkorolev:
Looks like I managed to do the JSON error reporting part cleanly. Please refer to https://github.com/c5t/Current/pull/256.

PR #257, opened 2015-11-08
Preparing to retire the old type system.

dkorolev:
Hi @mzhurovich 

Getting ready to move on with the new type system: Renamed cereal-based logic under a longer name to gradually get rid of it.

Thanks,
Dima


Bricks/strings/rounding.h
```diff
@@ -42,7 +42,7 @@ inline std::string RoundDoubleToString(double value, size_t n_digits) {
   const int dim = static_cast<int>(std::floor((std::log(value) / std::log(10.0)) + 1e-6));
   const double k = std::pow(10.0, static_cast<double>(dim - static_cast<int>(n_digits) + 1));
   std::ostringstream os;
-  os << k * std::round(value / k);
+  os << k* std::round(value / k);
```
mzhurovich: I'd suggest to add parentheses and get rid of this crap.

dkorolev: Yay! Fixed in the next PR.

PR #258, opened 2015-11-09
JSON `std::map<>` support.

dkorolev:
[ To review after https://github.com/c5t/Current/pull/257 ]

Hi @mzhurovich,
Cc @sompylasar,

`std:map<>` serialization into JSON. Uses `{"key1":"value1", ...}` objects for primitive key types (which is wrong for non-strings now!), and `[[{"s":"key1"},"value1"], ...]` arrays for composite types.

Bad news:

* I improvidently `#include`-d all primitive types as possible keys, whereas the logic would only support `std::string`. Need to think whether we would like have `map<int, Foo>` serialize into objects too, or just live with arrays of arrays in those cases.
* http://stackoverflow.com/questions/13842468/comma-in-c-c-macro May need to think further of field declaration syntax (with an obvious suggestion being custom map template types, such as `Dictionary<T>`, `MapStringInto<T>`, etc.). As of this pull request, it's ugly.

Good news:

* The definition `operator<()` worked right away with `CURRENT_RETURNS(bool)`.

Thanks,
Dima


sompylasar:
> Need to think whether we would like have `map<int, Foo>` serialize into objects too, or just live with arrays of arrays in those cases.

Better not, because objects can only have string keys in JSON, and if the strict typing is important, `map<int, Foo>` will be indistinguishable from `map<string, Foo>` in certain cases, e.g. `{{1, 1}, {2, 2}}` and `{{"1", 1}, {"2", 2}}`. The output of a JSON-based API is not as important as the expected input which may be deserialized improperly by the backend because of the ambiguity.

dkorolev:
@sompylasar How about some `map<Card ID, text>`, where `Card ID` is a `uint64` ?

sompylasar:
@dkorolev, same thing, if we prefer strict typing and compatibility. This is not so pretty, though.

dkorolev:
But JSON is never the source of schema; rather, it's always parsed assuming certain schema is being respected.

As an API user, would you prefer to see `{"names":[[1,"alice"],[2,"bob"]]}` or `{"names":{"1":"alice","2":"bob"}}`?

sompylasar:
As a user, I certainly prefer the latter. This means any primitive type
should be converted to string for the map keys.

dkorolev:
@sompylasar : I might well be able to make it to your best liking. :-)
@mzhurovich : Being fixed in the next PR.

PR #259, opened 2015-11-09
Further refactoring.

dkorolev:


dkorolev:
Hi Max!

Which counter -- the counter of registered so far types. Not important now.
Re. space before comma -- I tried, and it looked uglier.

Let's merge or amend this pull request -- a larger one, cleaning up `PSYKHANOOL` and befriending it with the new type system, is already in the works.

Thanks!

PR #260, opened 2015-11-11
Maturing new and phasing out old type systems.

dkorolev:
Hi @mzhurovich !

[[ Cc @sompylasar , but don't be too scared. ]]

`TransactionStorage` and `HTTP` now use the new type system. **NO SHIT!**

This largely retires Cereal from Current, and [hopefully] completely retires Cereal from API/REST/DataDictionary/SystemOfRecord Current users.

Fingers crossed for this to be all we need from `c5t/Current` to use the new type system in `CTFO_BE`. Work in progress there, just want to get this piece out.

Minor tweaks include supporting enum classes, and `FromString`/`ToString` fixes for `string` and `bool`.

`make check` and `make test` pass.

Thanks,
Dima

### DIFF
--- a/Blocks/HTTP/docu/server/docu_03httpserver_04_test.cc
+++ b/Blocks/HTTP/docu/server/docu_03httpserver_04_test.cc
@@ -38,25 +38,18 @@ DEFINE_int32(docu_net_server_port_04, 8082, "Okay to keep the same as in net/api
 using bricks::strings::Printf;
 
   // An input record that would be passed in as a JSON.
-  struct PennyInput {
-    std::string op;
-    std::vector<int> x;
-    std::string error;  // Not serialized.
-    template <typename A> void serialize(A& ar) {
-      ar(CEREAL_NVP(op), CEREAL_NVP(x));
-    }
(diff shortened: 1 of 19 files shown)


dkorolev:
Oh well. Notifications require polymorphism. I'm afraid it's something I niasilyu today.

mzhurovich:
@dkorolev LGTM! Merging it to move further.
A couple of comments:
1. Did you leave Cereal parts to support polymorphism?
2. Did we decide not to convert primitive types to string keys in maps?
3. +1 for defining `is_base_of<CurrentSuper...>::value` as IS_CURRENT_STRUCT.

dkorolev:
@mzhurovich Yay!

To your questions:

1. Not really, just didn't fully clean it up. Will do so as a) polymorphism is done, b) we agree on JSON format (named vs. not names top-level objects), and c) it's worth the time (ex. in Yoda wrt `Padawan` class now having to be a `CURRENT_STRUCT` if we keep Yoda for now.

2. Right, no conversion of keys into primitive types yet. Probably worth doing so though.

3. :+1: 

Thanks!
Dima

PR #265, opened 2015-11-14
Enum support in Reflection and Schema.

mzhurovich:
@dkorolev What do you think?

### DIFF
--- a/TypeSystem/Reflection/schema.h
+++ b/TypeSystem/Reflection/schema.h
@@ -48,6 +48,8 @@ namespace reflection {
 
 CURRENT_STRUCT(TypeInfo) {
   CURRENT_FIELD(type_id, TypeID, TypeID::INVALID_TYPE);
+  // Used only for storing the name of `enum` types.
+  CURRENT_FIELD(name, std::string);
   // Ascending index order in vector corresponds to left-to-right order in the definition of the type,
   // i.e. for `map<int32_t, string>`:
   //   included_types[0] = TypeID::Int32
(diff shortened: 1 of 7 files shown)


TypeSystem/Reflection/schema.h
```diff
@@ -48,6 +48,8 @@ namespace reflection {
 
 CURRENT_STRUCT(TypeInfo) {
   CURRENT_FIELD(type_id, TypeID, TypeID::INVALID_TYPE);
+  // Used only for storing the name of `enum` types.
+  CURRENT_FIELD(name, std::string);
```
dkorolev: `enum_name` ? :)

TypeSystem/enum.h
```diff
@@ -0,0 +1,83 @@
...
+    names_[std::type_index(typeid(T))] = name;
+  }
+
+  template <typename T>
+  ENABLE_IF<std::is_enum<T>::value, std::string> GetName() const {
+    const std::type_index type_index = std::type_index(typeid(T));
+    if (names_.count(type_index) != 0u) {
+      return names_.at(type_index);
+    } else {
+      // TODO: exception?
```
dkorolev: Just `std::cerr << ...` followed by `std::exit(-1);` would be perfect for my taste.

TypeSystem/enum.h
```diff
@@ -0,0 +1,83 @@
...
+#include "../Bricks/template/enable_if.h"
+#include "../Bricks/util/singleton.h"
+
+namespace current {
+namespace reflection {
+
+struct EnumNameSingletonImpl {
+  template <typename T>
+  ENABLE_IF<std::is_enum<T>::value> Register(const char* name) {
+    std::lock_guard<std::mutex> lock_guard(mutex_);
```
dkorolev: If we need this, do we need it in a getter too?

dkorolev:
Wow, great!

Do `enum`-s support forward declarations of some sort? (Perhaps some `TemplateEnum<...>`?)

The only thing that separates this pull request from perfect is `CURRENT_REGISTER_ENUM()` -- could it be some `CURRENT_ENUM(foo, uint64_t) {}` or `CURRENT_ENUM(foo, uint64_t) { Bar = 1ull }` ?

mzhurovich:
@dkorolev The biggest problem is that there is no forward declaration for `enum`-s :(((

mzhurovich:
@dkorolev PTAL

PR #267, opened 2015-11-16
Binary (de)serialization.

mzhurovich:
@dkorolev I didn't make changes to floating point values (de)serialization that we discussed over the phone.

### DIFF
--- a/TypeSystem/Serialization/binary.h
+++ b/TypeSystem/Serialization/binary.h
@@ -1,0 +1,265 @@
+/*******************************************************************************
+The MIT License (MIT)
+
+Copyright (c) 2015 Maxim Zhurovich <zhurovich@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+#ifndef CURRENT_TYPE_SYSTEM_SERIALIZATION_BINARY_H
+#define CURRENT_TYPE_SYSTEM_SERIALIZATION_BINARY_H
+
+#include "exceptions.h"
+
+#include "../struct.h"
+
+namespace current {
+namespace serialization {
+
+// Using platform-independent size type for binary (de)serialization.
+typedef uint64_t BINARY_FORMAT_SIZE_TYPE;
+
+inline void SaveSizeIntoBinary(std::ostream& ostream, const size_t size) {
+  BINARY_FORMAT_SIZE_TYPE save_size = size;
+  size_t bytes_written =
+      ostream.rdbuf()->sputn(reinterpret_cast<const char*>(&save_size), sizeof(BINARY_FORMAT_SIZE_TYPE));
+  if (bytes_written != sizeof(BINARY_FORMAT_SIZE_TYPE)) {
+    throw BinarySaveToStreamException(sizeof(BINARY_FORMAT_SIZE_TYPE), bytes_written);
+  }
+};
+
+inline BINARY_FORMAT_SIZE_TYPE LoadSizeFromBinary(std::istream& istream) {
+  BINARY_FORMAT_SIZE_TYPE result;
+  size_t bytes_read = istream.rdbuf()->sgetn(reinterpret_cast<char*>(&result), sizeof(BINARY_FORMAT_SIZE_TYPE));
+  if (bytes_read != sizeof(BINARY_FORMAT_SIZE_TYPE)) {
+    throw BinaryLoadFromStreamException(sizeof(BINARY_FORMAT_SIZE_TYPE), bytes_read);
+  }
+  return result;
+}
+
+template <typename T>
+struct SaveIntoBinaryImpl;
+
+struct SavePrimitiveTypeIntoBinary {
+  template <typename T>
+  static void Save(std::ostream& ostream, const T& value) {
+    const size_t bytes_written =
+        ostream.rdbuf()->sputn(reinterpret_cast<const char*>(std::addressof(value)), sizeof(T));
+    if (bytes_written != sizeof(T)) {
+      throw BinarySaveToStreamException(sizeof(T), bytes_written);
+    }
+  }
+
+  static void Save(std::ostream& ostream, const std::string& value) {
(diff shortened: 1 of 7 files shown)


TypeSystem/Serialization/binary.h
```diff
@@ -0,0 +1,265 @@
...
+#include "../struct.h"
+
+namespace current {
+namespace serialization {
+
+// Using platform-independent size type for binary (de)serialization.
+typedef uint64_t BINARY_FORMAT_SIZE_TYPE;
+
+inline void SaveSizeIntoBinary(std::ostream& ostream, const size_t size) {
+  BINARY_FORMAT_SIZE_TYPE save_size = size;
```
dkorolev: const?

TypeSystem/Serialization/binary.h
```diff
@@ -0,0 +1,265 @@
...
+  size_t bytes_written =
```
dkorolev: const?

TypeSystem/Serialization/binary.h
```diff
@@ -0,0 +1,265 @@
...
+      ostream.rdbuf()->sputn(reinterpret_cast<const char*>(&save_size), sizeof(BINARY_FORMAT_SIZE_TYPE));
+  if (bytes_written != sizeof(BINARY_FORMAT_SIZE_TYPE)) {
+    throw BinarySaveToStreamException(sizeof(BINARY_FORMAT_SIZE_TYPE), bytes_written);
+  }
+};
+
+inline BINARY_FORMAT_SIZE_TYPE LoadSizeFromBinary(std::istream& istream) {
+  BINARY_FORMAT_SIZE_TYPE result;
+  size_t bytes_read = istream.rdbuf()->sgetn(reinterpret_cast<char*>(&result), sizeof(BINARY_FORMAT_SIZE_TYPE));
```
dkorolev: const ?

TypeSystem/Serialization/binary.h
```diff
@@ -0,0 +1,265 @@
...
+  template <typename T>
+  static void Save(std::ostream& ostream, const T& value) {
+    const size_t bytes_written =
+        ostream.rdbuf()->sputn(reinterpret_cast<const char*>(std::addressof(value)), sizeof(T));
+    if (bytes_written != sizeof(T)) {
+      throw BinarySaveToStreamException(sizeof(T), bytes_written);
+    }
+  }
+
+  static void Save(std::ostream& ostream, const std::string& value) {
```
dkorolev: `bool` probably needs special treatment too, it's not always one byte.

TypeSystem/Serialization/binary.h
```diff
@@ -0,0 +1,265 @@
...
+        istream.rdbuf()->sgetn(reinterpret_cast<char*>(std::addressof(destination)), sizeof(T));
+    if (bytes_read != sizeof(T)) {
+      throw BinaryLoadFromStreamException(sizeof(T), bytes_read);
+    }
+  }
+
+  static void Load(std::istream& istream, std::string& destination) {
+    BINARY_FORMAT_SIZE_TYPE size = LoadSizeFromBinary(istream);
+    destination.resize(static_cast<size_t>(size));
+    // Should be legitimate since c++11 requires internal buffer to be contiguous.
```
dkorolev: Yes.

TypeSystem/Serialization/test.cc
```diff
@@ -69,20 +72,74 @@ CURRENT_STRUCT(ComplexSerializable) {
...
+  const auto tmp_file_remover = bricks::FileSystem::ScopedRmFile(tmp_file);
+  {
+    std::ofstream ofs(tmp_file);
+
+    Serializable simple_object;
+    simple_object.i = 42;
+    simple_object.s = "foo";
+    simple_object.b = true;
+    simple_object.e = Enum::SET;
+    SaveIntoBinary(ofs, simple_object);
```
dkorolev: How about `std::ostringstream`, and check its size?

dkorolev: Although the file one is probably more end-to-end. Disregard. :-)

TypeSystem/Serialization/test.cc
```diff
@@ -144,6 +202,52 @@ TEST(Serialization, JSON) {
       "{\"j\":43,\"q\":\"bar\",\"v\":[\"one\",\"two\"],\"z\":{\"i\":42,\"s\":\"foo\",\"b\":true,\"e\":100}}",
       JSON(complex_object));
 
+  // Serialization of floating point types.
+  // TODO: Think of dealing with 'NaN' and `inf` - JSON standard does not allow them.
+  //       Their string representation in C++ is platform-specific.
+  //       Thus, this test now is GCC/clang-specific.
+  {
+    static_assert(std::numeric_limits<float>::is_iec559, "`float` type is not IEC-559 compliant.");
```
dkorolev: Eventually move into `port.h`?

TypeSystem/Serialization/test.cc
```diff
@@ -144,6 +202,52 @@ TEST(Serialization, JSON) {
...
+    static_assert(std::numeric_limits<double>::is_iec559, "`double` type is not IEC-559 compliant.");
+
+    WithFloatingPoint floating_min;
+    floating_min.f = std::numeric_limits<float>::min();
+    floating_min.d = std::numeric_limits<double>::min();
+    const std::string serialized_fmin = JSON(floating_min);
+    EXPECT_EQ("{\"f\":1.175494350822287508e-38,\"d\":2.2250738585072013831e-308}", serialized_fmin);
+    WithFloatingPoint fmin_from_json = ParseJSON<WithFloatingPoint>(serialized_fmin);
+    EXPECT_FLOAT_EQ(std::numeric_limits<float>::min(), fmin_from_json.f);
+    // Due to precision issues, minimum double value becomes 0.
```
dkorolev: As in, the next line fails, or what?

TypeSystem/Serialization/test.cc
```diff
@@ -144,6 +202,52 @@ TEST(Serialization, JSON) {
...
+    EXPECT_NEAR(std::numeric_limits<double>::min(), fmin_from_json.d, 2.2250738586e-308);
```
dkorolev: Better would be to either test input as a golden number too, or test input vs. output, or both. Assigning the input from a named C++ constant, but comparing it to an immediate value is a bit confusing.

TypeSystem/struct.h
```diff
@@ -155,21 +155,23 @@ struct WithoutParentheses<int(T)> {
 
 #define CURRENT_FIELD_REFLECTION(idx, type, name)                                                              \
   template <class F>                                                                                           \
-  static void CURRENT_REFLECTION(F&& f,                                                                        \
+  static void CURRENT_REFLECTION(F&& CURRENT_CALL_F,                                                           \
```
dkorolev: :trollface:

dkorolev:
LGTM++, see a few minor comments inside.

TypeSystem/Serialization/binary.h
mzhurovich: :+1:

TypeSystem/Serialization/binary.h
mzhurovich: :+1:

TypeSystem/Serialization/binary.h
mzhurovich: :+1:

TypeSystem/Serialization/binary.h
mzhurovich: :ok_hand:

TypeSystem/Serialization/test.cc
mzhurovich: This is the reason for the next line to be `EXPECT_NEAR` instead of `EXPECT_DOUBLE_EQ`.

TypeSystem/Serialization/test.cc
mzhurovich: I'm not sure I understand you right. This is a comparison with the same named constant `std::numeric_limits<double>::min()` with the precision passed as third parameter.

TypeSystem/Serialization/test.cc
mzhurovich: Yep.

mzhurovich:
@dkorolev PTAL
Btw, speaking of floating point - printing Infinity as `%g` gives `inf` as well.
We may think of serializing NaN and Infinity as, for example, `1e999` and `1e1000` in our JSON output.

dkorolev:
LGTM++

PR #272, opened 2015-11-22
Fixed two bugs and confused by the 3rd.

dkorolev:
Hi @mzhurovich ,

The `"wow"` test is failing, and I can't wrap my mind around why so. Moving the declaration of `entry` into the inner loop of deserializing `std::map` helps, but, after 15 minutes, I have no idea why.

Thoughts?


dkorolev:
Confused be no more: https://github.com/dkorolev/Current/blob/0ed5f266d6180d73f0e6439f1eba2e46a2cbccd8/TypeSystem/Serialization/binary.h#L192

mzhurovich:
@dkorolev Cool! Looks like it's c++ library dependent bug :( OS X uses libc++, while Linux uses libstdc++.

PR #273, opened 2015-11-22
Optional

dkorolev:
Hi @mzhurovich ,

After https://github.com/c5t/Current/pull/272 -- added `Optional<T>`, and support for it in JSON serialization / deserialization.

Polymorphics next.

Thanks,
Dima

PR #274, opened 2015-11-22
Polymorphic types.

dkorolev:
Hi Max!

For after https://github.com/c5t/Current/pull/273.

Polymorphic types, all but serialization by now.

**Suggestion**: Looks like with a bit of tweaking, the dream may come true. Specifically:

* Use global `Exists(...)` or `Exists<T>(...)` for base types, optionals, and polymorphics.
* Use global `Value(...)` or `Value<T>(...)` for base types, optionals, and polymorphics.

We'll have to be careful with type hierarchies of types within `Polymorphic<>`. As of now, `Match()` would do the right thing with RTTI dispatching, while `Value<T>` would happily case derived to base if requested. The above does sound right to me, but want to double-check with you.

Thanks,
Dima

PS: Better naming suggestions for `Match()`? I don't like it. `MatchWith()` is more functional, but still doesn't sound right in C++.

### DIFF
--- a/TypeSystem/Serialization/json.h
+++ b/TypeSystem/Serialization/json.h
@@ -30,6 +30,10 @@ SOFTWARE.
 
 #include "exceptions.h"
 
+#include "../struct.h"
+#include "../optional.h"
+#include "../polymorphic.h"
+
 #include "../Reflection/reflection.h"
 
 #include "../../Bricks/template/enable_if.h"
--- a/TypeSystem/Serialization/json.h
+++ b/TypeSystem/Serialization/json.h
@@ -368,8 +368,8 @@ struct LoadFromJSONImpl<std::map<TK, TV>> {
                                                              const std::string& path) {
     if (source && source->IsObject()) {
       destination.clear();
+      std::pair<TK, TV> entry;
       for (rapidjson::Value::MemberIterator cit = source->MemberBegin(); cit != source->MemberEnd(); ++cit) {
-        std::pair<TK, TV> entry;  // TODO(dkorolev): Investigate.
         LoadFromJSONImpl<TK>::Load(&cit->name, entry.first, path);
         LoadFromJSONImpl<TV>::Load(&cit->value, entry.second, path);
         destination.insert(entry);
@@ -384,8 +384,8 @@ struct LoadFromJSONImpl<std::map<TK, TV>> {
                                                               const std::string& path) {
     if (source && source->IsArray()) {
       destination.clear();
+      std::pair<TK, TV> entry;
       for (rapidjson::Value::ValueIterator cit = source->Begin(); cit != source->End(); ++cit) {
-        std::pair<TK, TV> entry;  // TODO(dkorolev): Investigate.
         if (!cit->IsArray()) {
           throw JSONSchemaException("map entry as array", source, path);
         }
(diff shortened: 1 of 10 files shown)


TypeSystem/Serialization/json.h
```diff
@@ -212,160 +231,185 @@ struct LoadFromJSONImpl {
...
   }
 };
 
 template <typename T>
 struct LoadFromJSONImpl<std::vector<T>> {
-  static void Load(rapidjson::Value& source, std::vector<T>& destination, const std::string& path) {
-    if (!source.IsArray()) {
+  static void Load(rapidjson::Value* source, std::vector<T>& destination, const std::string& path) {
+    if (source && source->IsArray()) {
+      const size_t n = source->Size();
```
mzhurovich: `n`->`size`?

TypeSystem/polymorphic.h
```diff
@@ -0,0 +1,182 @@
...
+template <typename T, typename... TS>
+struct PolymorphicImpl<T, TS...> {
+  using T_TYPELIST = TypeListImpl<T, TS...>;
+  enum { T_TYPELIST_SIZE = TypeListSize<T_TYPELIST>::value };
+
+  std::unique_ptr<CurrentSuper> object_;
+
+  // Initializes an `std::unique_ptr<CurrentSuper>` given the object of the right type.
+  // The input object could be an object itself (in which case it's copied),
+  // an `std::move()`-d `std::unique_ptr` to that object (in which cast it's moved),
```
mzhurovich: s/cast/case/

TypeSystem/polymorphic.h
```diff
@@ -0,0 +1,182 @@
...
+  // or a bare pointer (in which case it's captured).
+  // TODO(dkorolev): The bare pointer one is sure unsafe -- consult with @mzhurovich.
+  struct TypeCheckedAsisgnment {
```
mzhurovich: s/Asisgnment/Assignment/

TypeSystem/Serialization/json.h
dkorolev: :+1: 90329ba

TypeSystem/polymorphic.h
dkorolev: :+1: 90329ba

TypeSystem/polymorphic.h
dkorolev: :+1: 90329ba

mzhurovich:
@dkorolev As for `Match()` - how about `Apply()` or `Call()` or `RTTICall()`?

dkorolev:
Renamed `Match()` into `Call()`, and removed that confusing `const` in the constructor parameter to `Optional<>`.

@mzhurovich PTAL.

Thanks,
Dima

TypeSystem/polymorphic.h
```diff
@@ -0,0 +1,184 @@
...
+    template <typename Q>
+    static void Perform(Q&& source, std::unique_ptr<CurrentSuper>& destination) {
+      Instance instance;
+      instance(DerivedTypesDifferentiator<bricks::decay<Q>>(), std::forward<Q>(source), destination);
+    }
+  };
+
+  PolymorphicImpl() = delete;
+
+  // PolymorphicImpl(std::unique_ptr<CurrentSuper>&& rhs) : object_(std::move(rhs)) {}
```
mzhurovich: Is it deliberately commented out?

dkorolev: Fixed in 33fc228 !

mzhurovich:
@dkorolev Помолясь применимъ.

dkorolev:
Воистинъ Аминъ.

PR #275, opened 2015-11-23
`Optional` support in reflection and binary serialization.

mzhurovich:
@dkorolev PTAL

### DIFF
--- a/TypeSystem/Serialization/test.cc
+++ b/TypeSystem/Serialization/test.cc
@@ -94,11 +94,11 @@ CURRENT_STRUCT(WithOptional) {
 
 TEST(Serialization, Binary) {
   using namespace serialization_test;
+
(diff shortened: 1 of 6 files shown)


TypeSystem/Serialization/test.cc
```diff
@@ -540,6 +540,49 @@ TEST(Serialization, JSONForCppTypes) {
...
+
+    with_optional.b = true;
+    SaveIntoBinary(ofs, with_optional);
+
+    with_optional.i = nullptr;
+    SaveIntoBinary(ofs, with_optional);
+  }
+  {
+    std::ifstream ifs(tmp_file);
+    const auto parsed_empty = LoadFromBinary<WithOptional>(ifs);
```
dkorolev: Nit: Wrapping per-deserialized tests into individual subscopes would eliminate the room for mistakenly comparing the result of the 2nd read to the result of the 1st one.

PR #278, opened 2015-11-24
Migrated to latest RapidJSON.

mzhurovich:
1. Added RapidJSON files and changes in Current in two separate commits.
2. Deprecated floating point test completely, since:
  * RapidJSON now outputs NaN's and Inf's as max type value;
  * Looks like output precision is not constant due to a new sophisticated conversion algorithm: http://www.cs.tufts.edu/~nr/cs257/archive/florian-loitsch/printf.pdf


dkorolev:
Looking into. `Blocks/HTTP` tests failed from the 1st run.

mzhurovich:
@dkorolev
One more thing. Custom files in '3rdparty/rapidjson':
* `version.h` with the info about the version of the files
* `streamwrapper.h` - new iostream wrapper based on original example.

dkorolev:
Oh, let's put them into the README of that directory then.

mzhurovich:
@dkorolev Looks like it works now :)
What did you mean exactly to put into 'README'?

dkorolev:
Looking. Your comments about added/modified files. Perhaps rename `version.txt` into `README[.md`, and add the above comment there?

dkorolev:
MERGING!

mzhurovich:
Yay!

PR #280, opened 2015-11-24
Full JSON support for `Polymorphic` from the new type system.

dkorolev:


TypeSystem/polymorphic.h
```diff
@@ -109,6 +111,9 @@ struct PolymorphicImpl<T, TS...> {
 
   PolymorphicImpl() = delete;
 
+  // `ParseJSON`, needs to create a temporary object prior to populating it.
```
mzhurovich: Extra comma.

PR #284, opened 2015-11-27
Classes and logic for timestamps + checking in the loadtest.

dkorolev:
Hi @mzhurovich ,

What we discussed about timestamps: I took the liberty to use C++'s default top-level class and only add two specializations to it, for `us` and `ms` respectively.

Thanks,
Dima


mzhurovich:
@dkorolev The biggest question for me is why we can't just use `std::chrono::{milli/micro}seconds`?

PR #286, opened 2015-11-28
Polymorphic serialization and retired non-standard timestamps.

dkorolev:
Hi @mzhurovich !

Good news in the commits: fully retired our timestamp in favor of `std::chrono` ones, defeated `Polymorphic<>` to have it construct itself in the way we need it, and cleaned up the example quite a bit.

Remains to discuss:
* Apparently, we don't serialize base classes into JSONs now. :-( Just as I've removed all TODOs from `example/EmbeddedDB`, I've realized this one has to be put back.
* I've made timestamps signed (`int64_t` instead of `uint64_t`). They support time intervals too, after all, and `std::chrono::microseconds::count()` is signed by itself.
* Sorry if `clang-format` messed up something. Hope I've cleaned it, but just in case.

Thanks,
Dima


TypeSystem/test.cc
```diff
@@ -42,23 +42,27 @@ CURRENT_STRUCT(Foo) {
...
+};
+CURRENT_STRUCT(Baz) {
   CURRENT_FIELD(v1, std::vector<uint64_t>);
   CURRENT_FIELD(v2, std::vector<Foo>);
   CURRENT_FIELD(v3, std::vector<std::vector<Foo>>);
   CURRENT_FIELD(v4, (std::map<std::string, std::string>));
   CURRENT_FIELD(v5, (std::map<Foo, int>));
 };
-CURRENT_STRUCT(DerivedFromFoo, Foo) { CURRENT_FIELD(bar, Bar); };
+CURRENT_STRUCT(DerivedFromFoo, Foo) { CURRENT_FIELD(bar, Baz); };
```
mzhurovich: s/bar/baz/ as well?

dkorolev: :+1: 22417c2

PR #292, opened 2015-11-29
More refactoring.

dkorolev:
[ After https://github.com/c5t/Current/pull/291 ]

More refactoring, mostly to have the example do its job, and to highlight what remains to be done.

Thanks,
Dima


examples/EmbeddedDB/benchmark.cc
```diff
@@ -0,0 +1,101 @@
...
+// instead of using `epoll()` to maximize client-side throughput.
+
+#include "../../current.h"
+
+#include "schema.h"
+
+using namespace current;
+
+DEFINE_double(seconds, 2.5, "Run the load test for this many seconds.");
+DEFINE_string(url, "http://localhost:%d", "The URL for the load test, default to localhost:${FLAGS_port}.");
```
mzhurovich: ...default to **http://**localhost...

dkorolev: Good catch, fixed.

PR #295, opened 2015-11-29
Scoped registerers, for HTTP and examples/EmbeddedDB.

dkorolev:

PR #296, opened 2015-11-30
New reflection

dkorolev:
[ After https://github.com/c5t/Current/pull/295 ]

It would be a bit too long of a story of suffering and misery; and the story not over, even though the code does what it should now.

The major issue had to do with templated constructors not being first-class C++ citizens, while in quite a few places, including `CURRENT_CONSTRUCTOR`, our constructors are indeed templated. Thus, while making `operator=` work was a relatively straightforward task, the `a = Clone(b);` syntax is, well, substantially different.

After some meditation я психанул and have done it "right": with the proper `Clone` method.

Implementation gist: each type `T` can be `reinterpret_cast` into a binary-compatible `Stripped<T>`, which supports copy construction and assignment. In case of `Polymorphic` and `Optional`, the stripped version does the right thing internally. Stripped versions also can be default-constructed.

Remaining pain point: as templated constructors are not first-class citizens, I've copy-pasted (sic!) the implementations of `Polymorphic` and `Optional` to clearly separate the "production" and "copyable" logic of them -- since some should support construction natively, and some should not. This can surely be done better, but I don't feel empowered to do it as of now.

The implementation is complete, to the degree that `Stripped<T>` traverses C++ types recursively. Thus, tricky cases, like polymorphic within optional, should be good to go now.

Hope it unblocks us.

Thanks,
Dima


mzhurovich:
Звучит композиция ВИА "Раммштайн": http://www.youtube.com/watch?v=Qb_Sfq6spbk

mzhurovich:
Yay!

PR #302, opened 2015-12-02
Baby steps.         

mzhurovich:


TypeSystem/README.md
```diff
@@ -0,0 +1,88 @@
+# Current Framework Type System
+
+**Current** framework uses its own type system to define data structures and schemas. The main building block is structure, which natively supports reflection and serialization into binary and JSON formats.
```
dkorolev: `is a` / `is the` structure.

Reflection for structures is very important. Serialization -- not so much here, imho, since it's a) more complex (think zero-copy), and b) comes after type invariant checking and type system guarantees.

TypeSystem/README.md
```diff
@@ -0,0 +1,88 @@
...
+Structure inheritance is strictly linear, no diamond-shape inheritance is allowed.
+
+### Structure fields
+Each field inside the structure **must** be declared via the following macro:
+```cpp
+CURRENT_FIELD(FieldName, FieldType[, DefaultValue]);
+```
+**Important notes**: 
+* pure C++ declaration of member variables is **not allowed** for Current structures;
+* if field type contains commas, it should be put into parentheses:
```
dkorolev: :+1:
`if the field type`

TypeSystem/README.md
```diff
@@ -0,0 +1,88 @@
...
+* fixed-size integral
+  * `int8_t` / `uint8_t`
+  * `int16_t` / `uint16_t`
+  * `int32_t` / `uint32_t`
+  * `int64_t` / `uint64_t`
+* IEC-559 compliant floating point
+  * `float` (32-bit)
+  * `double` (64-bit)
+* STL containers
+  * `std::pair`
```
dkorolev: pair should probably go last :)
and let's keep std::tuple in mind.

TypeSystem/Serialization/test.cc
```diff
@@ -151,7 +150,7 @@ TEST(Serialization, Binary) {
 
     const DerivedSerializable c = LoadFromBinary<DerivedSerializable>(ifs);
     EXPECT_EQ(48ull, c.i);
-    EXPECT_EQ("baz", c.s);
+    EXPECT_EQ("baz\0baz", c.s);
```
dkorolev: :+1:

dkorolev:
Great doc!

TypeSystem/README.md
```diff
@@ -0,0 +1,88 @@
...
+  * `std::map`
+  * `std::vector`
+* `std::string`
+* `std::chrono::milliseconds/microseconds`
+* `enum class`-es declared via `CURRENT_ENUM`:
+```cpp
```
dkorolev: Nit: concepts and semantics got mixed together here. "enum classes declared via `CURRENT_ENUM`" is enough, with the definition of the syntax of `CURRENT_ENUM` being a paragraph or two away.

dkorolev:
Could you please add a link to `TypeSystem/README.md` at the top of the top-level `README.md` as well?

mzhurovich:
@dkorolev fixed.

PR #303, opened 2015-12-03
A couple minor nits.

dkorolev:


dkorolev:
@mzhurovich : Forgot to rename that header file or class within it. On me.

PR #305, opened 2015-12-03
TypeSystem refactoring.

mzhurovich:


dkorolev:
@mzhurovich Merging.

Minor: Fixing the test in a coming PR. Interestingly, `clang++` didn't catch it, while `g++` did.

Three comments:
1. Perhaps rename `CurrentStructSuper` -> `CurrentStruct`, and introduce just `CurrentVariant` instead?
2. Re, `INVALID_TYPE = 0u, CurrentStructSuper = 1u` -- perhaps just `UnspecifiedType = 0` would do here?
3. It still confuses part of me that renaming `Polymorphic` into `Variadic` changes the type ID. Unless I'm misreading something.

Thanks,
Dima

mzhurovich:
@dkorolev 
1. Looks ok.
2. I'd prefer to distinguish unspecified type from particular base type.
3. Type ID depends on the name of the structure. Do you believe that it's better to treat different structures containing the same set of fields as equivalent?

PR #311, opened 2015-12-06
Speedups for `Variant<>` compilation.

dkorolev:
Hi @mzhurovich ,

Gist: Template metaprogramming should be manageable by clang++ on your machine with this change. I've confirmed the memory usage doesn't skyrocket.

[[ Prerequisite https://github.com/c5t/Current/pull/310 ]]

After some research, it appears `RTTIDynamicCall` performs worse than `combine`. I ended up aggressively (but w/o preoptimization!) tweaking the existing approach.

As the result of the research, it seems there are two slow parts of metaprogramming:

1. auto-generating many temporary inter-dependent classes (combine), and
2. auto-generating many disconnected classes as the result of auto-generating other classes (rtti).

At the same time, `enable_if` / SFINAE and `weed` / `decltype` don't seem to make a big difference. True they can make compilation time some 1.5 times slower, but they don't eat memory as aggressively, and they don't make the compiler go crazy and spend 10x time building a simple thing.

Also, it looks like both `g++` and `clang++` are smart, to the degree that the "complexity" of compilation follows the usual complexity rules, as in `O(N) + O(N) = O(N)`. In other words, minor improvements don't show much of an improvement; at least I wasn't able to "hit the gold", but only to make small steps.

Anyway, the `combine` approach does look like the most promising one, and I've cleaned it up a lot for the `Variadic<>` usecase.

Atop the above, I've removed that questionable bare pointer usecase.

Also see the comment to the https://github.com/c5t/Current/commit/0acfe188451052e79746145124f8c8b07c8f7c22 commit. The top set of benchmarks are the ones on Hetzner for the commit from this pull request.

Thanks,
Dima

PS: `Variant<>::Clone` is unprofiled now, and it may require some love too.

### DIFF
--- a/TypeSystem/variant.h
+++ b/TypeSystem/variant.h
@@ -41,38 +41,28 @@ SOFTWARE.
 namespace current {
 
 // A high-octane version of `current::metaprogramming::combine.
-// A "poor man's version" for higher performance.
 namespace variant_high_octane_tmp {
 
+template <typename T, typename... TS>
+struct fast_combine : fast_combine<T>, fast_combine<TS...> {
+  using fast_combine<T>::operator();
+  using fast_combine<TS...>::operator();
+};
+
 template <typename T>
-struct fast_dispatch {
+struct fast_combine<T> {
   T instance;
 
-  template <typename P2, typename P3>
-  void operator()(typename T::enabled_type_1 selector, P2&& p2, P3&& p3) {
-    instance.operator()(selector, std::forward<P2>(p2), std::forward<P3>(p3));
+  template <typename PARAM>
+  void operator()(typename T::first_parameter_1 first, PARAM&& param, typename T::third_parameter& third) {
+    instance.operator()(first, std::forward<PARAM>(param), third);
   }
 
-  template <typename P2, typename P3>
-  void operator()(typename T::enabled_type_2 selector, P2&& p2, P3&& p3) {
-    instance.operator()(selector, std::forward<P2>(p2), std::forward<P3>(p3));
+  template <typename PARAM>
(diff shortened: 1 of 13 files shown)


TypeSystem/variant.h
```diff
@@ -40,6 +40,32 @@ SOFTWARE.
...
+  template <typename PARAM>
+  void operator()(typename T::first_parameter_1 first, PARAM&& param, typename T::third_parameter& third) {
+    instance.operator()(first, std::forward<PARAM>(param), third);
+  }
+
+  template <typename PARAM>
+  void operator()(typename T::first_parameter_2 first, PARAM&& param, typename T::third_parameter& third) {
+    instance.operator()(first, std::forward<PARAM>(param), third);
+  }
+};
```
mzhurovich: It could be written even shorter:
```cpp
template <typename T>
struct fast_combine_impl {
  T instance;

  template <typename PARAM>
  void operator()(typename T::first_parameter_1 first, PARAM&& param, typename T::third_parameter& third) {
    instance.operator()(first, std::forward<PARAM>(param), third);
  }

  template <typename PARAM>
  void operator()(typename T::first_parameter_2 first, PARAM&& param, typename T::third_parameter& third) {
    instance.operator()(first, std::forward<PARAM>(param), third);
  }
};

template <typename... TS>
struct fast_combine : fast_combine_impl<TS>... {};
```

dkorolev: Without `using`-s it won't compile on g++, only on clang++. Just checked.

And on clang++, the difference in compilation time is negligible.

mzhurovich: Ah, ok.

PR #315, opened 2015-12-14
Storage v0.01

dkorolev:
Hi @mzhurovich,

YOUR CODE HAZ WORK!

Thanks,
Dima


dkorolev:
@mzhurovich We're probably better off renaming `FieldsByReference` into just `Data`.

PR #317, opened 2015-12-14
Added Storage test into TypeTest.

mzhurovich:
@dkorolev Storage test with 100 structures takes ~71 seconds to compile on my Air :-/


mzhurovich:
Looks like it grows linearly.

dkorolev:
LGTM

PR #319, opened 2015-12-15
Work in progress.

dkorolev:
Hi Max,

Didn't get much keyboard time today. Branched out the `Future` and make it possible to require it to be waited for, haven't connected it to the Storage yet.

Thanks,
Dima


mzhurovich:
@dkorolev Merging

PR #320, opened 2015-12-16
Minor tweaks + semantics.

dkorolev:
Hi @mzhurovich,

Minor tweaks, with the main one being the (functional!) prototype of a call to `Transaction()` accepting the 2nd parameter.

Thanks,
Dima


mzhurovich:
LGTM!

PR #321, opened 2015-12-17
Mostly cleanup.

dkorolev:

PR #322, opened 2015-12-17
Sherlock + JSON + transactional persistence = <3 !

dkorolev:
Hi @mzhurovich,

After https://github.com/c5t/Current/pull/321, just one commit: https://github.com/c5t/Current/commit/e2bf8b1a5a39186e40027f13859db96d06a9f7cf

It turns out cleaner than I anticipated. And `pair<vector<Variant<mutations...>>, timestamp>` worked like a charm.

Remaining items:
* Proper transactions: Merge in your code.
* Replay: Move it into Persister's constructor?
* Custom type: The demo should probably include adding a custom storage type too, like `DictionaryWithTopElements`, that maintains an `unordered_map` by key and another data structure sorted by some `score` in descending order?

Minor comment: Let's rename the `_ALIAS` into `*_PERSISTENCE` or something similar?

Thanks,
Dima


mzhurovich:
@dkorolev Yay! Let's discuss everything over the phone today?
One more question - how do we deal with the timestamps if we only set them in transactions? I mean, end user code should somehow be aware of the time when any particular entry has been added.

dkorolev:
@mzhurovich Timestamps are available while listening to the log; during the transaction itself one can only know the current time, and the transaction will be recorded strictly in the future compared to it. Ready in 1.5 hours!

PR #325, opened 2015-12-22
Midichlorians OS X test + minor Apple tweaks.

mzhurovich:


Midichlorians/Dev/Beta/iOS/MidichloriansImpl.mm
```diff
@@ -55,7 +55,13 @@ of this software and associated documentation files (the "Software"), to deal
 #endif  // (TARGET_OS_IPHONE > 0)
 
 namespace midichlorians {
-     
+
+    // Conversion from [possible nullptr] `const char*` to `std::string`.
+    static std::string UnsafeToStdString(const char* s) {
+    return s == nullptr ? std::string() : s;
```
dkorolev: indent?

Midichlorians/test.cc
```diff
@@ -0,0 +1,135 @@
...
+
+// Use the define below to enable debug output via `NSLog`.
+// #define CURRENT_APPLE_ENABLE_NSLOG
+
+#include "../port.h"
+
+#ifdef CURRENT_APPLE
+
+// The following line is needed to avoid OS X headers conflicts with C++.
+#define __ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES 0
```
dkorolev: :+1:

Midichlorians/test.cc
```diff
@@ -0,0 +1,135 @@
...
+#include "../3rdparty/gtest/gtest-main-with-dflags.h"
+
+DEFINE_int32(http_port, 8383, "Port to spawn server on.");
+DEFINE_string(http_route, "/log", "HTTP route of the server.");
+
+struct Server {
+  using T_EVENTS = TypeList<iOSAppLaunchEvent, iOSIdentifyEvent, iOSFocusEvent, iOSGenericEvent>;
+
+  Server(int http_port, const std::string& http_route)
+      : messages_processed(0u), http_port(http_port), http_route(http_route) {
```
dkorolev: We haz scoped registerers! https://github.com/c5t/Current/blob/master/examples/EmbeddedDB/db.cc#L63-L64

Midichlorians/test.cc
```diff
@@ -0,0 +1,135 @@
...
+  NSDictionary* launchOptions = [NSDictionary new];
+  [Midichlorians setup:[NSString
+                           stringWithFormat:@"http://localhost:%d%s", FLAGS_http_port, FLAGS_http_route.c_str()]
+      withLaunchOptions:launchOptions];
+
+  [Midichlorians focusEvent:YES source:@"applicationDidBecomeActive"];
+
+  [Midichlorians identify:@"unit_test"];
+
+  NSDictionary* eventParams = @{ @"s" : @"str", @"x" : @1 };
```
dkorolev: :+1: excellent!

Perhaps a bool too?

port.h
```diff
@@ -116,7 +116,15 @@ SOFTWARE.
 #define M_PI 3.14159265359
 #endif
 
+#endif  // CURRENT_WINDOWS
+
+#ifdef CURRENT_APPLE
+#ifdef CURRENT_APPLE_ENABLE_NSLOG
+#define CURRENT_NSLOG(...) ::NSLog(__VA_ARGS__)
+#else
+void CURRENT_NSLOG(...) {}
```
dkorolev: Why `{}`? I don't know Objective C well, but won't an empty define do it?

If the users may put a semicolon after `CURRENT_NSLOG(...)`, perhaps some `static_cast<void>(0)` is a safer call?

dkorolev:
LGTM++, thanks for doing this!

Midichlorians/Dev/Beta/iOS/MidichloriansImpl.mm
mzhurovich: :+1:

dkorolev: Nit, as we're on this page: `return s ? s : "";` is easier :)

Midichlorians/test.cc
mzhurovich: :ok_hand:

port.h
mzhurovich: Empty define may cause warnings, if the argument is used inly for logging.
What's wrong with global no-op function? I didn't get the idea with semicolon.

Midichlorians/Dev/Beta/iOS/MidichloriansImpl.mm
mzhurovich: Yes!

Midichlorians/test.cc
mzhurovich: Ok, though it serializes to 0/1 :)

port.h
dkorolev: The user may say something like `if (blah) CURRENT_NSLOG(...); else { ... }`. This macro would translate it into `if (blah) {}; else { ... }`, which won't compile.

A global no-op function is fine, but `{}` is not it.

Midichlorians/test.cc
dkorolev: Might well, but why not test for it as well as we're on this page? :)

port.h
mzhurovich: It is `void CURRENT_NSLOG(...) {}` - global no-op funcion

mzhurovich: Maybe it's worth to make it `inline` :)

dkorolev: **AUCH**. Me needs a vacation then.

With `inline` it would work just fine. Thanks!

mzhurovich:
@dkorolev PTAL

dkorolev:
(LGTM++)++!

PR #326, opened 2015-12-23
Midichlorians using Current type system.

mzhurovich:
@dkorolev Slightly redesigned event system, which now uses Current structs. The old code left almost unchanged, except for `emit` method.

### DIFF
--- a/Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
+++ b/Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
@@ -26,7 +26,7 @@ SOFTWARE.
 #ifndef __cplusplus
 // If you get this error building an Objective-C project, #include neither this file nor "MidichloriansImpl.h".
 // Use "Midichlorians.h" instead.
-// It wraps the inner, platform-independent, event tracking code, and allows client code to stay `.m`.
+// It wraps the inner, platform-independent, event tracking code, and allows the client code to stay `.m`.
 #error "This C++ header should be `#include`-d or `#import`-ed from an `.mm`, not an `.m` source file."
 #endif
 
@@ -51,10 +51,7 @@ CURRENT_STRUCT(MidichloriansBaseEvent) {
 };
 
 // iOS-specific base event.
-CURRENT_STRUCT(iOSBaseEvent, MidichloriansBaseEvent) {
-  // TODO: do we need this field?
-  CURRENT_FIELD(description, std::string);
-};
+CURRENT_STRUCT(iOSBaseEvent, MidichloriansBaseEvent){};
 
 // iOS event denoting the very first application launch.
 // Emitted once during Midichlorians initialization.
@@ -65,17 +62,15 @@ CURRENT_STRUCT(iOSFirstLaunchEvent, iOSBaseEvent){};
 CURRENT_STRUCT(iOSAppLaunchEvent, iOSBaseEvent) {
   CURRENT_FIELD(binary_version, std::string);
   CURRENT_FIELD(cf_version, std::string);
-  CURRENT_FIELD(app_install_time, uint64_t);
-  CURRENT_FIELD(app_update_time, uint64_t);
+  CURRENT_FIELD(app_install_time, uint64_t, 0ull);
+  CURRENT_FIELD(app_update_time, uint64_t, 0ull);
   CURRENT_DEFAULT_CONSTRUCTOR(iOSAppLaunchEvent) {}
   CURRENT_CONSTRUCTOR(iOSAppLaunchEvent)(
       const std::string& cf_version, uint64_t app_install_time, uint64_t app_update_time)
       : binary_version(__DATE__ " " __TIME__),
         cf_version(cf_version),
         app_install_time(app_install_time),
-        app_update_time(app_update_time) {
-    iOSBaseEvent::description = "*FinishLaunchingWithOptions";
-  }
+        app_update_time(app_update_time) {}
 };
 
 // iOS event with the detailed device information.
@@ -97,19 +92,18 @@ CURRENT_STRUCT(iOSGenericEvent, iOSBaseEvent) {
   CURRENT_FIELD(event, std::string);
   CURRENT_FIELD(source, std::string);
   CURRENT_FIELD(fields, (std::map<std::string, std::string>));
-  // TODO: do we need this separation to `fields` + `complex_fields`?
-  CURRENT_FIELD(complex_fields, (std::map<std::string, std::string>));
   CURRENT_FIELD(unparsable_fields, std::vector<std::string>);
   CURRENT_DEFAULT_CONSTRUCTOR(iOSGenericEvent) {}
 #ifdef COMPILE_MIDICHLORIANS_DATA_DICTIONARY_FOR_IOS_CLIENT
-  CURRENT_CONSTRUCTOR(iOSGenericEvent)(NSString * input_event, NSString * input_source, NSDictionary * input) {
+  CURRENT_CONSTRUCTOR(iOSGenericEvent)(
+      NSString * event_name, NSString * event_source, NSDictionary * properties) {
     // A somewhat strict yet safe way to parse dictionaries. Imperfect but works. -- D. K.
-    event = [input_event UTF8String];
-    source = [input_source UTF8String];
-    for (NSString* k in input) {
+    event = [event_name UTF8String];
+    source = [event_source UTF8String];
+    for (NSString* k in properties) {
       const char* key = (k && [k isKindOfClass:[NSString class]]) ? [k UTF8String] : nullptr;
       if (key) {
-        NSObject* o = [input objectForKey:k];
+        NSObject* o = [properties objectForKey:k];
         NSString* v = (NSString*)o;
         const char* value = (v && [v isKindOfClass:[NSString class]]) ? [v UTF8String] : nullptr;
         if (value) {
@@ -118,7 +112,7 @@ CURRENT_STRUCT(iOSGenericEvent, iOSBaseEvent) {
           NSString* d = [o description];
           const char* value = (d && [d isKindOfClass:[NSString class]]) ? [d UTF8String] : nullptr;
           if (value) {
-            complex_fields[key] = value;
+            fields[key] = value;
           } else {
             unparsable_fields.push_back(key);
           }
--- a/Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
+++ b/Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
@@ -1,0 +1,142 @@
+/*******************************************************************************
+The MIT License (MIT)
+
+Copyright (c) 2015 Maxim Zhurovich <zhurovich@gmail.com>
+          (c) 2015 Dmitry "Dima" Korolev <dmitry.korolev@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+#ifndef __cplusplus
+// If you get this error building an Objective-C project, #include neither this file nor "MidichloriansImpl.h".
+// Use "Midichlorians.h" instead.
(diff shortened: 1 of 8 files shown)


Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+#ifndef __cplusplus
+// If you get this error building an Objective-C project, #include neither this file nor "MidichloriansImpl.h".
+// Use "Midichlorians.h" instead.
+// It wraps the inner, platform-independent, event tracking code, and allows client code to stay `.m`.
```
dkorolev: Nit: `allows the client code`. Probably in both old and new sources.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+#error "This C++ header should be `#include`-d or `#import`-ed from an `.mm`, not an `.m` source file."
+#endif
+
+#ifndef CURRENT_MIDICHLORIANS_DATA_DICTIONARY_H
```
dkorolev: Should these include `BETA2`, or shall we be retiring the old code same time as checking this in?

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+// iOS event denoting the very first application launch.
+// Emitted once during Midichlorians initialization.
+CURRENT_STRUCT(iOSFirstLaunchEvent, iOSBaseEvent){};
+
+// iOS event with the detailed application information.
+// Emitted once during Midichlorians initialization.
+CURRENT_STRUCT(iOSAppLaunchEvent, iOSBaseEvent) {
+  CURRENT_FIELD(binary_version, std::string);
+  CURRENT_FIELD(cf_version, std::string);
+  CURRENT_FIELD(app_install_time, uint64_t);
```
dkorolev: Defaulting these two to `0ll` would be nice.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+        cf_version(cf_version),
+        app_install_time(app_install_time),
+        app_update_time(app_update_time) {
+    iOSBaseEvent::description = "*FinishLaunchingWithOptions";
+  }
+};
+
+// iOS event with the detailed device information.
+// Emitted once during Midichlorians initialization.
+CURRENT_STRUCT(iOSDeviceInfo, iOSBaseEvent) { CURRENT_FIELD(info, (std::map<std::string, std::string>)); };
```
dkorolev: Could you use a `typedef` or `using` please? The parentheses don't work on VC++/Windows, and we'll probably compile the client with it at least once.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+  CURRENT_FIELD(source, std::string);
+  CURRENT_FIELD(fields, (std::map<std::string, std::string>));
+  // TODO: do we need this separation to `fields` + `complex_fields`?
+  CURRENT_FIELD(complex_fields, (std::map<std::string, std::string>));
+  CURRENT_FIELD(unparsable_fields, std::vector<std::string>);
+  CURRENT_DEFAULT_CONSTRUCTOR(iOSGenericEvent) {}
+#ifdef COMPILE_MIDICHLORIANS_DATA_DICTIONARY_FOR_IOS_CLIENT
+  CURRENT_CONSTRUCTOR(iOSGenericEvent)(NSString * input_event, NSString * input_source, NSDictionary * input) {
+    // A somewhat strict yet safe way to parse dictionaries. Imperfect but works. -- D. K.
+    event = [input_event UTF8String];
```
dkorolev: These too can be null pointers, can't they?

dkorolev: If yes, perhaps `Optional<std::string>` is a more appropriate type for `event` and `source`.

Midichlorians/Dev/Beta2/iOS/Midichlorians.mm
```diff
@@ -0,0 +1,46 @@
...
+    [MidichloriansImpl setup:serverUrl withLaunchOptions:options];
+}
+
++ (void)identify:(NSString *)identifier {
+    [MidichloriansImpl identify:identifier];
+    [MidichloriansImpl emit:iOSIdentifyEvent()];
+}
+
++ (void)focusEvent:(BOOL)hasFocus source:(NSString *)source {
+    [MidichloriansImpl emit:iOSFocusEvent(static_cast<bool>(hasFocus), [source UTF8String])];
```
dkorolev: Possible null pointer here?

Midichlorians/Dev/Beta2/iOS/MidichloriansImpl.mm
```diff
@@ -0,0 +1,357 @@
...
+#import <UIKit/UIScreen.h>
+#import <UIKit/UIApplication.h>
+#import <AdSupport/ASIdentifierManager.h>  // Need to add AdSupport.Framework to the project.
+#endif  // (TARGET_OS_IPHONE > 0)
+
+namespace current {
+namespace midichlorians {
+
+    // Conversion from [possible nullptr] `const char*` to `std::string`.
+    static std::string UnsafeToStdString(const char* s) {
```
dkorolev: Actually, it's `Safe`, not `Unsafe` :-)

Midichlorians/Dev/Beta2/iOS/MidichloriansImpl.mm
```diff
@@ -0,0 +1,357 @@
...
+        return s ? s : "";
+    }
+
+#if (TARGET_OS_IPHONE > 0)
+    // Conversion from [possible nil] `NSString` to `std::string`.
+    static std::string UnsafeToStdString(NSString *nsString) {
+        if (nsString) {
```
dkorolev: Same `nsString ? ... : ""` would do it here too.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+// Base event.
+CURRENT_STRUCT(MidichloriansBaseEvent) {
+  CURRENT_FIELD(user_ms, std::chrono::milliseconds);
+  CURRENT_FIELD(device_id, std::string);
+  CURRENT_FIELD(client_id, std::string);
+};
+
+// iOS-specific base event.
+CURRENT_STRUCT(iOSBaseEvent, MidichloriansBaseEvent) {
+  // TODO: do we need this field?
```
dkorolev: Indeed, looks like we don't.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+  CURRENT_FIELD(source, std::string);
+  CURRENT_CONSTRUCTOR(iOSFocusEvent)(bool gained_focus = false, const std::string& source = "")
+      : gained_focus(gained_focus), source(source) {}
+};
+
+CURRENT_STRUCT(iOSGenericEvent, iOSBaseEvent) {
+  CURRENT_FIELD(event, std::string);
+  CURRENT_FIELD(source, std::string);
+  CURRENT_FIELD(fields, (std::map<std::string, std::string>));
+  // TODO: do we need this separation to `fields` + `complex_fields`?
```
dkorolev: Let's keep it for now. Can make `complex_fields` `Optional`. It's because some times are not string-conversible; we'd love to keep them nonetheless, but their format may change in later releases.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
mzhurovich: I was thinking of retiring, but existing CTFO users will continue to use the old version. Should we support both in our new server?

dkorolev: I vote to retire it. The new CTFO binary will use the new backend, and the old one will slowly fade away, server-wise :)

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
mzhurovich: This is pure iOS part, it should never be compiled under Windows. Maybe move `iOS*` events to a separate header?

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
mzhurovich: They ARE convertible. For non-convertible ones we have `unparsable_fields`.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
mzhurovich: Theoretically yes, but it means some weird and incorrect usage of our API. Maybe we should ignore such kind of call and `NSLog` this fact?

Midichlorians/Dev/Beta2/iOS/MidichloriansImpl.mm
mzhurovich: Maybe rename it to `ToStdString`?

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
dkorolev: The iOS event collection part will likely not be compiled on Windows, but the analytics on those might well. Why not?

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
dkorolev: Hmm.
Hmmmmm.

So, `complex_fields` are the return values of `description` for records that are not `string`-s.

I see your point. What bugs me is that if the user changes the fields from a custom object (on which `description`)is being invoked) to a string of their own format (which is passed natively), this change could possibly go unnoticed.

Given a) the "message format" itself is a `map<string, string>`, and b) binary version is part of the first logged event (it it?), I hereby declare by worry groundless, and suggest we go with your proposal of retiring `complex_fields`.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
dkorolev: +1

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
```diff
@@ -0,0 +1,142 @@
...
+  CURRENT_FIELD(complex_fields, (std::map<std::string, std::string>));
+  CURRENT_FIELD(unparsable_fields, std::vector<std::string>);
```
dkorolev: Did you change `unparsable_fields` from `set` into `vector` since we don't yet support `set` in `CURRENT_STRUCT`, or is there a deeper underlying reason?

Midichlorians/Dev/Beta2/iOS/MidichloriansImpl.mm
```diff
@@ -0,0 +1,357 @@
...
+            static T instance;
+            return instance;
+        }
+    };
+    
+    using Stats = consumer::POSTviaHTTP;
+
+    struct EventsVariantDispatchedAssigner {
+        explicit EventsVariantDispatchedAssigner(Variant<T_IOS_EVENTS>& variant) : variant_(variant) {}
+        void operator()(const iOSFirstLaunchEvent& event) { variant_ = event; }
```
dkorolev: I believe a single`template`-d method would do it here.

Midichlorians/Dev/Beta2/iOS/MidichloriansImpl.mm
```diff
@@ -0,0 +1,357 @@
...
+#else
+    static_cast<void>(options);
+#endif  // (TARGET_OS_IPHONE > 0)
+}
+
++ (void)emit:(const iOSBaseEvent &)event {
+    Stats &instance = Singleton<Stats>::Instance();
+    Variant<T_IOS_EVENTS> v;
+    EventsVariantDispatchedAssigner assigner(v);
+    current::metaprogramming::RTTIDynamicCall<T_IOS_EVENTS>(event, assigner);
```
dkorolev: What am I missing assuming `v = event;`, or `Variant<T_OS_EVENTS> v(event);` should just work?

dkorolev:
@mzhurovich Looks good, a couple design-ish notes to think of. Thanks!

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
mzhurovich: No more reasons, just because of still not supported `set` :(

dkorolev: Ack: https://trello.com/c/KR7BrOuT/110-support-set-and-unordered-set-in-current-struct

Midichlorians/Dev/Beta2/iOS/MidichloriansImpl.mm
mzhurovich: This syntax doesn't work - it tries to put an object of `iOSBaseEvent` type into Variant and fails to compile due to `iOSBaseEvent` missing in the type list.

dkorolev: I see. Well, the `template<>` declaration of that helper member function should still do it.

Midichlorians/Dev/Beta2/MidichloriansDataDictionary.h
mzhurovich: Ok, so I keep this file as is.

mzhurovich:
@dkorolev PTAL

PR #329, opened 2015-12-25
New, ugly, but working in VS2015, way to treat parenthesized types.

mzhurovich:
@dkorolev Check with your Studio, pls.


dkorolev:
IT WORKS ON MSVS!!!!111

dkorolev:
LGTM+

PR #330, opened 2015-12-27
QPS measurement.

dkorolev:
Hi @mzhurovich,

The tool I put together yesterday to measure C++ vs. F# QPS (spoiler alert: 10x).

It's still threaded; I coudn't make an `std::async()` version as clean and as robust as the code below.

Eventually, we'll probably unify the piece that does benchmarking -- now this code intersects a lot with the one from `EmbeddedDB`. For now, I suggest we live with this duplication.

Thanks,
Dima

### DIFF
--- a/examples/Benchmark/Makefile
+++ b/examples/Benchmark/Makefile
@@ -1,0 +1,1 @@
+../../scripts/Makefile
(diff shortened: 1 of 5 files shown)


mzhurovich:
LGTM!

PR #334, opened 2016-01-05
Key-only deleters in Storage.

mzhurovich:


Storage/storage.h
```diff
@@ -78,11 +76,14 @@ namespace storage {
 #define CURRENT_STORAGE_STRUCT_TAG(base, alias)                                               \
   CURRENT_STRUCT(CURRENT_STORAGE_ADDER_##alias, base) {                                       \
     CURRENT_DEFAULT_CONSTRUCTOR(CURRENT_STORAGE_ADDER_##alias) {}                             \
-    CURRENT_CONSTRUCTOR( CURRENT_STORAGE_ADDER_##alias)(const base& value) : SUPER(value) {}  \
+    CURRENT_CONSTRUCTOR(CURRENT_STORAGE_ADDER_##alias)(const base& value) : SUPER(value) {}   \
   };                                                                                          \
-  CURRENT_STRUCT(CURRENT_STORAGE_DELETER_##alias, base) {                                     \
+  CURRENT_STRUCT(CURRENT_STORAGE_DELETER_##alias) {                                           \
+    CURRENT_FIELD(key, ::current::storage::sfinae::ENTRY_KEY_TYPE<base>); \
```
dkorolev: `s/key/key_to_erase/` ?

Storage/storage.h
```diff
@@ -78,11 +76,14 @@ namespace storage {
...
     CURRENT_DEFAULT_CONSTRUCTOR(CURRENT_STORAGE_DELETER_##alias) {}                           \
-    CURRENT_CONSTRUCTOR(CURRENT_STORAGE_DELETER_##alias)(const base& value) : SUPER(value) {} \
+    CURRENT_CONSTRUCTOR(CURRENT_STORAGE_DELETER_##alias)(const base& value) { \
```
dkorolev: Why assignment, not initialization?

Storage/test.cc
```diff
@@ -48,6 +48,8 @@ namespace transactional_storage_test {
 CURRENT_STRUCT(Element) {
   CURRENT_FIELD(x, int32_t);
   CURRENT_CONSTRUCTOR(Element)(int32_t x = 0) : x(x) {}
+  // Dummy key function to support key-only deleter.
+  int32_t key() const { return x; }
```
dkorolev: Shouldn't `Element` already have it own key field/method defined to be used in the storage?

On the general note, why not have some `CURRENT_PRIMARY_KEY(...);` (and `CURRENT_PRIMARY_KEY_METHOD()` macro, much like we have `CURRENT_TIMESTAMP()`?

Storage/storage.h
mzhurovich: For two reasons:
1) It's shorter
2) `map_.erase(sfinae::GetKey(object));` works out of the box :)

Storage/storage.h
mzhurovich: Just thought it's more readable.

Storage/test.cc
mzhurovich: Speaking of must-have key field - generally yes, but this structure is used in Vector, which doesn't need keys. Are we going to deprecate it?
In general, we'll need composite keys for matrices. So, let's return to macro design question when we implement `ManyToMany` etc.

Storage/storage.h
dkorolev: Sold :+1:

Storage/storage.h
dkorolev: In the future far, far away, `key` might not be default-constructible.

Storage/test.cc
dkorolev: If the issue is with `Vector` only, which is going away, I'm fine not thinking of it now.

Storage/storage.h
mzhurovich: Ok, changed to init-list.

Storage/test.cc
mzhurovich: Yep, it's only for Vector.

dkorolev:
LGTM!

PR #336, opened 2016-01-07
Minor tweaks.

dkorolev:
Hi @mzhurovich ,

Two minor tweaks:
1. "Moved" our `make_unique<>` into `std::`. This pleases Visual Studio, and, to the best of my understanding, does not violate the Standard when compiled in C++14, that has its own `std::make_unique<>`.
2. Renamed `.json` files into a different extension, as the top-level `make test` has more `.json`-s in `golden/`, which makes the schema inference test fail.

Thanks!
Dima


mzhurovich:
@dkorolev all green!

PR #340, opened 2016-01-17
 Zero-copy `strings::Split()`.

dkorolev:


mzhurovich:
@dkorolev LGTM!
BTW, what do you think about the use case, when the original string is left in modified state and you have a bunch of `Chunk`s for more sophisticated processing?

dkorolev:
@mzhurovich Useful usecase! Need to think how to best integrate it. Perhaps add some `EfficientModifyingSplit(...)` ?

PR #341, opened 2016-01-17
Member pointer extraction by our reflector.

dkorolev:
[ after https://github.com/c5t/Current/pull/340 ]

Enabled the reflector to emit field pointers.

Thanks,
Dima


dkorolev:
@mzhurovich Made the `names` change in the forthcoming pull request.

PR #344, opened 2016-01-18
Minor tweaks.

dkorolev:
[ after https://github.com/c5t/Current/pull/343 ]

Integration work.

Thanks,
Dima


mzhurovich:
@dkorolev LGTm as well :)

PR #346, opened 2016-01-18
Storage RESTful API design doc.

mzhurovich:
Hi, @dkorolev!
I took the liberty of summarizing our Google Docs drafts into this design doc.
PTAL, if everything is described correctly.
Fixes/additions/comments are highly appreciated :)
Thanks!

### DIFF
--- a/Storage/REST-API.md
+++ b/Storage/REST-API.md
@@ -1,0 +1,183 @@
+# RESTful binding for Storage, RFC
+
+This document describes the RESTful API with CRUD functionality for Current Storage.
+
+### Bindings semantics
+`TODO: Describe storage field <-> endpoint bindings.`
+
+### Response schema
+
+The response body is always either a valid JSON or empty.
+
+The generic format of a successful JSON response is:
+
+```
+{
+  "url": "https://api.example.com/...",
+  "url_*": "https://api.example.com/...",
+  ... more "url_*" fields ...
+  
+  [optional]
+  "meta": {
+    ...
+  },
+  
+  [required]
+  "data": {
+    ...
+  }
+
+}
+```
+
+The responses are strongly typed. The contents of the request URL define the schema for the response.
+
+The value for `"data"` can be:
+* an object,
+* an array, if the response is expected to contain a list,
+* primitive type (string, number, boolean) in special cases.
+
+Error responses are JSON objects with the top-level `"error":...` field:
+
+```
+{
+  “error”: “Detailed human-readable error message.”,
+  ... other fields ...
+}
+```
+
+Successful responses are JSON objects without the top-level `"error"` field.
+
+`TODO: We also may have redirects, which contain the "Location" in the header, with empty body.`
+
+### Notation
+
+The default way to encode JSON objects is JavaScript-friendly. The API also supports F#-friendly format.
+
+The differences are:
+
+1. Variant types.
+
+   If a field `x` can be either an `apple` or an `orange`, the JavaScript-friendly representation is `{...,"x":{"apple":{...},...}`, and the F#-friendly representation is `{...,"x":{"Case":"apple","Fields":[...]},...}}`.
+
+2. Coding of absent optional objects.
+
+   The JavaScript-friendly implementation omits missing fields from the output (and does not expect them in the input).
+The F#-friendly implementation outputs missing fields as `null`-s (and expects `null`-s in the input).
+
+3. Coding of dictionaries with the keys being strings.
+
+   Generally, both JavaScript- and F#-friendly implementations encode dictionaries as arrays of arrays of two elements: `[[key1,value1],[key2,value2],...]`.
+   
+   However, if the type of `key` is string, the JavaScript-friendly format of this dictionary becomes an object: `{"key1":value1,"key2":value2,...}`.
+
+   `TODO(dkorolev): Check if some other F#-related issues should be added into this section.`
+
+### Views
+
+**View** is a representation of a record, which contains some non-empty subset of the record’s fields.
+The API supports different views via the `&format=...` URL parameter.
+
+By default, records returned by the API contain all fields.
+
+Two implicitly defined views (both default to "all fields") are `full` and `preview`.
+
+The `full` view is the default, so that `&format=full` is equivalent to not specifying the `&format` URL parameter.
+
+The `&format=preview` is used when the response schema for certain endpoint contains not one record, but a collection of them. In other words, the representation of each record of `"data":[...]` for the endpoints where `"data"` is an array is equivalent to the representation of those records with `&format=preview`.
+
+The views are defined by the user, and the default representation of any view is "all fields". In particular, unless the user defines the `"preview"` view, it will be identical to the `"full"` one.
+
+### Methods
+
+Current Storage API interprets HTTP verbs in the following way:
+
+* **GET**: Get the resource.
+
+  Strongly typed with respect to returning an individual record or a collection of records. For example, `/users` can be the way to GET the list of users (`"data":[...]`), and `/user/1` can be the way to GET an individual user with the ID of `1`.
+   
+* **PATCH**: Update some fields of the resource.
+
+  `PATCH` should be called on the URL that has been (or could have been) returned as the top-level `"url":"..."` field of the payload of `GET`.
+
+  Generally, the user is expected to retrieve the value of the object under the key `"data"` of the top-level payload of a `GET` request, alter this object, and send its modified version back to the server.
+
+  `PATCH` respects the format of the `"data"` controlled by the `&format=` URL parameter. Sending a `PATCH` request to the URL with certain `&format=X` can only modify the fields returned by the `GET` request of this `&format=X`.
+
+  For `PATCH` requests, the key of the record being updated is taken from the URL. The user is expected to either not send in the key as part of the payload, or ensure the key contained in the payload matches the key contained in the URL.
+  
+  `PATCH` will fail if attempted to modify a non-existing resource, resulting in `404 (Not Found)` error code.
+  
+  Also, `PATCH` performs conflict resolution. The URL being `PATCH`-ed must contain a revision / timestamp of the resource. The exact semantics of the revision / timestamp is not specified by this RFC, but the top-level `"url"` field returned by the `GET` request is guaranteed to have it set.
+
+  An attempt to `PATCH` to a URL that points to the version of the resource other than the most recent one will result in no data mutation and a `409 Conflict` error code. The user is then expected to re-run the `GET` request, followed by the re-run `PATCH` request onto the URL of the most recent version of the resource returned by the `GET` request. Confirming that the modifications that occurred in the meantime do not conflict with the new payload being sent in the body of the `PATCH` request is then the responsibility of the API user.
+
+* **PUT**: Create a new record for a resource, or update all fields of an existing resource.
+
+  Attempting to `PUT` to a URL that has an incomplete view of a record would result in a `405 Method Not Allowed` error.
+  
+  In the `&format=full` view, `PUT` is equivalent to `PATCH`. 
+  
+  `PUT` behaves the same way as `PATCH` when it comes to avoiding merge conflicts.
+
+  When legal, the only difference between `PUT` and `PATCH` is that `PUT` will create a new resource if it did not exist before the call. A `201 Created` response code will then be returned. Use `PATCH` if this behavior is undesirable.
+
+  `TODO: Location header https://tools.ietf.org/html/rfc7231#section-4.3.3`
+
+* **POST**: Create a new record and assign it to some recourse.
+
+  `POST` will create a new resource. The URL for POST must not contain the key(s) for the record being posted, and the payload body should not contain the key(s) either.
+
+  The generation of the new key is user-defined. Three common strategies are: 1) random, 2) the [salted] hash of the payload, 3) the value of certain passed in field or fields.
+  
+  In the latter two, as well as other, cases, a `POST` request may result in an attempt to overwrite an already existing resource. By default, `POST` will not overwrite, but return a `409 Conflict` error code, with the top-level `"url"` field of the returned JSON object containing the URL for the resource that was about to be overwritten.
+
+  The `&overwrite=1` / `&overwrite=true` URL parameter can be set to indicate that the `POST` request should complete successfully even if it would result in overwriting existing data.
+
+Successful `POST` results in either `200 (OK)` error code if the resource has been updated, or `201 (Created)`.
+
+* **DELETE**: Deletes the resource.
+
+   Similarly to `PATCH` and `PUT`, `DELETE` should be called on the URL that has been (or could have been) returned as the top-level `"url":"..."` field of the payload of `GET`.
+   
+   `DELETE` follows the same conflict resolution rules as `PATCH` and `PUT`.
+   
+`TODO: User-defined data integrity checks, and what errors are returned if they fail?`
+
+
+### Discoverability
+
+The entry point (usually, `"/"`) URL will contain the list of inner `"url_*"`-s to access respective fields ("tables") of the storage.
+
+The entry point URL may also contain the URLs by which the user can access the respective schema files. This is applicable to F# and C++.
+
+[ TO DISCUSS ] The payload when `GET`-ting the resource corresponding to an individual record (ex. `/user/123`) will contain the `{...,"url_collection":".../users"}` reference to the URL to browse the whole collection, the record from which is being requested.
+
+As a record is successfully created or modified, the response payload will contain its URL as the top-level `"url"` field.
+
+When a resource can not be found, the error payload will contain the URL to browse the respective collection.
+
+When the operation fails due to a conflict, the top-level `"url"` field will be set as well.
+
+### Pagination
+
+When retrieving the collection of records, they are returned in pages. If there are more records to browse, the top-level response object will contain the `"url_next_page"` field, as well as the `"url_previous_page"` one for the pages past the first one.
+
+As mentioned above, in the pagination mode:
+* the `"data":[...]` top-level field becomes an array, not object, and
+* the elements of this array default to the `&format=preview` format, if defined by the user.
+
+With respect to pagination:
+* The user SHOULD NOT attempt to skip through records (ex. browse page 1 with records 1..20, and then skip immediately to records 101..120.
+* The user CAN expect the records they have already seen as part of this session will not change.
+* The user SHOULD NOT expect any record they have seen or may see will exist and/or be available to them by the time the user decides to perform actions on that record.
+
+In other words, the behavior of pagination for the user mimics the architecture where on the first `GET` request to browse the collection, the complete list of possible IDs is populated on the server, and the user is just browsing through these IDs.
+
+The server will honor the entries the user has already browsed through, even if the user requests the previous page, and some entries from it have been deleted. At the same time, while browsing further, the server may re-fetch the records and re-apply the filters.
+
+In particular, scanning through pages CAN return records created AFTER the time of the first `GET` request, unless the explicit filter "created before X" has been requested by the user.
+
+The token returned by the API to page through the collection expires by itself. The default period for which the token will be live is 10 minutes since it was last used.
+
+`TODO: Document page size and the ability to dynamically change it.`
--- a/Storage/REST-API.md
+++ b/Storage/REST-API.md
@@ -7,7 +7,7 @@ This document describes the RESTful API with CRUD functionality for Current Stor
 
 ### Response schema
 
-The response body is always either a valid JSON or empty.
+The response body is always either a valid JSON or empty with `204 (No content)` error code.
 
 The generic format of a successful JSON response is:
 
@@ -124,7 +124,7 @@ Current Storage API interprets HTTP verbs in the following way:
 
   `TODO: Location header https://tools.ietf.org/html/rfc7231#section-4.3.3`
 
-* **POST**: Create a new record and assign it to some recourse.
+* **POST**: Create a new record and assign it to some resourse.
 
   `POST` will create a new resource. The URL for POST must not contain the key(s) for the record being posted, and the payload body should not contain the key(s) either.
 
@@ -151,7 +151,7 @@ The entry point (usually, `"/"`) URL will contain the list of inner `"url_*"`-s 
 
 The entry point URL may also contain the URLs by which the user can access the respective schema files. This is applicable to F# and C++.
 
-[ TO DISCUSS ] The payload when `GET`-ting the resource corresponding to an individual record (ex. `/user/123`) will contain the `{...,"url_collection":".../users"}` reference to the URL to browse the whole collection, the record from which is being requested.
+[ TO DISCUSS ] The payload when `GET`-ting the resource corresponding to an individual record (ex. `/user/123`) will contain the `{...,"url_collection":".../users"}` reference to the URL to browse the whole collection, the record from which is being requested. Each resource can belong to one and only one collection.
 
 As a record is successfully created or modified, the response payload will contain its URL as the top-level `"url"` field.
 
@@ -160,6 +160,8 @@ When a resource can not be found, the error payload will contain the URL to brow
 When the operation fails due to a conflict, the top-level `"url"` field will be set as well.
 
 ### Pagination
+
+`TODO: Revisit this section - discuss stateful/stateless approaches.`
 
 When retrieving the collection of records, they are returned in pages. If there are more records to browse, the top-level response object will contain the `"url_next_page"` field, as well as the `"url_previous_page"` one for the pages past the first one.
 


Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
...
+The value for `"data"` can be:
+* an object,
+* an array, if the response is expected to contain a list,
+* primitive type (string, number, boolean) in special cases.
+
+Error responses are JSON objects with the top-level `"error":...` field:
+
+```
+{
+  “error”: “Detailed human-readable error message.”,
```
sompylasar: @mzhurovich  What do you think of making `error` an object with a human-readable `message` field and other error detail fields next to it, like the exception name (code)? Error objects are quite standard, some APIs even respond with arrays of error objects. Otherwise, to reliably determine an error condition from the response, you have to guarantee that the human-readable message is always present and non-empty.

Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
+# RESTful binding for Storage, RFC
+
+This document describes the RESTful API with CRUD functionality for Current Storage.
+
+### Bindings semantics
+`TODO: Describe storage field <-> endpoint bindings.`
+
+### Response schema
+
+The response body is always either a valid JSON or empty.
```
sompylasar: The empty body should be declared a (very limited) special case. Both successful and unsuccessful responses should always render a JSON.

Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
...
+
+By default, records returned by the API contain all fields.
+
+Two implicitly defined views (both default to "all fields") are `full` and `preview`.
+
+The `full` view is the default, so that `&format=full` is equivalent to not specifying the `&format` URL parameter.
+
+The `&format=preview` is used when the response schema for certain endpoint contains not one record, but a collection of them. In other words, the representation of each record of `"data":[...]` for the endpoints where `"data"` is an array is equivalent to the representation of those records with `&format=preview`.
+
+The views are defined by the user, and the default representation of any view is "all fields". In particular, unless the user defines the `"preview"` view, it will be identical to the `"full"` one.
```
sompylasar: Can the user define a custom view, beyond `full` and `preview`?

Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
...
+
+3. Coding of dictionaries with the keys being strings.
+
+   Generally, both JavaScript- and F#-friendly implementations encode dictionaries as arrays of arrays of two elements: `[[key1,value1],[key2,value2],...]`.
+   
+   However, if the type of `key` is string, the JavaScript-friendly format of this dictionary becomes an object: `{"key1":value1,"key2":value2,...}`.
+
+   `TODO(dkorolev): Check if some other F#-related issues should be added into this section.`
+
+### Views
```
sompylasar: Just curious, what do you think of [GraphQL](http://facebook.github.io/graphql/#sec-Overview)?

Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
...
+  
+  In the `&format=full` view, `PUT` is equivalent to `PATCH`. 
+  
+  `PUT` behaves the same way as `PATCH` when it comes to avoiding merge conflicts.
+
+  When legal, the only difference between `PUT` and `PATCH` is that `PUT` will create a new resource if it did not exist before the call. A `201 Created` response code will then be returned. Use `PATCH` if this behavior is undesirable.
+
+  `TODO: Location header https://tools.ietf.org/html/rfc7231#section-4.3.3`
+
+* **POST**: Create a new record and assign it to some recourse.
```
sompylasar: `s/recourse/resource/`

Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
...
+
+* **PUT**: Create a new record for a resource, or update all fields of an existing resource.
+
+  Attempting to `PUT` to a URL that has an incomplete view of a record would result in a `405 Method Not Allowed` error.
+  
+  In the `&format=full` view, `PUT` is equivalent to `PATCH`. 
+  
+  `PUT` behaves the same way as `PATCH` when it comes to avoiding merge conflicts.
+
+  When legal, the only difference between `PUT` and `PATCH` is that `PUT` will create a new resource if it did not exist before the call. A `201 Created` response code will then be returned. Use `PATCH` if this behavior is undesirable.
```
sompylasar: Will it create the new resource with the key sent from the client, or with a generated key, which will be returned in the response?

Storage/REST-API.md
dkorolev: As of now, we return context-dependent fields alongside `"error"` (ex. `{"error":"Not resource not found","url_requested":"https:/...",...}` I'm fine with `error` being an object with a required `message` field.

@sompylasar, can you cite some examples? Extra nesting looks a unnecessarily complex to me, especially given that HTTP response code is usually sufficient for programmatic access.

CC @nyospe.

Storage/REST-API.md
dkorolev: +1
Unless it's `204 No Content`.

Storage/REST-API.md
dkorolev: Yes. We're still thinking of the best way to introduce them into C++. Slowly converging to field-level custom user-defined tags.

CC @nyospe

Storage/REST-API.md
dkorolev: `PUT` is when the user knows the key for the newly created object. `POST` is when the user expects the key to be assigned by the server.

Storage/REST-API.md
sompylasar: HTTP response code is not always available for programmatic access, that's why it should be mirrored in the response. [A longer explanation](http://labs.omniti.com/labs/jsend#WhitherHTTP)

https://developers.google.com/doubleclick-search/v2/standard-error-responses
http://jsonapi.org/format/#error-objects
http://stackoverflow.com/questions/12806386/standard-json-api-response-format

dkorolev: Doesn't look convincing to me.

Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
...
+`TODO: User-defined data integrity checks, and what errors are returned if they fail?`
+
+
+### Discoverability
+
+The entry point (usually, `"/"`) URL will contain the list of inner `"url_*"`-s to access respective fields ("tables") of the storage.
+
+The entry point URL may also contain the URLs by which the user can access the respective schema files. This is applicable to F# and C++.
+
+[ TO DISCUSS ] The payload when `GET`-ting the resource corresponding to an individual record (ex. `/user/123`) will contain the `{...,"url_collection":".../users"}` reference to the URL to browse the whole collection, the record from which is being requested.
```
sompylasar: Can a resource belong to more than one collections? If yes, which collection reference will be returned?

dkorolev: The collection URL is not returned as part of the resource (i.e., not within `"data":{...}`), but as part of the metadata (i.e., alongside `"url":"..."`).

Storage/REST-API.md
sompylasar: OK then.

Storage/REST-API.md
sompylasar: Anyway, the response contains the resource representation, and next to it there is one URL pointing to one collection. Can this represented resource be a part of more than one collection?

dkorolev: Out of context, I'm leaning towards "no, because de-denormalization". The resource belongs to one collection; other resources may point to it.

In the RESTful world, however, when there is a reference from one resource to another, instead of just returning the URL, we may return the object containing the `preview` format (which obviously includes the `.url` field too).

Storage/REST-API.md
```diff
@@ -0,0 +1,183 @@
...
+* the elements of this array default to the `&format=preview` format, if defined by the user.
+
+With respect to pagination:
+* The user SHOULD NOT attempt to skip through records (ex. browse page 1 with records 1..20, and then skip immediately to records 101..120.
+* The user CAN expect the records they have already seen as part of this session will not change.
+* The user SHOULD NOT expect any record they have seen or may see will exist and/or be available to them by the time the user decides to perform actions on that record.
+
+In other words, the behavior of pagination for the user mimics the architecture where on the first `GET` request to browse the collection, the complete list of possible IDs is populated on the server, and the user is just browsing through these IDs.
+
+The server will honor the entries the user has already browsed through, even if the user requests the previous page, and some entries from it have been deleted. At the same time, while browsing further, the server may re-fetch the records and re-apply the filters.
```
sompylasar: Does the server guarantee that the records that the user has seen on some pages won't appear on the other pages?

Storage/REST-API.md
sompylasar: Agree. I think this one-collection-per-resource decision should be somehow pointed out in the spec, even though it looks standard in the RESTful world.

Storage/REST-API.md
dkorolev: + @mzhurovich -- Excellent question. We're undecided as of now.

One inclination is to have Current REST as user-friendly as possible, in which case the server will guarantee no duplicates. This, however, makes the logic stateful. Imagine the binary restarts between pages -- where does it get the state from?

Another approach is to remove pagination altogether, and have each request independent. In this case, some "&skip=20" would not guarantee pretty much anything.

No conclusion yet. Thoughts and references welcome.

Storage/REST-API.md
dkorolev: Good point, cross-referencing between resources should be documented.

nyospe: My inclination is to have an "as of change id" field embedded in the URL as
well, and to buffer results in a local session for a period of time for
performance reasons. If the session goes out of scope, the original result
set can be reconstructed exactly, thereby meeting the RESTful requirements,
but the optimization for iterative behavior remains in scope while
follow-up requests are most likely. If a session-invalidating event occurs
(e.g. a change in permissions) then "next N results" could change
radically, but this is not a typical occurrence, and can be reasonably
ignored.

dkorolev: Adding an "as of change ID" timestamp in the URL is the easiest way to ensure REST-fulness and statelessness ...

... but it's a nightmare to implement, as effectively we would be requiring the DB to not only return existing records in their past state, but also to enumerate the records that used to exist back then, but are gone by now.

Worse: In the same order these records would have been coming had the DB been in its past state.

The more I think about it, the more I realize token-based and timestamp-based pagination should be Current's user territory (read: RS). The built-in REST interface for the storage should not do anything sophisticated; it's closer to a simple Redis client than to PhpMyAdmin.

It should be extensible though, for the user to not have to copy-paste and/or re-implement existing pieces from scratch.

nyospe: Understood. In the RS model (where everything is journalled) it is a
reasonable (if not always optimal) approach, but without that assumption,
it becomes much, much uglier.

dkorolev: In Current, everything is journaled.

But if it's an `std::unordered_map<>`, no matter how efficiently can the log be replayed backwards, the order of iteration may change. Too ugly to bother, if you ask me.

Storage/REST-API.md
mzhurovich: Thnx, maybe worth considering at some point.
Not sure though, if we are ready to start with something more sophisticated than structure fields tagging, which @dkorolev mentioned below.

Storage/REST-API.md
mzhurovich: :+1: 
Added an error code in 83179f486ab1800dbffca6fdda287652d566f3ee

Storage/REST-API.md
mzhurovich: :+1: 83179f486ab1800dbffca6fdda287652d566f3ee

Storage/REST-API.md
dkorolev: Yes.

image

isagalaev:
Hi everyone, @dkorolev asked me to review this. Prepare for a little spamming, take at face value :-)

dkorolev:
@isagalaev Burn!

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
+# RESTful binding for Storage, RFC
+
+This document describes the RESTful API with CRUD functionality for Current Storage.
+
+### Bindings semantics
+`TODO: Describe storage field <-> endpoint bindings.`
+
+### Response schema
+
+The response body is always either a valid JSON or empty with `204 (No content)` error code.
```
isagalaev: I'd say there's no real value in 204, nobody ever is going to treat it differently from 200 with Content-Length: 0. A good client would probably shun it under a blanket [200;300) condition, so this just takes space in the spec.

isagalaev: Also, depending on how fancy you want the style of this spec to be, it may be worth explicitly mentioning the content-type of responses. But not very useful again, as the API seems to assume only custom clients which will expect only one type of representation anyway.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+  
+  [required]
+  "data": {
+    ...
+  }
+
+}
+```
+
+The responses are strongly typed. The contents of the request URL define the schema for the response.
```
isagalaev: I keep re-reading this and seems to me that the words "strongly typed" should be accompanied with a formal way to enforce it. It's not a written rule, I'm just expecting it here, subjectively. Since the schema (as far as I see) is in the prose of the spec, it may be better to just say that all responses adhere to the format(s) defined in this spec.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+```
+{
+  “error”: “Detailed human-readable error message.”,
+  ... other fields ...
+}
+```
+
+Successful responses are JSON objects without the top-level `"error"` field.
+
+`TODO: We also may have redirects, which contain the "Location" in the header, with empty body.`
```
isagalaev: Technically, 3xx responses and successes. There's no need to mentioned them at all, as this is implied by REST as a part of the unified HTTP interface. In practical terms, I've never seen a library that wouldn't handle redirects by default (my experience is not exhaustive, of course.)

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Notation
+
+The default way to encode JSON objects is JavaScript-friendly. The API also supports F#-friendly format.
```
isagalaev: Well, *that* does deserve different content types. Requires, in fact. Because you don't want clients to do content sniffing. Something like `application/json;flavor={json|fsharp}` might do.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+2. Coding of absent optional objects.
+
+   The JavaScript-friendly implementation omits missing fields from the output (and does not expect them in the input).
+The F#-friendly implementation outputs missing fields as `null`-s (and expects `null`-s in the input).
+
+3. Coding of dictionaries with the keys being strings.
+
+   Generally, both JavaScript- and F#-friendly implementations encode dictionaries as arrays of arrays of two elements: `[[key1,value1],[key2,value2],...]`.
+   
+   However, if the type of `key` is string, the JavaScript-friendly format of this dictionary becomes an object: `{"key1":value1,"key2":value2,...}`.
```
isagalaev: This is a potential problem for parsing. Better to have a content type to dictate the exact format.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+Two implicitly defined views (both default to "all fields") are `full` and `preview`.
+
+The `full` view is the default, so that `&format=full` is equivalent to not specifying the `&format` URL parameter.
+
+The `&format=preview` is used when the response schema for certain endpoint contains not one record, but a collection of them. In other words, the representation of each record of `"data":[...]` for the endpoints where `"data"` is an array is equivalent to the representation of those records with `&format=preview`.
+
+The views are defined by the user, and the default representation of any view is "all fields". In particular, unless the user defines the `"preview"` view, it will be identical to the `"full"` one.
+
+### Methods
```
isagalaev: Here goes the pet peeve :-)

Don't clutter the spec with the description of HTTP methods that are implied by REST. It's not an educational material. 

A REST spec should be structured around Resources. For each Resource there should be:

- An URL (or an URL template) with resource-specific parameters.
- Notes on the special meaning of methods, *if any*. This is usually needed for POST as it lacks a defined semantics.
- In a general case, the representation format(s). Though it's okay to have a common format described elsewhere.

Storage/REST-API.md
dkorolev: I know it goes against REST, but what would be wrong with `/api/js/...` and `/api/fs/...` ? Schema-wise, the corresponding JSON formats can be seen as substantially different (not to mention we may have more than JS and F#), and I'd like to leave no room for possible confusion from the user's end.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+Current Storage API interprets HTTP verbs in the following way:
+
+* **GET**: Get the resource.
+
+  Strongly typed with respect to returning an individual record or a collection of records. For example, `/users` can be the way to GET the list of users (`"data":[...]`), and `/user/1` can be the way to GET an individual user with the ID of `1`.
+   
+* **PATCH**: Update some fields of the resource.
+
+  `PATCH` should be called on the URL that has been (or could have been) returned as the top-level `"url":"..."` field of the payload of `GET`.
```
isagalaev: That (potentially) contradicts REST. Resources should be addressable only by URLs, having to request PATCH on a different URL is completely unexpected. It's okay for a resource to have a canonical URL mentioned in the representation but if it's accessible by another one, that another one should work for all methods as well.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+  Generally, the user is expected to retrieve the value of the object under the key `"data"` of the top-level payload of a `GET` request, alter this object, and send its modified version back to the server.
+
+  `PATCH` respects the format of the `"data"` controlled by the `&format=` URL parameter. Sending a `PATCH` request to the URL with certain `&format=X` can only modify the fields returned by the `GET` request of this `&format=X`.
```
isagalaev: Again, this should be implied. A format is defined by content type, it should not depend on the method being used. Often PATCH *does* want a special format, but that should be conveyed to the server with the `Content-type` header.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+  For `PATCH` requests, the key of the record being updated is taken from the URL. The user is expected to either not send in the key as part of the payload, or ensure the key contained in the payload matches the key contained in the URL.
```
isagalaev: A common practice is to simply ignore any ids sent as part of representation in PATCH and PUT as the resource is identified by the URL.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+  
+  `PATCH` will fail if attempted to modify a non-existing resource, resulting in `404 (Not Found)` error code.
+  
+  Also, `PATCH` performs conflict resolution. The URL being `PATCH`-ed must contain a revision / timestamp of the resource. The exact semantics of the revision / timestamp is not specified by this RFC, but the top-level `"url"` field returned by the `GET` request is guaranteed to have it set.
```
isagalaev: Any reason you don't want to use the standard ETag/If-Match pair? Since you have a revision number anyway it would serve as a natural ETag.

Storage/REST-API.md
dkorolev: Good point. Allowing to `PUT/PATCH/DELETE` only the resource containing the most recent `&timestamp` in the URL is the most elegant way we found to ensure no "merge conflicts" when multiple clients are accessing the same resource.

Suggestions for a better solution would be greatly appreciated. Thank you!

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+  An attempt to `PATCH` to a URL that points to the version of the resource other than the most recent one will result in no data mutation and a `409 Conflict` error code. The user is then expected to re-run the `GET` request, followed by the re-run `PATCH` request onto the URL of the most recent version of the resource returned by the `GET` request. Confirming that the modifications that occurred in the meantime do not conflict with the new payload being sent in the body of the `PATCH` request is then the responsibility of the API user.
```
isagalaev: That's useful. I think it deserves its own chapter "Conflict resolution" instead of being buried in the HTTP primer.

Storage/REST-API.md
dkorolev: No reason except for the URL is always there, and asking the user to send the mutating request onto the URL returned in the `.url` field of the GET-ted JSON looks straightforward.

The solution involving headers looks heavier to my taste (for instance, I'd have to look it up how to send a header from F# or Ruby).

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+* **PUT**: Create a new record for a resource, or update all fields of an existing resource.
+
+  Attempting to `PUT` to a URL that has an incomplete view of a record would result in a `405 Method Not Allowed` error.
```
isagalaev: If I understand it right, 405 is a bad choice. It says that this whole resource doesn't allow PUTs at all. I'd use a custom 400 — the default choice for mal-formed representations.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+  
+  In the `&format=full` view, `PUT` is equivalent to `PATCH`.
```
isagalaev: That contradicts REST too. PUT has a well-defined semantics of sending the whole representation to be placed at the specified URL. Also, there's simply no need to have two different ways to do PATCHing.

Storage/REST-API.md
isagalaev: @dkorolev the important difference also is that PUT and POST are intended to different resources: a creating POST is done to a *collection*, it doesn't make sense for a single non-existing item (resulting in 404).

Storage/REST-API.md
dkorolev: Looks like a documentation issue, not the architecture issue.

Our PUT follows the convention of RESTful PUT; it just so happens that the shortest way to explain it is to declare PUT an equivalent to PATCH for a full-fields view of the resource.

If it makes sense, how would you phrase it to look smoothly to a knowledgeable reader?

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+* **POST**: Create a new record and assign it to some resourse.
+
+  `POST` will create a new resource. The URL for POST must not contain the key(s) for the record being posted, and the payload body should not contain the key(s) either.
+
+  The generation of the new key is user-defined. Three common strategies are: 1) random, 2) the [salted] hash of the payload, 3) the value of certain passed in field or fields.
+  
+  In the latter two, as well as other, cases, a `POST` request may result in an attempt to overwrite an already existing resource. By default, `POST` will not overwrite, but return a `409 Conflict` error code, with the top-level `"url"` field of the returned JSON object containing the URL for the resource that was about to be overwritten.
+
+  The `&overwrite=1` / `&overwrite=true` URL parameter can be set to indicate that the `POST` request should complete successfully even if it would result in overwriting existing data.
```
isagalaev: Since you just returned an URL of the resource to be overwritten, the user can simply use PUT, no need for a special key. But this strategy should be explicitly described.

Storage/REST-API.md
```diff
@@ -0,0 +1,185 @@
...
+
+Successful `POST` results in either `200 (OK)` error code if the resource has been updated, or `201 (Created)`.
+
+* **DELETE**: Deletes the resource.
+
+   Similarly to `PATCH` and `PUT`, `DELETE` should be called on the URL that has been (or could have been) returned as the top-level `"url":"..."` field of the payload of `GET`.
```
isagalaev: The spec doesn't mention it, but per REST deletion is idempotent: a user should able to safely "delete" already deleted resource, resulting in 200 (or 204 if you want to be fancy), not 4xx.

Storage/REST-API.md
dkorolev: Of course -- terminology issue from our end, my bad.

Say, for `/user` resources, our (internal) way of thinking of `PUT` and `POST` implies assuming `/user` and `/user/123` are actually the same endpoint -- we effectively treat `/123` as a URL argument for `/user`, as `/user` is registered as "a route accepting zero or one slash-separated URL arguments after itself".

Agree it should be phrased cleaner. Thanks!

Storage/REST-API.md
dkorolev: There's a corner case: doing a `POST` without specifying the resource, of an object that contains enough data to imply what the resource should be, and that resource already existing.

For example, consider the DB where the server-side logic knows the primary key for users are e-mails. The user does not have to know it, and they do a POST. If the body of this POST request contains an object with an e-mail that already presents in the DB, a point can be made to have the default behavior not overwrite the existing record.

A PUT immediately following a POST that returned an error and a "canonical" URL for the resource to be updated is indeed a solution. It's extra logic on the client side though, hence we thought an `&overwrite=true` flag could be useful.

Storage/REST-API.md
isagalaev: Nothing wrong, and it's not "against" REST (unless you're a purist). Specifying a format in the URL is a common practice. What I meant was the Content-type header of the *response* that should say which format it's actually in.

Storage/REST-API.md
isagalaev: I mentioned the ETag header further down. It's what it's for. Inventing a custom URL parameter just for that will surprise people.

Storage/REST-API.md
dkorolev: :+1:

Storage/REST-API.md
isagalaev: > asking the user to send the mutating request onto the URL returned in the .url field of the GET-ted JSON looks straightforward

Straightforward may be but it is a custom deviation from the accepted practice.

> The solution involving headers looks heavier to my taste (for instance, I'd have to look it up how to send a header from F# or Ruby)

Not to argue, but this is completely contrary to my experience. Headers are a very expected tool when working with RESTful APIs and those APIs that try to mimic their functionality entirely through URLs always feel a bit unnatural.

dkorolev: Glad to hear it's me falling on the "outdated" end of the argument. Thank you!

Storage/REST-API.md
isagalaev: As I mentioned, I wouldn't have this whole chapter at all. Instead I'd have one dedicated to the Record resource that would list both PUT and PATCH as allowed methods and mention that PATCH accepts both the abbreviated and the full formats.
Also, borrowing from another my comment down below, a full-format PATCH should not, in fact, be equivalent to PUT as it's only latter of the two that should create non-existing resources.

Storage/REST-API.md
isagalaev: Yeah, that's an implementation detail that an API consumer will be glad to not know about :-) People are familiar with the idea that collections and items are fundamentally semantically different from the outside.

Storage/REST-API.md
isagalaev: You still have to modify your request — be it a method or an URL parameter — so there's nothing extra there. But PUT is already there, well-understood and implemented, while a custom parameter requires documentation and implementation.

isagalaev:
On a general note, the spec doesn't mention authorization at all. It usually should, unless the API is entirely privately accessible.

mzhurovich:
@sompylasar @nyospe @isagalaev thank you for the comments!

@dkorolev let's discuss the changes we'd like to add for the v.0.1 over the phone?

dkorolev:
@mzhurovich @isagalaev @sompylasar @nyospe : This has been great! Many of your comments have been already incorporated (thank you!), and we'll prioritize the rest to be driven by real life needs.

I'll merge&close the pull request.

PR #347, opened 2016-01-20
Serialization refactoring.

mzhurovich:
@dkorolev PTAL :)


dkorolev:
*WOW* !

It'll take me a while to process. Will try to find time today, ETA 24 hours to be safe.

Thanks,
Dima

PR #349, opened 2016-01-21
RESTful storage prototype.

mzhurovich:
Hi @dkorolev!
Here is the prototype of the RESTful layer for our storage.
It supports only `Dictionary` in a pretty simple manner, but workz ;)
What is needed to be discussed:
* Some transparent way to return more sophisticated results (Hypermedia etc.).
* Painless injection of custom user code into handlers (authentication etc.). Handlers chaining?

### DIFF
--- a/Storage/rest.h
+++ b/Storage/rest.h
@@ -1,0 +1,201 @@
+/*******************************************************************************
+The MIT License (MIT)
+
+Copyright (c) 2016 Maxim Zhurovich <zhurovich@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+#ifndef CURRENT_STORAGE_REST_H
+#define CURRENT_STORAGE_REST_H
+
+#include "storage.h"
+
+#include "../Blocks/HTTP/api.h"
+
+namespace current {
+namespace storage {
+namespace rest {
+namespace impl {
+
+template <int INDEX, typename T>
+struct RESTfulGETWrapper {
+  template <typename KEY>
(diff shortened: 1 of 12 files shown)


Storage/test.cc
```diff
@@ -48,7 +54,7 @@ namespace transactional_storage_test {
 CURRENT_STRUCT(Element) {
   CURRENT_FIELD(x, int32_t);
   CURRENT_CONSTRUCTOR(Element)(int32_t x = 0) : x(x) {}
-  // Dummy key function to support key-only deleter.
+  // To use `Element` in storage, it should either define a `.key` field, or `.key()` method.
```
dkorolev: This is actually a bug/feature: `Element` is only used in `Vector`, which doesn't need the `key/key()`. We'll take care of it soon-ish.

Storage/rest.h
```diff
@@ -0,0 +1,201 @@
...
+  static void Run(HTTPRoutesScope&, T_SERVER&, T_STORAGE&) {}
+};
+
+}  // namespace impl
+
+template <class T_STORAGE_IMPL>
+struct RESTfulStorage {
+  HTTPRoutesScope handlers_scope;
+  RESTfulStorage(T_STORAGE_IMPL& storage, int port) {
+    auto& http_server = HTTP(port);
```
dkorolev: :-)

Storage/rest.h
```diff
@@ -0,0 +1,201 @@
...
+  template <typename T_SERVER, typename T_STORAGE>
+  static void Run(HTTPRoutesScope& scope, T_SERVER& http_server, T_STORAGE& storage) {
+    scope += RegisterRESTfulStorageEndpoint<N>(http_server, storage);
+    RegisterRESTfulStorageEndpoints<current::variadic_indexes::indexes<NS...>>::Run(
+        scope, http_server, storage);
+  }
+};
+
+template <>
+struct RegisterRESTfulStorageEndpoints<current::variadic_indexes::indexes<>> {
```
dkorolev: Looks like a simple `template<int I, int N>` calling `<I+1, N>`, and having a corner case `<N, N>` would do here, no need in `variadic_indexes`.

Storage/test.cc
```diff
@@ -569,5 +613,147 @@ TEST(TransactionalStorage, OnDisk) {
...
+    EXPECT_EQ("MZ", ParseJSON<SimpleUser>(HTTP(GET(base_url + "/user/max")).body).name);
+
+    EXPECT_EQ(204, static_cast<int>(HTTP(DELETE(base_url + "/user/max")).code));
+    EXPECT_EQ(404, static_cast<int>(HTTP(GET(base_url + "/user/max")).code));
+  }
+}
+
+// TODO: `::current::storage::FieldNameAndTypeByIndex<0>()` can extract storage entry type ("Dictionary").
+
+// A "complete" "REST API" test.
```
dkorolev: Why quotes? It's indeed a REST API and a test for it! :-)

Storage/test.cc
```diff
@@ -569,5 +613,147 @@ TEST(TransactionalStorage, OnDisk) {
...
+  const auto persistence_file_remover = current::FileSystem::ScopedRmFile(persistence_file_name);
+
+  Storage storage(persistence_file_name);
+  EXPECT_EQ(2u, storage.FieldsCount());
+
+  const auto base_url = current::strings::Printf("http://localhost:%d", FLAGS_transactional_storage_test_port);
+
+  // Run twice to make sure the `GET-POST-GET-DELETE` cycle is complete.
+  for (size_t i = 0; i < 2; ++i) {
+    // Registed RESTful HTTP endpoints, in a scoped way.
```
dkorolev: s/Registed/Register/

TypeSystem/Serialization/types/variant.h
```diff
@@ -30,9 +30,7 @@ SOFTWARE.
 
 #include "primitive.h"
 
-// To fix `error: incomplete type ‘current::serialization::json::load::LoadFromJSONImpl<>`.
-#include "../json.h"
-#include "../binary.h"
+#include "current_typeid.h"
```
dkorolev: :+1:

dkorolev:
(LGTM++)++!

PR #350, opened 2016-01-23
`CURRENT_STRUCT_TAG` refactoring and split.

dkorolev:
Hi @mzhurovich ,

`CURRENT_STRUCT_TAG` refactoring, and separated `Dictionary` and `Vector` persistence-wise.

Thanks,
Dima

### DIFF
--- a/Storage/container/dictionary.h
+++ b/Storage/container/dictionary.h
@@ -83,7 +83,7 @@ class Dictionary {
     }
   }
 
-  void operator()(const T_UPDATE_EVENT& object) { map_[sfinae::GetKey(object)] = object; }
+  void operator()(const T_UPDATE_EVENT& object) { map_[sfinae::GetKey(object)] = object.data; }
   void operator()(const T_DELETE_EVENT& object) { map_.erase(sfinae::GetKey(object)); }
 
   struct Iterator final {
--- a/Storage/container/dictionary.h
+++ b/Storage/container/dictionary.h
@@ -40,13 +40,13 @@ namespace container {
 template <typename T,
           typename T_UPDATE_EVENT,
           typename T_DELETE_EVENT,
-          template <typename...> class MAP = Unordered>
-class Dictionary {
+          template <typename...> class MAP>
+class GenericDictionary {
  public:
   using T_KEY = sfinae::ENTRY_KEY_TYPE<T>;
   using T_MAP = MAP<T_KEY, T>;
 
-  explicit Dictionary(MutationJournal& journal) : journal_(journal) {}
+  explicit GenericDictionary(MutationJournal& journal) : journal_(journal) {}
 
   bool Empty() const { return map_.empty(); }
   size_t Size() const { return map_.size(); }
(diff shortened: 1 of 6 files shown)


Storage/container/dictionary.h
```diff
@@ -83,7 +83,7 @@ class Dictionary {
     }
   }
 
-  void operator()(const T_UPDATE_EVENT& object) { map_[sfinae::GetKey(object)] = object; }
+  void operator()(const T_UPDATE_EVENT& object) { map_[sfinae::GetKey(object)] = object.data; }
```
mzhurovich: `s/T_UPDATE_EVENT& object/T_UPDATE_EVENT& event/`?

Storage/storage.h
```diff
@@ -73,22 +73,47 @@ namespace storage {
...
+  CURRENT_STRUCT(alias##PoppedElement) {};                                        \
+  struct alias {                                                                  \
+    using T_ENTRY = base;                                                         \
+    using T_PUSHED_BACK_EVENT = alias##PushedElement;                             \
+    using T_POPPED_BACK_EVENT = alias##PoppedElement;                             \
+    using T_PERSISTED_EVENT_1 = alias##PushedElement;                             \
+    using T_PERSISTED_EVENT_2 = alias##PoppedElement;                             \
+  }
+
+// TODO(dkorolev) || TODO(mzhurovich): Move to `TypeSystem/` and test.
```
mzhurovich: Honestly speaking, I'm strongly against adding dependencies to Current components if we may not to.
If I got it right, you propose to move some Storage-related code to the `TypeSystem`. Why not just left it in the Storage?

Storage/storage.h
```diff
@@ -73,22 +73,47 @@ namespace storage {
...
+#define CURRENT_STRUCT_TAG(tag, base, alias) CURRENT_STRUCT_TAG_##tag(base, alias)
```
mzhurovich: And this semantics became much less obvious for the user :(
Looking just at the name of the macro `CURRENT_STRUCT_TAG` I suppose it to implement some way to "tag" any Current structure. So I think about possible `MyStruct`<->`SomeTag` connection.
Surprise! What this macro does has nothing in common with my expectations :)
* `tag` means storage field container type
* `alias` actually means `tag`
* this whole thing is supposed to be used for Storage. WTF?! :confounded:

mzhurovich:
BTW, what do you think of this:
https://github.com/c5t/Current/compare/master...mzhurovich:pr-350?expand=1

Storage/storage.h
dkorolev: I'm only considering adding one line to `TypeSystem`: The `#define` mapping `CURRENT_STRUCT_TAG(tag, original_struct_name, tagged_struct_name)` to some `CURRENT_STRUCT_TAG_IMPL_##tag(original_struct_name, tagged_struct_name)`.

Storage/storage.h
dkorolev: Yes, tagging is a way to tag any `CURRENT_STRUCT`, with possibly a custom behavior.

How about we flip the parameters and allow two "signatures": `CURRENT_STRUCT_TAG(original_struct_name, tagged_struct_name)` and `CURRENT_STRUCT_TAG(original_struct_name, tagged_struct_name, tag_type)`?

As long as tagged names can be used later, and (for some tags, at least) there is a way to retrieve the "base" structure from its tagged type, we should be good to go.

Practical example: `CURRENT_STRUCT_TAG(User, HypermediaUserResponse, HypermediaGETResponse);`

Storage/container/dictionary.h
dkorolev: :+1:

dkorolev:
Like it. If we want to minimize the amount of changes generic enough to be part of the `TypeSystem`, `CURRENT_STORAGE_FIELD_ENTRY` and removing type from `CURRENT_STORAGE_FIELD` sounds like a way to go.

dkorolev:
And all the tests pass.

dkorolev:
@mzhurovich , factored your changes in, thanks & PTAL.

mzhurovich:
LGTM!



jn

PR #351, opened 2016-01-23
`static int PickPortForUnitTest()` [ffs] :-)

dkorolev:
Justification: One of the ports used in tests conflicted with one of the prod ports on our Hetzner machine. 

Why not fix it, figured I?


mzhurovich:
:-)

PR #353, opened 2016-01-24
REST v0.5

dkorolev:
Hi @mzhurovich 

Getting there. Done:

* Cleaned up template metaprogramming.
* Have dedicated template-selected per-method handlers now.
* The user logic class represents the state of the request itself, and the initial, pre-`Transaction` URL++ parsing logic is done by it too. Easy to inject custom code now, auth included.
* Separated REST itself and REST wrappers. As an example, have `Basic` and `Hypermedia` wrappers now.

Not done:
* "Proper" hypermedia (it formats errors as JSONs now, but returns objects plain straight still.)
* GET for the whole collection (I suggest we just return the array of strings on `/api/${field}` for now.)
* Support for anything other than `Dictionary`. (Should be easy as field type selector is already a template parameter of the implementation of RESTful wrappers.)
* Top-level URL listing possible fields accessible via the API.
* The top-level `make test` stopped passing, some `CURRENT_STRUCT` circular dependency in sight.

Thanks,
Dima

### DIFF
--- a/Storage/api.h
+++ b/Storage/api.h
@@ -2,6 +2,7 @@
 The MIT License (MIT)
 
 Copyright (c) 2016 Maxim Zhurovich <zhurovich@gmail.com>
+          (c) 2016 Dmitry "Dima" Korolev <dmitry.korolev@gmail.com>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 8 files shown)


mzhurovich:
@dkorolev Thanks, great job!
Merging.

PR #370, opened 2016-02-10
per-field exclusion from RESTfulStorage

nyospe:


Storage/api.h
```diff
@@ -219,6 +219,13 @@ STORAGE_HANDLERS_MAP_ENTRY GenerateRESTfulHandler(STORAGE& storage, const std::s
 }
 
 }  // namespace impl
+template <bool CALL>
```
nyospe: This really needs to be a general utility. I usually create a collection of files in a directory std_ext, where I add things like this. It would then take a syntax like
if_true<CONDITIONAL>::call(FUNCTION_LIKE);
if_false<CONDITIONAL> becomes a type alias.

Note that I actually have the boolean logic reversed for this case, because I already had the ExcludeTypeFromPersistence defined. TODO: make it make sense.

Storage/storage.h
```diff
@@ -73,6 +73,14 @@ namespace storage {
 // 2) Splits the type into `T_PERSISTED_EVENT_1` and `T_PERSISTED_EVENT_2` to enable persisting deletions too.
 
 // clang-format off
+#define CURRENT_STORAGE_FIELD_EXCLUDE_FROM_PERSISTENCE(entry_name)                                          \
+  template<> struct current::storage::ExcludeTypeFromPersistence<entry_name::T_FIELD_TYPE<                  \
```
nyospe: Yay! for template specialization across namespaces getting less obnoxious in C++11

Storage/api.h
mzhurovich: It could be placed in `Bricks/template/`. Though I still don't get the idea of reversed logic :)

Current.vcxproj.filters
```diff
@@ -0,0 +1,1080 @@
...
+      <Filter>Header Files\Yoda\container\matrix</Filter>
+    </ClInclude>
+    <ClInclude Include="Yoda\container\matrix\exceptions.h">
+      <Filter>Header Files\Yoda\container\matrix</Filter>
+    </ClInclude>
+    <ClInclude Include="Yoda\container\matrix\metaprogramming.h">
+      <Filter>Header Files\Yoda\container\matrix</Filter>
+    </ClInclude>
+  </ItemGroup>
+</Project>
```
mzhurovich: Do these Visual Studio files build full test?

Storage/api.h
```diff
@@ -287,8 +294,11 @@ class RESTfulStorage {
                            const std::string& restful_url_prefix,
                            impl::STORAGE_HANDLERS_MAP& handlers) {
       ForEachFieldByIndex<BLAH, I - 1>::RegisterIt(storage, restful_url_prefix, handlers);
-      handlers.insert(
+      using T_SPECIFIC_FIELD = impl::RESTfulHandlerGenerator<T_REST_IMPL, I - 1, T_STORAGE_IMPL>::T_SPECIFIC_FIELD;
+      ConditionalOn<ExcludeTypeFromPersistence<T_SPECIFIC_FIELD>::excluded>::Call([&] {
```
mzhurovich: `ExcludeTypeFromPersistence` in fact excludes field from RESTful interface, not persistence, right?

Storage/base.h
```diff
@@ -85,6 +85,11 @@ struct StorageExtractedFieldType {
   using T_PARTICULAR_FIELD = T;
 };
 
+template<typename TYPE>
```
mzhurovich: `template <typename T>` for consistency :)

Current.vcxproj.filters
nyospe: No. I should probably add that and set up a build target. This just allows full intellisense and macro expansion in the IDE, which makes it a lot easier to work through VC++ compile errors.

Storage/api.h
nyospe: Right now, "ConditionalOn<false>::Call(...) executes whatever is passed to it. It really should be ConditionalOn<true>::Call(...) that executes, and <false> that should elide the lambda.

Storage/api.h
nyospe: Yes. I'll rename it.

Storage/base.h
nyospe: :+1:

Storage/api.h
dkorolev: @nyospe Yes un-reversing the boolean logic would make sense. How about `CallIf` or `ExecuteIf` as the name instead? And +1 to Max suggesting this belongs in `Bricks/template`.

Storage/api.h
```diff
@@ -219,6 +219,13 @@ STORAGE_HANDLERS_MAP_ENTRY GenerateRESTfulHandler(STORAGE& storage, const std::s
 }
 
 }  // namespace impl
+template <bool CALL>
+struct ConditionalOn {
+  template <typename Fn> static void Call(Fn&& f) { f(); }
```
dkorolev: `<typename F>` ?

Storage/base.h
dkorolev: :+1:

Storage/storage.h
```diff
@@ -73,6 +73,14 @@ namespace storage {
...
+                                                                 entry_name::T_ENTRY,                       \
+                                                                 entry_name::T_PERSISTED_EVENT_1,           \
+                                                                 entry_name::T_PERSISTED_EVENT_2 >> {       \
+    static constexpr bool excluded = true;                                                                  \
```
dkorolev: Looking at the with a fresh eye, `static constexpr bool persisted = false`, with it being true by default, reads cleaner.

Current.vcxproj.filters
nyospe: Looks like I'm going to need to autogenerate a project per test file. I'll do that another time.

Storage/storage.h
```diff
@@ -73,6 +73,26 @@ namespace storage {
...
+  struct FieldParticipatesIn##feature {         \
+    static constexpr bool participates = true;  \
+  }
+
+// defined here because Storage/api.h doesn't need to be included at the point where the EXCLUDE_FROM is used
+CURRENT_STORAGE_DEFINE_FEATURE(RESTfulAPI);
+
+#define CURRENT_STORAGE_FIELD_EXCLUDE_FROM_RESTFUL_API(entry_name)   \
+			CURRENT_STORAGE_FIELD_EXCLUDE_FROM(RESTfulAPI, entry_name)
+
```
dkorolev: Hi @nyospe ,

I've added tests and tweaked the logic slightly:
* https://github.com/dkorolev/Current/commit/0d1b1fd9bfcced55de5ec2037e8e3fa990a1b450
* https://github.com/dkorolev/Current/commit/7311df419315737b79159cf740f69c87c69bd313

It's going to master as @mzhurovich would be merging https://github.com/c5t/Current/pull/373.

Thanks!
Dima

PR #371, opened 2016-02-11
CURRENT_STRUCT_T

dkorolev:
@mzhurovich @nyospe -- here's a templated `CURRENT_STRUCT_T`.

### DIFF
--- a/TypeSystem/Serialization/test.cc
+++ b/TypeSystem/Serialization/test.cc
@@ -814,4 +814,80 @@ TEST(Serialization, TimeAsBinary) {
   }
 }
 
+namespace serialization_test {
+
+CURRENT_STRUCT_T(TemplatedValue) {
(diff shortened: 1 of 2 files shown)


TypeSystem/Serialization/test.cc
```diff
@@ -814,4 +814,80 @@ TEST(Serialization, TimeAsBinary) {
   }
 }
 
+namespace serialization_test {
+
+CURRENT_STRUCT_T(TemplatedValue) {
```
dkorolev: @nyospe hope this saves you some pain!

nyospe: Nice! Thank you.

mzhurovich:
@dkorolev wow, looks good!
We may have some problems with `TypeID`-s, but let's think about it later :)
Merging.

PR #376, opened 2016-02-15
Minor Sherlock and Storage improvements.

mzhurovich:


Sherlock/test.cc
```diff
@@ -347,6 +347,12 @@ TEST(Sherlock, SubscribeToStreamViaHTTP) {
   EXPECT_EQ(s[0] + s[1], HTTP(GET(base_url + "?cap=2")).body);
   EXPECT_EQ(s[0], HTTP(GET(base_url + "?cap=1")).body);
 
+  // Test `?cap_bytes=...`'
+  // Request slightly more max bytes than the size of the first entry.
+  EXPECT_EQ(s[0], HTTP(GET(base_url + "?cap_bytes=50")).body);
+  // Request exactly the size of the forst two entries.
```
dkorolev: s/forst/first/

Storage/persister/sherlock.h
```diff
@@ -100,7 +102,7 @@ class SherlockStreamPersisterImpl<TypeList<TS...>, PERSISTER, CLONER> {
     bool allow_terminate_ = false;
   };
 
-  sherlock::StreamImpl<T_TRANSACTION, PERSISTER, CLONER> stream_;
+  sherlock::Stream<T_TRANSACTION, PERSISTER, CLONER> stream_;
```
dkorolev: Does it compile this way? I believe it should be `StreamImpl` as it was.

PR #392, opened 2016-02-24
Several fixes in persisters and Sherlock.

mzhurovich:
@dkorolev This fixes two bugs:
1) `file.h` line 97
2) Sherlock didn't know last index and timestamp when started from non-empty persistence file.

### DIFF
--- a/Blocks/Persistence/file.h
+++ b/Blocks/Persistence/file.h
@@ -94,7 +94,7 @@ template <typename ENTRY>
 class FilePersister {
  private:
   struct Impl {
-    const std::string& filename;
+    const std::string filename;
     idxts_t next_initializer;
     std::atomic<uint64_t> next_index;
     std::atomic<std::chrono::microseconds> next_timestamp;
@@ -211,6 +211,14 @@ class FilePersister {
   }
 
   uint64_t Size() const noexcept { return impl_->next_index; }
+
+  idxts_t LastIndexAndTimestamp() const noexcept {
+    if (impl_->next_index) {
+      return idxts_t(impl_->next_index - 1, impl_->next_timestamp.load() - std::chrono::microseconds(1));
+    } else {
+      return idxts_t();
+    }
+  }
 
   IterableRange Iterate(uint64_t begin_index, uint64_t end_index) const {
     if (end_index == static_cast<uint64_t>(-1)) {
--- a/Sherlock/test.cc
+++ b/Sherlock/test.cc
@@ -181,7 +181,7 @@ TEST(Sherlock, SubscribeAndProcessThreeEntries) {
   }
   ASSERT_FALSE(d.subscriber_alive_);
 
-  std::vector<std::string> expected_values{"[0:10,2:30] 1", "[1:20,2:30] 2", "[2:30,2:30] 3"};
+  const std::vector<std::string> expected_values{"[0:10,2:30] 1", "[1:20,2:30] 2", "[2:30,2:30] 3"};
   // A careful condition, since the subscriber may process some or all entries before going out of scope.
   EXPECT_TRUE(
       CompareValuesMixedWithTerminate(d.results_, expected_values, SherlockTestProcessor::kTerminateStr))
@@ -209,7 +209,7 @@ TEST(Sherlock, SubscribeSynchronously) {
     ;  // Spin lock.
   }
 
-  std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
+  const std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
   // A careful condition, since the subscriber may process some or all entries before going out of scope.
   EXPECT_TRUE(
       CompareValuesMixedWithTerminate(d.results_, expected_values, SherlockTestProcessor::kTerminateStr))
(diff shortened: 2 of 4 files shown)


Blocks/Persistence/file.h
```diff
@@ -94,7 +94,7 @@ template <typename ENTRY>
 class FilePersister {
  private:
   struct Impl {
-    const std::string& filename;
+    const std::string filename;
```
dkorolev: :sos: :floppy_disk: :heart:

Sherlock/test.cc
```diff
@@ -175,9 +209,10 @@ TEST(Sherlock, SubscribeSynchronously) {
     ;  // Spin lock.
   }
 
+  std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
   // A careful condition, since the subscriber may process some or all entries before going out of scope.
-  EXPECT_TRUE((d.results_ == "TERMINATE,4,5,6") || (d.results_ == "4,TERMINATE,5,6") ||
-              (d.results_ == "4,5,TERMINATE,6") || (d.results_ == "4,5,6,TERMINATE") || (d.results_ == "4,5,6"))
+  EXPECT_TRUE(
+      CompareValuesMixedWithTerminate(d.results_, expected_values, SherlockTestProcessor::kTerminateStr))
```
dkorolev: Looks less readable to me than it was :-(

Sherlock/test.cc
```diff
@@ -185,18 +220,22 @@ TEST(Sherlock, SubscribeAsynchronously) {
...
-  std::unique_ptr<SherlockTestProcessor> p(new SherlockTestProcessor(d, false));
+  std::unique_ptr<SherlockTestProcessor> p(new SherlockTestProcessor(d, false, true));
   p->SetMax(4u);
   bar_stream.Subscribe(std::move(p)).Detach();  // `.Detach()` results in the subscriber running on its own.
   while (d.seen_ < 3u) {
     ;  // Spin lock.
   }
   EXPECT_EQ(3u, d.seen_);
-  EXPECT_EQ("4,5,6", d.results_);  // No `TERMINATE` for an asyncronous subscriber.
+  std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
```
dkorolev: const

Sherlock/test.cc
dkorolev: Understand we're better off keeping it though.

dkorolev:
@mzhurovich LGTM, but why not retire `LastIndexAndTimestamp` / `last_idx_ts_` and use `next` / `end` semantics instead, as we're doing in `Persister`-s now?

Those `-1`-s look unnecessarily complex to me.

Sherlock/test.cc
mzhurovich: non-const could be moved into the function :)

mzhurovich:
@dkorolev Does `Iterate(0, static_cast<uint64_t>(-1)).end()` points to the last entry?

dkorolev:
@mzhurovich `Iterate(...)` has three forms now: `Iterate()`, `Iterate(begin_index)`, `Iterate(begin_index, end_index)`.

`Iterate(previous_end_index)` would begin from the entry right after the last one scanned with `Iterate(..., previous_end_index)`.

`Iterate(0, static_cast<uint64_t>(-1))` (as well as `Iterate(0)`, and `Iterate()`) would include the last entry available at the time of this call.

mzhurovich:
@dkorolev I don't get your idea :( How do you propose to change this:
`last_idx_ts_(std::make_shared<current::WaitableAtomic<idxts_t>>(storage_->LastIndexAndTimestamp()))`?

Sherlock/test.cc
dkorolev: The function accepts a copy by value. The following works w/ my `g++` and `clang++`:

```
 $ git diff
diff --git a/Sherlock/test.cc b/Sherlock/test.cc
index 176738b..7aef243 100644
--- a/Sherlock/test.cc
+++ b/Sherlock/test.cc
@@ -181,7 +181,7 @@ TEST(Sherlock, SubscribeAndProcessThreeEntries) {
   }
   ASSERT_FALSE(d.subscriber_alive_);
 
-  std::vector<std::string> expected_values{"[0:10,2:30] 1", "[1:20,2:30] 2", "[2:30,2:30] 3"};
+  const std::vector<std::string> expected_values{"[0:10,2:30] 1", "[1:20,2:30] 2", "[2:30,2:30] 3"};
   // A careful condition, since the subscriber may process some or all entries before going out of scope.
   EXPECT_TRUE(
       CompareValuesMixedWithTerminate(d.results_, expected_values, SherlockTestProcessor::kTerminateStr))
@@ -209,7 +209,7 @@ TEST(Sherlock, SubscribeSynchronously) {
     ;  // Spin lock.
   }
 
-  std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
+  const std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
   // A careful condition, since the subscriber may process some or all entries before going out of scope.
   EXPECT_TRUE(
       CompareValuesMixedWithTerminate(d.results_, expected_values, SherlockTestProcessor::kTerminateStr))
@@ -234,7 +234,7 @@ TEST(Sherlock, SubscribeAsynchronously) {
     ;  // Spin lock.
   }
   EXPECT_EQ(3u, d.seen_);
-  std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
+  const std::vector<std::string> expected_values{"[0:40,2:60] 4", "[1:50,2:60] 5", "[2:60,2:60] 6"};
   EXPECT_EQ(Join(expected_values, ','), d.results_);  // No `TERMINATE` for an asyncronous subscriber.
   EXPECT_TRUE(d.subscriber_alive_);
   bar_stream.Publish(42);  // Need the 4th entry for the async subscriber to terminate.
@@ -477,7 +477,7 @@ TEST(Sherlock, ParsesFromFile) {
   }
   ASSERT_FALSE(d.subscriber_alive_);
 
-  std::vector<std::string> expected_values{"[0:100,2:300] 1", "[1:200,2:300] 2", "[2:300,2:300] 3"};
+  const std::vector<std::string> expected_values{"[0:100,2:300] 1", "[1:200,2:300] 2", "[2:300,2:300] 3"};
   // A careful condition, since the subscriber may process some or all entries before going out of scope.
   EXPECT_TRUE(
       CompareValuesMixedWithTerminate(d.results_, expected_values, SherlockTestProcessor::kTerminateStr))
```

mzhurovich: Of course it works, it will make a copy.

dkorolev: But you don't use `std::move()` anywhere for those vars, or what am I missing? :)

PR #393, opened 2016-02-27
Fixed `idxts_t` in Sherlock.

dkorolev:


Blocks/Persistence/file.h
```diff
@@ -43,6 +43,8 @@ SOFTWARE.
 #include "../../Bricks/time/chrono.h"
 #include "../../Bricks/sync/scope_owned.h"
 
+#include "../../Bricks/util/atomic_that_works.h"  // TODO(dkorolev): Remove once GCC 5 is mainstream.
```
mzhurovich: I did some research and the main problem is not the GCC version. The thing is that `std::pair` is not "trivially copyable" (which is pretty weird platform-dependent concept). The best way I found to make it work both with Clang/OSX and GCC 4.8.2/Ubuntu is this one:
```
struct pod_idxts_t {
  uint64_t index;
  std::chrono::microseconds us;
  pod_idxts_t() = default;
  pod_idxts_t(uint64_t index, std::chrono::microseconds us) : index(index), us(us) {}
};
struct pod_idxts_pair_t {
  pod_idxts_t first;
  pod_idxts_t second;
};
std::atomic<pod_idxts_t> a1;
std::atomic<pod_idxts_pair_t> a2;
```
**NOTE:** GCC 4.8.2 requires `-mcx16` flag.

Blocks/Persistence/file.h
```diff
@@ -200,44 +211,47 @@ class FilePersister {
...
     }
+    last_published_and_end.second.second = timestamp;
+    const auto current = idxts_t(last_published_and_end.second.first, timestamp);
     impl_->appender << JSON(current) << '\t' << JSON(std::forward<E>(entry)) << std::endl;
-    ++impl_->next_index;
-    impl_->next_timestamp = current.us + std::chrono::microseconds(1);
+    last_published_and_end.first = last_published_and_end.second;
+    ++last_published_and_end.second.first;
+    last_published_and_end.second.second += std::chrono::microseconds(1);
+    impl_->last_published_and_end.store(last_published_and_end);
```
mzhurovich: Is the only benefit of storing `last` and `end` separately that all plus-minus-one logic is kept here?

Bricks/util/atomic_that_works.h
```diff
@@ -0,0 +1,66 @@
...
+  atomic_that_works(atomic_that_works&&) = delete;
+
+  T load() const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return value_;
+  }
+
+  T& load_into(T& destination) const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    std::memcpy(&destination, &value_, sizeof(T));
```
mzhurovich: Again, guaranteed only for `is_trivially_copyable` types, UB otherwise.

dkorolev: Good point. My implicit assumption may well violate the standard.

Blocks/Persistence/file.h
dkorolev: This doesn't quite nail it, unfortunately. See https://github.com/c5t/Current/pull/394.

Blocks/Persistence/file.h
dkorolev: @mzhurovich indeed, and you've convinced me it's unnecessary.

Let's fall back to the simple logic. Please see https://github.com/c5t/Current/pull/394. I guess I'd put a poor man's atomic back, at least for g++, so that and we could move on with our lives.

dkorolev:
I'd assume, while not being atomic, in conjunction with https://github.com/c5t/Current/pull/394, this request would do :)

PR #396, opened 2016-02-28
LCOV exclusions and more cleanup.

dkorolev:


Storage/transaction_result.h
```diff
@@ -107,6 +108,8 @@ class TransactionResult<void> : public OptionalResult<void> {
 
   TransactionResult(const OptionalResultExists& exists, bool commited)
       : OptionalResult<void>(exists), commited_(commited) {}
+
+  // TODO(dkorolev) + TODO(mzhurovich): Test this.
```
mzhurovich: Isn't it tested in `TEST(TransactionalStorage, Exceptions)` lines 334-340?

dkorolev: IIRC, LCOV showed these two lines in read. Lo-pri.

PR #398, opened 2016-03-02
{To/From}String is already in global scope!

dkorolev:


dkorolev:
Abandoning this pull request in favor of https://github.com/c5t/Current/pull/399.

PR #402, opened 2016-03-04
Minor Storage imrovements.

mzhurovich:


Storage/transaction_policy.h
```diff
@@ -146,19 +149,35 @@ struct Synchronous final {
       return future;                                                                         // LCOV_EXCL_LINE
     }
     bool successful = false;
+    T_RESULT f1_result;
     try {
-      f2(f1());
+      f1_result = f1();
```
dkorolev: What if `T_RESULT` does not have a default constructor? Declaring it within the `try` block looks safer to me.

Storage/transaction_policy.h
```diff
@@ -146,19 +149,35 @@ struct Synchronous final {
...
+      // Transaction was rolled back and returned nothing we can pass to `f2`.
+      journal_.Rollback();
+      promise.set_value(TransactionResult<void>::RolledBack(OptionalResultMissing()));
+    } catch (...) {  // The exception is captured with `std::current_exception()` below.
+      // LCOV_EXCL_START
+      journal_.Rollback();
+      try {
+        promise.set_exception(std::current_exception());
+      } catch (const std::exception& e) {
+        std::cerr << "Storage internal error in Synchronous::Transaction: " << e.what() << std::endl;
```
dkorolev: Some `Error during 'promise.set_exception()`` message would be more descriptive here.

Storage/transaction_policy.h
mzhurovich: Agree. However, in `Transaction(F&& f)` above we have to declare it outside, because we want to pass it as the result. So I decided to make it same here.

Storage/transaction_policy.h
mzhurovich: :+1: Fixed in e7feebcc47dda9c551a806bd0dc855180556e215

Storage/transaction_policy.h
```diff
@@ -119,14 +121,14 @@ struct Synchronous final {
       try {
```
sompylasar: Can `journal_.Rollback();` throw?

dkorolev: Not supposed to. If it does, the binary should die.

sompylasar: Die without at least trying to log some details?

dkorolev: Standard failure handler would report at least the type of exception.

sompylasar: ok.

PR #413, opened 2016-03-10
Lowercased header names.

dkorolev:
Hey @mzhurovich @sompylasar !

Here's one heck of perfect timing of the Universe.

Same day we talked about header names being case-insensitive, it hits us in practice: Orchard's response sends `content-length` without capitalization.

Questions:
* Shall we lowercase all headers like I just did?
* Or shall we only do case-insensitive comparison instead of `std::strcmp`?

Thanks,
Dima

### DIFF
--- a/Bricks/net/http/impl/server.h
+++ b/Bricks/net/http/impl/server.h
@@ -58,8 +58,8 @@ const char kCRLF[] = "\r\n";
 const size_t kCRLFLength = strings::CompileTimeStringLength(kCRLF);
 const char kHeaderKeyValueSeparator[] = ": ";
 const size_t kHeaderKeyValueSeparatorLength = strings::CompileTimeStringLength(kHeaderKeyValueSeparator);
-const char* const kLowercaseContentLengthHeaderKey = "content-length";
-const char* const kLowercaseTransferEncodingHeaderKey = "transfer-encoding";
+const char* const kLowercaseContentLengthHeaderKey = "content_length";
+const char* const kLowercaseTransferEncodingHeaderKey = "transfer_encoding";
 const char* const kTransferEncodingChunkedValue = "chunked";  // Is this lowercase too? @sompylasar
 
 }  // namespace constants
@@ -256,6 +256,8 @@ class TemplatedHTTPRequestData : public HELPER {
               // Because `std::tolower` needs locale these days. -- D.K.
               if (c >= 'A' && c <= 'Z') {
                 c += 'a' - 'A';
+              } else if (c == '-') {
+                c = '_';
               }
             });
             HELPER::OnHeader(key, value);
--- a/Bricks/net/http/impl/server.h
+++ b/Bricks/net/http/impl/server.h
@@ -58,9 +58,9 @@ const char kCRLF[] = "\r\n";
 const size_t kCRLFLength = strings::CompileTimeStringLength(kCRLF);
 const char kHeaderKeyValueSeparator[] = ": ";
 const size_t kHeaderKeyValueSeparatorLength = strings::CompileTimeStringLength(kHeaderKeyValueSeparator);
-const char* const kContentLengthHeaderKey = "Content-Length";
-const char* const kTransferEncodingHeaderKey = "Transfer-Encoding";
-const char* const kTransferEncodingChunkedValue = "chunked";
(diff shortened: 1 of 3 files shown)


Bricks/net/http/impl/server.h
```diff
@@ -58,9 +58,9 @@ const char kCRLF[] = "\r\n";
 const size_t kCRLFLength = strings::CompileTimeStringLength(kCRLF);
 const char kHeaderKeyValueSeparator[] = ": ";
 const size_t kHeaderKeyValueSeparatorLength = strings::CompileTimeStringLength(kHeaderKeyValueSeparator);
-const char* const kContentLengthHeaderKey = "Content-Length";
-const char* const kTransferEncodingHeaderKey = "Transfer-Encoding";
-const char* const kTransferEncodingChunkedValue = "chunked";
+const char* const kLowercaseContentLengthHeaderKey = "content_length";
```
sompylasar: Why underscores?

Bricks/net/http/impl/server.h
```diff
@@ -58,9 +58,9 @@ const char kCRLF[] = "\r\n";
...
+const char* const kLowercaseTransferEncodingHeaderKey = "transfer_encoding";
+const char* const kTransferEncodingChunkedValue = "chunked";  // Is this lowercase too? @sompylasar
```
sompylasar: https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.6
> All transfer-coding values are case-insensitive.

Bricks/net/http/impl/server.h
```diff
@@ -252,10 +252,18 @@ class TemplatedHTTPRequestData : public HELPER {
             *p = '\0';
             const char* const key = &buffer_[current_line_offset];
             const char* const value = p + kHeaderKeyValueSeparatorLength;
+            std::for_each(&buffer_[current_line_offset], p, [](char& c) {
+              // Because `std::tolower` needs locale these days. -- D.K.
+              if (c >= 'A' && c <= 'Z') {
+                c += 'a' - 'A';
+              } else if (c == '-') {
+                c = '_';
```
sompylasar: Why prefer underscore, if dash is more standard?

sompylasar:
> Shall we lowercase all headers like I just did?
> Or shall we only do case-insensitive comparison instead of std::strcmp?

We should tolerate the readings, and unify the collection keys.

There are at least three approaches to unification:
* the one [in Node.js gives `x-forwarded-for`](https://nodejs.org/api/http.html#http_message_headers)
* the one [in PHP gives `HTTP_X_FORWARDED_FOR`](http://php.net/manual/ru/reserved.variables.server.php#87195).
* the one [in Nginx gives `$http_x_forwarded_for`](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_)

Bricks/net/http/impl/server.h
```diff
@@ -58,9 +58,9 @@ const char kCRLF[] = "\r\n";
 const size_t kCRLFLength = strings::CompileTimeStringLength(kCRLF);
 const char kHeaderKeyValueSeparator[] = ": ";
```
mzhurovich: Is the space mandatory? Looks wrong. RFC:
`header-field   = field-name ":" OWS field-value OWS`
Optional WhiteSpaces must be removed during parsing.

Bricks/net/http/impl/server.h
mzhurovich: +1 - why?

mzhurovich:
Speaking of pre-processing header fields. In NodeJS they have two fields - `headers` and `rawHeaders`. For `headers` they apply the following rules:
 * Duplicates of `age, authorization, content-length, content-type, etag, expires, from, host, if-modified-since, if-unmodified-since, last-modified, location, max-forwards, proxy-authorization, referer, retry-after`, or `user-agent` are discarded.
 * `set-cookie` is always an array. Duplicates are added to the array.
 * For all other headers, the values are joined together with `', '`.

sompylasar:
@mzhurovich :fire: :+1:

dkorolev:
@mzhurovich @sompylasar :fire: :+1:
Amending this one. Stay tuned.

PR #414, opened 2016-03-10
Current hash for `std::chrono::duration<>`.

mzhurovich:


dkorolev:
Probably best to cast it to microseconds first, so that hash from 42 millis != hash from 42 micros. Merging regardless to have it compile.

PR #416, opened 2016-03-11
Hypermedia REST redesign.

mzhurovich:


mzhurovich:
CC @sompylasar

Storage/rest/basic.h
```diff
@@ -174,7 +175,7 @@ struct Basic {
   };
 
   // LCOV_EXCL_START
-  static Response ErrorMethodNotAllowed() {
+  static Response ErrorMethodNotAllowed(const std::string&) {
     return Response("Method not allowed.\n", HTTPResponseCode.MethodNotAllowed);
```
dkorolev: Perhaps `Method "$METHOD" not allowed.` ?
And the '`\n'` is not required here any longer, right?

mzhurovich: 1b96a3db686cdbfc9dbdc284e66fe93a923433cf

TypeSystem/optional.h
```diff
@@ -96,6 +96,14 @@ class Optional final {
   // Another alternative is construct an `Optional<>` directly from `make_unique<>`.
 
   // TODO(dkorolev): Discuss the semantics with @mzhurovich.
+  // NOTE: I took the liberty to introduce these constructors as I see no harm. --M.Z.
```
dkorolev: This comment is safe to remove. LGTM!

mzhurovich: Done.

Storage/rest/advanced_hypermedia.h
```diff
@@ -34,15 +34,21 @@ SOFTWARE.
 #include "hypermedia.h"
 #include "sfinae.h"
 
-CURRENT_STRUCT(AdvancedHypermediaRESTRecordResponse) {
+namespace current {
+namespace storage {
+namespace rest {
+
+CURRENT_STRUCT_T(AdvancedHypermediaRESTRecordResponse) {
+  CURRENT_FIELD(success, Optional<bool>);
```
sompylasar: Why optional?

dkorolev: For it to not pollute every element of the returned array with `true`.

sompylasar: Oh, I see, this type is reused in the container as an item. Isn't the item type a different type than this response type?

Storage/rest/basic.h
```diff
@@ -123,7 +123,8 @@ struct Basic {
       input.entry.InitializeOwnKey();
       if (!Exists(input.field[current::storage::sfinae::GetKey(input.entry)])) {
         input.field.Add(input.entry);
-        return Response(current::ToString(current::storage::sfinae::GetKey(input.entry)), HTTPResponseCode.Created);
+        return Response(current::ToString(current::storage::sfinae::GetKey(input.entry)),
```
sompylasar: Why the key is put into the "message" string, not into a dedicated "ResourceCreatedResponse" struct? Does not look strongly-typed to me.

Storage/rest/hypermedia.h
```diff
@@ -31,39 +31,146 @@ SOFTWARE.
...
+      : error(error), description(description) {}
+  CURRENT_CONSTRUCTOR(HypermediaRESTError)(
+      const std::string& error, const std::string& description, const T_DETAILS& details)
+      : error(error), description(description), details(details) {}
+};
+
+CURRENT_STRUCT(HypermediaRESTGenericResponse) {
+  CURRENT_FIELD(success, bool, true);
+  CURRENT_FIELD(message, Optional<std::string>);
+  CURRENT_FIELD(errors, Optional<std::vector<HypermediaRESTError>>);
```
sompylasar: What's the use case for an array of errors?

Storage/rest/hypermedia.h
```diff
@@ -31,39 +31,146 @@ SOFTWARE.
...
 CURRENT_STRUCT(HypermediaRESTStatus) {
   CURRENT_FIELD(up, bool);
+
   CURRENT_CONSTRUCTOR(HypermediaRESTStatus)(bool up = false) : up(up) {}
 };
 
-CURRENT_STRUCT(HypermediaRESTContainerResponse) {
+CURRENT_STRUCT(HypermediaRESTError) {
+  using T_DETAILS = std::map<std::string, std::string>;
+  CURRENT_FIELD(error, std::string);
```
sompylasar: This is commonly called "code".

Storage/rest/hypermedia.h
```diff
@@ -31,39 +31,146 @@ SOFTWARE.
...
+  CURRENT_FIELD(description, std::string);
```
sompylasar: This is commonly called "message" ("description" is good, too).

Storage/rest/basic.h
mzhurovich: This is `basic` REST, not `HyperMedia`.

Storage/rest/hypermedia.h
mzhurovich: Just left some room for future extension. Do you think it's better to leave one entry?

Storage/rest/hypermedia.h
mzhurovich: Even if it's a string like `ParseJSONError`?

Storage/rest/advanced_hypermedia.h
sompylasar: > For it to not pollute every element of the returned array with true.

Yes, for arrays, it's obvious. But for arrays, isn't an array item of a different type than a single response type?

Storage/rest/basic.h
sompylasar: Anyways, reading an identifier from a (human readable) message looks wrong to me.

Storage/rest/advanced_hypermedia.h
mzhurovich: I also was thinking about it. Do you know smth about the common practice of returning collections in Hypermedia world?

Storage/rest/hypermedia.h
sompylasar: I'd keep a room for extension inside the error object. I can't imagine how I, as an API client, would manage several errors received in response to a single request. If that request was a batch (several operations at once), then, for each item of the batch, I would expect an optional error, again one error per request.

Storage/rest/basic.h
mzhurovich: ?
Just returning `body`=`ToString(key)`.

Storage/rest/hypermedia.h
sompylasar: I've edited my comment:
> This is commonly called "name" or "code" (code is generally a number, but sometimes string).

"ParseJSONError" looks more like a "name", not a "code".
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name

Storage/rest/hypermedia.h
mzhurovich: There is also `details` field inside error, which is effectively `std::map<std::string, std::string>`.
@dkorolev what do you think about replacing array with a single object?

sompylasar: Yes, details are good (probably not as good as a structured object, but key-value would do well, too).

Storage/rest/basic.h
sompylasar: I'd expect
`{ success: true, message: "Created.\n", data: { id: "new-id" } }`
or
`{ success: true, message: "Created.\n", id: "new-id" }`
but not
`{ success: true, message: "new-id" }`.

sompylasar: "id" or "key", how you call it.

Storage/rest/advanced_hypermedia.h
dkorolev: @sompylasar This structure we're looking it as `AdvancedHypermediaRESTRecordResponse`. Perhaps putting `RecordOf` in front of `Hypermedia`, or something, could make it cleaner.

Storage/rest/basic.h
dkorolev: Or `{"success":true, "url":"https://..."}` ?

mzhurovich: Let's leave it as is for now. Nobody is going to use it in the nearest future :)

Storage/rest/advanced_hypermedia.h
sompylasar: In Hypermedia (REST), there is a resource, and a collection of resources. So, I'd expect:
* for a request querying a resource, a ResourceResponse:
```
// success
{ success: true, data: { url: "/resource1", data: { foo: "bar1" } } }

// error
{ success: false, error: { name: "ResourceNotFoundError", details: { id: "1" } } }
```
* for a request querying a collection of resources, a CollectionOfResourcesResponse: 
```
// success
{ success: true, data: [
  { url: "/resource1", data: { foo: "bar1" } },
  { url: "/resource2", data: { foo: "bar2" } }
] }

// error
{ success: false, error: { name: "CollectionNotFoundError" } }
```

Storage/rest/basic.h
sompylasar: > Or {"success":true, "url":"https://..."} ?

Probably. Also a separate short representation of the created resource would also be useful, though this could likely not be fully restful (just want to save a GET request).
`{"success":true, "url":"https://...", "data": {"foo": "bar"}}`

sompylasar: @mzhurovich didn't you want a code review? :) :no_mouth:

sompylasar: P.S.
> Nobody is going to use it in the nearest future :)
I remember one of you telling someone to get the identifier right from the response body, it wasn't even a JSON.

mzhurovich: @sompylasar Yeah, but I'd prefer to concentrate on the parts that are in the process of going live :)

Storage/rest/advanced_hypermedia.h
mzhurovich: @sompylasar That's almost what we have now, except for the fact that non-advanced Hypermedia returns just an array of URLs for collection. Is it a problem that there is `success: null` field in `data` array entries?

sompylasar: Strictly speaking, `null` is an explicit falsy value, so either explicit `true`, or no such `success` field at all in the scheme. But, of course, `success` for collections could be just ignored on the client.

Storage/rest/basic.h
sompylasar: Sure, no problem, LGTM and :ship: it.

Storage/rest/advanced_hypermedia.h
dkorolev: @sompylasar `null` won't be printed by the JSON serializer used for Hypermedia.
`Optional` is effectively the easiest way to hint the JSON builder to skip this field unless it's set.

Storage/rest/basic.h
dkorolev: @sompylasar : Yes, that was me. That auth token being returned as plain text response is certainly not good. We're no track to fix it.

:pray: :ship: :airplane: !

PR #418, opened 2016-03-12
Fixed setting cookies to an external server.

dkorolev:


Blocks/HTTP/impl/posix_client.h
```diff
@@ -81,8 +81,15 @@ class HTTPClientPOSIX final {
       for (const auto& h : request_headers_) {
         connection.BlockingWrite(h.header + ": " + h.value + "\r\n", true);
       }
-      for (const auto& h : request_headers_.cookies) {
-        connection.BlockingWrite("Set-Cookie: " + h.first + "=" + h.second + "\r\n", true);
+      if (!request_headers_.cookies.empty()) {
```
mzhurovich: Don't you want to make it a method `CookiesAsString()` in `Headers`?

mzhurovich: It would be useful for Mac client as well :)

mzhurovich: And even more useful when we add extra parameters to cookies

dkorolev: :+1: da6a41206b005fe8425c4d44fcfa11948605d2a2

dkorolev: And :+1: d643dea

PR #420, opened 2016-03-13
REST matrix support prototype.

dkorolev:
Hi @mzhurovich 

REST is functional for matrices now, for some degree of "functional". Two bits of mess are:

1. `ToString` and `FromString`. They now support `std::pair<>`-s. I tried to compile CTFO too, which would probably coincide with extending `{To/From}String` and/or migrating to the proper `Serialize` altogether. In person in 12+ hours.

2. I've moved `HasInitializeOwnKey` to be `ENABLE_IF`-guarded. Now, for types that do not expose this method, `POST` would return `"Method not allowed"`. Let's decide what's the best logic here.

3. Helpers for "REST" "behavior". Not sure if the name is best. On the bright side, it does the job.

PTAL && thanks!
Dima

### DIFF
--- a/Bricks/strings/util.h
+++ b/Bricks/strings/util.h
@@ -40,14 +40,14 @@ namespace strings {
 
 template <typename DECAYED_T, bool IS_ENUM>
 struct ToStringImpl {
-  static ENABLE_IF<std::is_pod<DECAYED_T>::value, std::string> ToString(DECAYED_T value) {
+  static ENABLE_IF<std::is_pod<DECAYED_T>::value, std::string> DoIt(DECAYED_T value) {
     return std::to_string(value);
   }
 };
 
 template <typename DECAYED_T>
 struct ToStringImpl<DECAYED_T, true> {
-  static ENABLE_IF<std::is_pod<DECAYED_T>::value, std::string> ToString(DECAYED_T value) {
+  static ENABLE_IF<std::is_pod<DECAYED_T>::value, std::string> DoIt(DECAYED_T value) {
     return std::to_string(static_cast<typename std::underlying_type<DECAYED_T>::type>(value));
   }
 };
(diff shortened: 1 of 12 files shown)


Storage/test.cc
```diff
@@ -1002,23 +1021,70 @@ TEST(TransactionalStorage, RESTfulAPITest) {
...
+              HTTP(GET(base_url + "/api/data/like/dima-beer")).body);
+    // Meh, this test is not for `AdvancedHypermedia`. -- D.K.
+    // EXPECT_EQ("{\"row\":\"max\",\"col\":\"beer\"}\n",
+    //           HTTP(GET(base_url + "/api/data/like/max-beer?fields=brief")).body);
+    EXPECT_EQ("{\"row\":\"max\",\"col\":\"beer\",\"details\":\"Cheers!\"}\n",
+              HTTP(GET(base_url + "/api/data/like/max-beer")).body);
+
+    // GET matrix as the collection.
+    EXPECT_EQ(200, static_cast<int>(HTTP(GET(base_url + "/api/data/like")).code));
+    EXPECT_EQ("max:beer\ndima:beer\n", HTTP(GET(base_url + "/api/data/like")).body);
```
sompylasar: Not JSON?

Bricks/strings/util.h
```diff
@@ -138,6 +147,23 @@ inline const T_OUTPUT& FromString(T_INPUT&& input, T_OUTPUT& output) {
                                                                                   output);
 }
 
+// Special case for `std::pair<>`. For Storage REST only for now.
+// TODO(dkorolev) + TODO(mzhurovich): Unify `FromString()` and `JSON()` under one `Serialize()` w/ policies?
+template <typename T_INPUT, typename T_OUTPUT_FST, typename T_OUTPUT_SND>
+struct FromStringEnumImpl<T_INPUT, std::pair<T_OUTPUT_FST, T_OUTPUT_SND>, false> {
+  static const std::pair<T_OUTPUT_FST, T_OUTPUT_SND>& Go(const std::string& input,
+                                                         std::pair<T_OUTPUT_FST, T_OUTPUT_SND>& output) {
+    const size_t dash = input.find('-');
```
mzhurovich: I don't get it: `ToStringImpl` combines with `:` and `FromStringEnumImpl` searches for `-` as a delimiter.

Storage/test.cc
dkorolev: Yeah, legacy of our separation between `Rest/{Basic,Hypermedia/AdvancedHypermedia}`.

In a nutshell, JSON is not a strict requirement for REST, and the unit test for `Basic`, non-JSON, REST, is cleaner to read.

We'll probably revisit this some time soon.

PR #421, opened 2016-03-13
Extension to have REST for CTFO build&run.

dkorolev:
Hi @mzhurovich --

Here's a quick hack on top of https://github.com/c5t/Current/pull/420 that makes CTFO build.

Thanks,
Dima


dkorolev:
@mzhurovich How about we merge this one too? Member `ToString()/FromString()` is the part that actually makes sense. :-)

PR #428, opened 2016-03-17
Minor tweaks on Windows.

dkorolev:


Bricks/net/tcp/impl/posix.h
```diff
@@ -296,7 +296,7 @@ class Connection : public SocketHandle {
   }
 
   inline Connection& BlockingWrite(const void* buffer, size_t write_length, bool more) {
-#ifdef CURRENT_APPLE
+#if defined(CURRENT_APPLE) || defined(CURRENT_WINDOWS)
     static_cast<void>(more);  // Supress the 'unused parameter' warning on Apple.
```
mzhurovich: `s/on Apple//`?

Storage/transaction_result.h
```diff
@@ -33,7 +33,12 @@ namespace storage {
 template <typename T>
 class TransactionResult : public OptionalResult<T> {
  public:
+#ifndef CURRENT_WINDOWS
   TransactionResult() = delete;
+#else
+  // MSVS needs this for `_Associated_state<>`, required by `std::promise<>`.
```
mzhurovich: :disappointed:

Bricks/net/tcp/impl/posix.h
dkorolev: :+1: 568dd8e

Storage/transaction_result.h
dkorolev: Yes, but given we build & run on non-Windows every single time, I don't think this "sacrifice" is in any way big.

mzhurovich:
@dkorolev :+1:

PR #429, opened 2016-03-17
Storage meta fields access + REST API refactoring.

mzhurovich:
* Factored out RESTful handlers input parameters
 * Eliminated `ALL_FIELDS` template parameter
 * Reviewed capture in cascaded lambdas
 * Storage meta fields are now added/deleted from within transaction via `{Set/Erase}TransactionMetaField(...)`

Will add storage access for `RegisterTopLevel` tomorrow.

### DIFF
--- a/Storage/api.h
+++ b/Storage/api.h
@@ -126,14 +126,12 @@ struct RESTfulHandlerGenerator {
         // Top-level capture by value to make own copy.
         [&storage, restful_url_prefix, field_name, data_url_component](Request request) {
           auto generic_input = RESTfulGenericInput<STORAGE>(storage, restful_url_prefix, data_url_component);
-          typename T_STORAGE::T_TRANSACTION_META_FIELDS transaction_meta_fields;
           if (request.method == "GET") {
             GETHandler handler;
             handler.Enter(
                 std::move(request),
                 // Capture by reference since this lambda is supposed to run synchronously.
-                [&handler, &generic_input, &field_name, &transaction_meta_fields](Request request,
-                                                                                  const std::string& url_key) {
+                [&handler, &generic_input, &field_name](Request request, const std::string& url_key) {
                   const T_SPECIFIC_FIELD& field =
                       generic_input.storage(::current::storage::ImmutableFieldByIndex<INDEX>());
 
@@ -146,16 +144,14 @@ struct RESTfulHandlerGenerator {
                                                   std::move(generic_input), fields, field, field_name, url_key);
                                               return handler.Run(input);
                                             },
-                                            std::move(request),
-                                            transaction_meta_fields).Detach();  // Meta fields are copied here;
-                },
-                transaction_meta_fields);
+                                            std::move(request)).Detach();
+                });
           } else if (request.method == "POST") {
             POSTHandler handler;
             handler.Enter(
                 std::move(request),
                 // Capture by reference since this lambda is supposed to run synchronously.
-                [&handler, &generic_input, &field_name, &transaction_meta_fields](Request request) {
+                [&handler, &generic_input, &field_name](Request request) {
                   try {
                     auto mutable_entry = ParseJSON<typename ENTRY_TYPE_WRAPPER::T_ENTRY>(request.body);
                     T_SPECIFIC_FIELD& field =
@@ -175,21 +171,17 @@ struct RESTfulHandlerGenerator {
                                                                 mutable_entry);
                                                 return handler.Run(input);
                                               },
-                                              std::move(request),
-                                              transaction_meta_fields)
-                        .Detach();  // Meta fields are copied here;
+                                              std::move(request)).Detach();
                   } catch (const TypeSystemParseJSONException& e) {
                     request(handler.ErrorBadJSON(e.What()));
                   }
-                },
-                transaction_meta_fields);
+                });
           } else if (request.method == "PUT") {
             PUTHandler handler;
             handler.Enter(
                 std::move(request),
                 // Capture by reference since this lambda is supposed to run synchronously.
-                [&handler, &generic_input, &field_name, &transaction_meta_fields](
-                    Request request, const std::string& key_as_string) {
+                [&handler, &generic_input, &field_name](Request request, const std::string& key_as_string) {
                   try {
                     const auto url_key = current::FromString<typename ENTRY_TYPE_WRAPPER::T_KEY>(key_as_string);
                     const auto entry = ParseJSON<typename ENTRY_TYPE_WRAPPER::T_ENTRY>(request.body);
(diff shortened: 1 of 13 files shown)


Storage/api.h
```diff
@@ -141,23 +163,21 @@ struct RESTfulHandlerGenerator {
...
                     auto mutable_entry = ParseJSON<typename ENTRY_TYPE_WRAPPER::T_ENTRY>(request.body);
                     T_SPECIFIC_FIELD& field = storage(::current::storage::MutableFieldByIndex<INDEX>());
                     storage.Transaction(
+                                // Capture mostly by value for safe async transactions.
                                 [handler,
                                  &storage,
                                  &field,
-                                 mutable_entry,
-                                 &restful_url_prefix,
+                                 &mutable_entry,
```
dkorolev: This is the temporary one returned by `ParseJSON<>`, right?

Storage/api.h
```diff
@@ -160,14 +165,21 @@ struct RESTfulHandlerGenerator {
                                                  data_url_component};
                                           return handler.Run(args);
                                         },
-                                        std::move(request)).Detach();
-                          });
+                                        std::move(request),
+                                        transaction_meta_fields).Detach();  // Meta fields are copied here;
```
dkorolev: Semicolon at the end of the comment.

Storage/docu/docu_3_code.cc
```diff
@@ -42,6 +42,16 @@ SOFTWARE.
 
 #include "../../3rdparty/gtest/gtest.h"
 
+#ifndef _MSC_VER
```
dkorolev: `CURRENT_WINDOWS`?

Storage/docu/docu_3_code.cc
```diff
@@ -403,4 +413,82 @@ TEST(StorageDocumentation, RESTifiedStorageExample) {
 
 }
 
+namespace storage_docu {
+
+// Example of custom REST implementation that annotates transactions.
+struct RESTWithMeta: current::storage::rest::AdvancedHypermedia {
```
dkorolev: Space before the `:` ?

Storage/docu/docu_3_code.cc
```diff
@@ -403,4 +413,82 @@ TEST(StorageDocumentation, RESTifiedStorageExample) {
...
+  using SUPER = current::storage::rest::AdvancedHypermedia;
+
+  template <class HTTP_VERB, typename PARTICULAR_FIELD, typename ENTRY, typename KEY>
+  struct RESTful: SUPER::RESTful<HTTP_VERB, PARTICULAR_FIELD, ENTRY, KEY> {
```
dkorolev: And here.

Storage/docu/docu_3_code.cc
```diff
@@ -403,4 +413,82 @@ TEST(StorageDocumentation, RESTifiedStorageExample) {
...
+  current::time::SetNow(std::chrono::microseconds(2000));
+  EXPECT_EQ(200, static_cast<int>(HTTP(DELETE(base_url + "/api/data/client/101")).code));
+
+  // Check that everything has been persisted correctly, including meta fields.
+  const auto persisted_entries =
+      current::strings::Split<current::strings::ByLines>(current::FileSystem::ReadFileAsString(client_storage_file_name));
+  ASSERT_EQ(2u, persisted_entries.size());
+
+  const std::string& persisted_add = persisted_entries[0];
+  const size_t tab_pos = persisted_add.find('\t');
```
dkorolev: `const auto add_fields = strings::Split(persisted_add, '\t'); ASSERT_TRUE(add_fields.size() == 2u);` is probably cleaner.

Storage/transaction_policy.h
```diff
@@ -90,7 +90,8 @@ class Synchronous final {
       // LCOV_EXCL_STOP
     }
     if (successful) {
-      journal_.meta_fields = std::move(meta_fields);
+      journal_.meta_fields = transaction_meta_fields_;
```
dkorolev: `swap()` them instead?

dkorolev: (Or at least `std::move()` to keep the semantics clean?)

Storage/transaction_policy.h
```diff
@@ -127,7 +127,8 @@ class Synchronous final {
...
```
dkorolev: Same here.

Storage/transaction_policy.h
```diff
@@ -211,6 +211,7 @@ class Synchronous final {
 
   PERSISTER& persister_;
   MutationJournal& journal_;
+  TransactionMetaFields& transaction_meta_fields_;
```
dkorolev: If I understand it right and this is the reference owned by the object outside transaction policy, perhaps call it such? Ex. `transaction_meta_fields_reference_`?

mzhurovich: So `journal_reference_` as well?

dkorolev: Yes, please!

Storage/docu/docu_3_code.cc
mzhurovich: :+1: e3715ecd8608fcd0b3e9b056deba324c3f03d02a

Storage/docu/docu_3_code.cc
mzhurovich: :+1: e3715ecd8608fcd0b3e9b056deba324c3f03d02a

Storage/docu/docu_3_code.cc
mzhurovich: :+1: e3715ecd8608fcd0b3e9b056deba324c3f03d02a

Storage/docu/docu_3_code.cc
```diff
@@ -42,6 +42,16 @@ SOFTWARE.
 
 #include "../../3rdparty/gtest/gtest.h"
 
+#ifndef _MSC_VER
+DEFINE_string(client_storage_test_tmpdir,
+              ".current",
+              "Local path for the test to create temporary files in.");
+#else
+DEFINE_string(client_storage_test_tmpdir,
+              "Debug",
```
nyospe: Windows would actually be better off either using the same path as Linux/Mac or calling GetTempPath(), storing the result in a std::string, and appending "\\docu_3_code". "Debug" is a problematic path name in a typical VC++ build environment, because the actual build directories are platform first in 64 bit builds...

mzhurovich: Oh, thnx. We should probably fix this in a couple of other places.

dkorolev:
`make test` && `make check` passed. LGTM && merging!

Storage/api_base.h
```diff
@@ -0,0 +1,179 @@
...
+                      const std::string& restful_url_prefix,
+                      const std::string& data_url_component)
+      : storage(storage), restful_url_prefix(restful_url_prefix), data_url_component(data_url_component) {}
+  RESTfulGenericInput(const RESTfulGenericInput&) = default;
+  RESTfulGenericInput(RESTfulGenericInput&&) = default;
+};
+
+template <typename STORAGE, typename FIELD>
+struct RESTfulGETInput : RESTfulGenericInput<STORAGE> {
+  using T_ALL_FIELDS_BY_REFERENCE = MutableFields<STORAGE>;
```
nyospe: Shouldn't this actually be `ImmutableFields<STORAGE>`? Or are we anticipating logically but not physically mutable behavior for GETs in the future? Audit logging, maybe?

dkorolev: Yep, @mzhurovich convinced me extra journaling may happen as part of a `READ` transaction. We'll be addressing thus as soon as `Synchronous` would becoming not the very only `TransactionPolicy`.

PR #434, opened 2016-03-20
Moved `hash_combine` and hash for `std::pair` to Bricks.

mzhurovich:


mzhurovich:
@dkorolev Or maybe it's better to use `CurrentHashFunction` as the only hasher in `HashCombine`?

dkorolev:
@mzhurovich Too late, already merged.

PR #435, opened 2016-03-21
Local and remote IP and port for TCP and HTTP connections.

mzhurovich:
@dkorolev Could you try it on Windows pls?

### DIFF
--- a/Bricks/net/tcp/impl/posix.h
+++ b/Bricks/net/tcp/impl/posix.h
@@ -172,11 +172,35 @@ class SocketHandle : private SocketSystemInitializer {
 #endif
 };
 
+struct IPAndPort {
+  std::string ip;
+  int port;
+  IPAndPort(const std::string& ip = "", int port = 0) : ip(ip), port(port) {}
+  IPAndPort(const IPAndPort&) = default;
+  IPAndPort(IPAndPort&&) = default;
+};
+
+inline std::string InetAddrToString(const struct in_addr* in) {
+  const char* result = ::inet_ntoa(*in);
+  if (!result) {
+    CURRENT_THROW(InetAddrToStringException());
+  } else {
+    return std::string(result);
+  }
+}
+
 class Connection : public SocketHandle {
  public:
-  inline Connection(SocketHandle&& rhs) : SocketHandle(std::move(rhs)) {}
+  Connection(SocketHandle&& rhs, IPAndPort&& local_ip_and_port, IPAndPort&& remote_ip_and_port)
+      : SocketHandle(std::move(rhs)),
+        local_ip_and_port_(std::move(local_ip_and_port)),
+        remote_ip_and_port_(std::move(remote_ip_and_port)) {}
 
-  inline Connection(Connection&& rhs) : SocketHandle(std::move(rhs)) {}
+  Connection(Connection&& rhs) = default;
+
+  const IPAndPort& LocalIPAndPort() const { return local_ip_and_port_; }
+
+  const IPAndPort& RemoteIPAndPort() const { return remote_ip_and_port_; }
 
   // By default, BlockingRead() will return as soon as some data has been read,
   // with the exception being multibyte records (sizeof(T) > 1), where it will keep reading
@@ -343,6 +367,9 @@ class Connection : public SocketHandle {
   }
 
  private:
+  const IPAndPort local_ip_and_port_;
+  const IPAndPort remote_ip_and_port_;
+
   Connection() = delete;
   Connection(const Connection&) = delete;
   void operator=(const Connection&) = delete;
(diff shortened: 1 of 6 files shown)


Bricks/net/tcp/impl/posix.h
```diff
@@ -172,11 +172,35 @@ class SocketHandle : private SocketSystemInitializer {
 #endif
 };
 
+struct IPAndPort {
+  std::string ip;
+  int port;
+  IPAndPort(const std::string& ip = "", int port = 0) : ip(ip), port(port) {}
```
dkorolev: If the default constructor is disabled, why not `= delete;` it explicitly?

Bricks/net/tcp/impl/posix.h
```diff
@@ -440,7 +467,23 @@ class Socket final : public SocketHandle {
...
 #endif
     BRICKS_NET_LOG("S%05d accept() : OK, FD = %d.\n", static_cast<SOCKET>(socket), handle);
-    return Connection(SocketHandle::FromHandle(handle));
+
+    sockaddr_in addr_serv;
+    auto addr_serv_length = sizeof(sockaddr_in);
+#ifndef CURRENT_WINDOWS
+    if (::getsockname(handle,
+                      reinterpret_cast<struct sockaddr*>(&addr_serv),
+                      reinterpret_cast<socklen_t*>(&addr_serv_length)) == -1) {
```
dkorolev: Why not `!= 0` here, same as in Windows? The `man` page says `... On success, zero is returned.`.

Bricks/net/tcp/impl/posix.h
```diff
@@ -483,18 +526,43 @@ inline Connection ClientSocket(const std::string& host, T port_or_serv) {
...
       if (retval2) {
         CURRENT_THROW(SocketConnectException());  // LCOV_EXCL_LINE -- Not covered by the unit tests.
       }
+
+      sockaddr_in addr_client;
+      auto addr_client_length = sizeof(sockaddr_in);
+#ifndef CURRENT_WINDOWS
+      if (::getsockname(socket,
+                        reinterpret_cast<struct sockaddr*>(&addr_client),
+                        reinterpret_cast<socklen_t*>(&addr_client_length)) == -1) {
```
dkorolev: Nit here and above: If instead of `auto`, the variable cast into `socklen_t` is declared with the type specific to the platform, then the code, not data, part -- the `if (::getsockname(...) != 0)` --- would be identical on Windows and Linux.

I'll be able to test on Windows in about 4 hours.

dkorolev:
@mzhurovich LGTM, will report back in a few hours.

Bricks/net/tcp/impl/posix.h
mzhurovich: When both arguments are not specified, it's the default constructor, no?

Bricks/net/tcp/impl/posix.h
```diff
@@ -449,19 +449,19 @@ class Socket final : public SocketHandle {
...
     if (handle == static_cast<SOCKET>(-1)) {
       BRICKS_NET_LOG("S%05d accept() : Failed.\n", static_cast<SOCKET>(socket));
       CURRENT_THROW(SocketAcceptException());  // LCOV_EXCL_LINE -- Not covered by the unit tests.
     }
 #else
-    const SOCKET handle = ::accept(
-        socket, reinterpret_cast<struct sockaddr*>(&addr_client), reinterpret_cast<int*>(&addr_client_length));
+    int addr_client_length = sizeof(sockaddr_in);
+    const SOCKET handle =
+        ::accept(socket, reinterpret_cast<struct sockaddr*>(&addr_client), &addr_client_length);
```
dkorolev: This line is identical within and outside the `#ifdef CURRENT_WINDOWS`, right?

Bricks/net/tcp/impl/posix.h
dkorolev: Yeah, my bad. Perhaps `= default;` `operator=` for `const IPAndPort` and `IPAndPort&&` ?

Bricks/net/tcp/impl/posix.h
mzhurovich: Right. But the next one isn't :(

dkorolev: Yes, but why not DRY out the `const SOCKET handle = ...` one?

mzhurovich: And `ifdef` next line again? How about readability?

dkorolev: /me is shutting up then :+1:

mzhurovich:
@dkorolev How about this? 11fcb266fe1a27c23f8a35037d94e63ab6e2e5b5

mzhurovich:
Merging as is.
TBD: test it on Windows.

PR #436, opened 2016-03-21
Convenience exposure of template type in RESTfulGenericInput

nyospe:
Also fixing a lossfully converting assignment from size_t to int and back to size_t.

### DIFF
--- a/Bricks/util/base64.h
+++ b/Bricks/util/base64.h
@@ -74,7 +74,7 @@ struct Impl {
     if (nbits > 0) {
       result.push_back(map[((buf << 6) >> nbits) & 0x3F]);
     }
-    const int num_pads = result_size - result.size();
+    const size_t num_pads = result_size - result.size();
     result.append(num_pads, pad_char);
     return result;
   }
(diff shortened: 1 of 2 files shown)


Bricks/util/base64.h
```diff
@@ -74,7 +74,7 @@ struct Impl {
     if (nbits > 0) {
       result.push_back(map[((buf << 6) >> nbits) & 0x3F]);
     }
-    const int num_pads = result_size - result.size();
+    const size_t num_pads = result_size - result.size();
```
nyospe: This conversion was generating a compiler warning, not triggering a data error (and would never do so, because it is logically constrained to the range 0-2) so there's no unit test, but the existing tests pass with the change.

mzhurovich: :+1:

Storage/api_base.h
```diff
@@ -33,6 +33,7 @@ namespace rest {
 
 template <typename STORAGE>
 struct RESTfulGenericInput {
+  using STORAGE_TYPE = STORAGE;
```
mzhurovich: Already wrote a todo for myself to add it :)

nyospe: This is to alleviate the awkwardness of `ImmutableFields<std::remove_reference<decltype(in.storage)>::type>`

nyospe:
Thank you, much obliged.

PR #440, opened 2016-03-29
Sherlock type filter.

dkorolev:


Sherlock/test.cc
```diff
@@ -507,3 +512,81 @@ TEST(Sherlock, ParsesFromFile) {
...
+  }
+
+  {
+    using Collector = current::ss::StreamSubscriber<CollectorImpl, Record>;
+    static_assert(!current::ss::IsStreamSubscriber<Collector, Variant<Record, AnotherRecord>>::value, "");
+    static_assert(current::ss::IsStreamSubscriber<Collector, Record>::value, "");
+    static_assert(!current::ss::IsStreamSubscriber<Collector, AnotherRecord>::value, "");
+
+    Collector c(3);
+    stream.Subscribe<Record>(c).Join();
```
dkorolev: @mzhurovich Actually, the semantics can be even easier -- as `.Subscribe()` requires a `current::ss::StreamSubscriber`, the latter may just expose some `entry_t` of its own. Talk tomorrow / tonight :-)

Sherlock/filter.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "../port.h"
+
+#include "../TypeSystem/variant.h"
+#include "../Blocks/SS/ss.h"
+
+namespace current {
+namespace sherlock {
+
+// The "generic" case: Assume `SUBSCRIBER_TYPE` is a specific type, and `ENTRY` is a variant containing it.
```
mzhurovich: Maybe it's better to rename `ENTRY` to `VARIANT`/`STREAM_VARIANT`/`TYPES_VARIANT`...?

Sherlock/sherlock.h
```diff
@@ -333,26 +337,34 @@ class StreamImpl {
...
   // Asynchronous subscription: `subscriber` is a heap-allocated object, the ownership of which
   // can be `std::move()`-d into the subscriber thread. It can be `Join()`-ed or `Detach()`-ed.
-  template <typename F>
-  using AsyncSubscriberScope = GenericSubscriberScope<F, std::unique_ptr<F>, true>;
-
-  template <typename F>
-  AsyncSubscriberScope<F> Subscribe(std::unique_ptr<F>&& subscriber) {
-    static_assert(current::ss::IsStreamSubscriber<F, ENTRY>::value, "");
-    return std::move(AsyncSubscriberScope<F>(data_, std::move(subscriber)));
+  // The order of two template parameters is flipped to provide the default value to the `SUBSCRIBER_TYPE` one,
```
mzhurovich: Why not flip them everywhere?

Sherlock/sherlock.h
```diff
@@ -333,26 +337,34 @@ class StreamImpl {
...
-
-  template <typename F>
-  SyncSubscriberScope<F> Subscribe(F& subscriber) {
-    static_assert(current::ss::IsStreamSubscriber<F, ENTRY>::value, "");
-    return std::move(SyncSubscriberScope<F>(data_, std::unique_ptr<F, current::NullDeleter>(&subscriber)));
+  template <typename F, typename SUBSCRIBER_TYPE = SubscribeToAllTypes>
+  using SyncSubscriberScope =
+      GenericSubscriberScope<SUBSCRIBER_TYPE, F, std::unique_ptr<F, current::NullDeleter>, false>;
+
+  template <typename SUBSCRIBER_TYPE = SubscribeToAllTypes, typename F>
```
mzhurovich: And here they are not flipped :-/

Sherlock/filter.h
```diff
@@ -0,0 +1,74 @@
...
+template <typename SUBSCRIBER_TYPE, typename ENTRY>
+struct SubscriberFilter {
+  static_assert(TypeListContains<typename ENTRY::T_TYPELIST, SUBSCRIBER_TYPE>::value,
+                "Sherlock subscription filter by type requires the top-level type of the stream"
+                " to be a `Variant<>` containing the desired type as a sub-type.");
+
+  using entry_t = SUBSCRIBER_TYPE;
+
+  template <typename F, typename E>
+  static ss::EntryResponse ProcessEntry(F&& f, E&& entry, idxts_t current, idxts_t last) {
```
mzhurovich: Don't we always pass a cref to a subscriber?

Sherlock/sherlock.h
dkorolev: The reason template parameters are flipped is to enable the use of specialized template parameters using the default value.

For `*SubscriberScope`, if the user wants to fully specialie the type instead of using `auto`, the one-template-parameter version would cite user subscriber class as an explicit parameter, assuming the type of entries this listener is subscribed to is the type of entries stored by this Sherlock stream. Thus, it makes sense to keep the type of entries to subscribe to as the second parameter.

At the same time, while creating a `Subscriber`, the type of user class is something which would naturally infer itself, while type filter, if set, must be provided explicitly. Thus, it makes sense to keep `F` as the second template parameter.

I've cleaned this up a bit. PTAL.

Sherlock/sherlock.h
dkorolev: See the comment above.

Sherlock/filter.h
dkorolev: Not really, the logic in `Blocks/SS/pubsub.h` does offer some flexibility. Refactored quite a bit, PTAL at https://github.com/C5T/Current/commit/f02943e50494144410fb009c9c43727443a29b5c.

Sherlock/filter.h
dkorolev: :+1:, done this and a few more renamings in https://github.com/C5T/Current/commit/95dd9dc20a7035091d06214030fe8329228f5de5

dkorolev:
Hi @mzhurovich , please take another look.

The `Schema.SmokeTestFullStruct` test fails when run from a top-level `make test`, while passes `make test` run from `TypeSystem/` or `TypeSystem/Schema/` directories. I don't have the energy to investigate it now :disappointed_relieved: 

Thanks,
Dima

Blocks/SS/pubsub.h
```diff
@@ -121,21 +124,61 @@ class StreamSubscriber : public GenericStreamSubscriber<ENTRY>, public EntrySubs
...
+struct PassEntryToSubscriberOrSkipItImpl {
+  static_assert(TypeListContains<typename STREAM_UNDERLYING_VARIANT::T_TYPELIST, TYPE_SUBSCRIBED_TO>::value,
+                "Subscribing to the type different from the underlying type requires the underlying type"
+                " to be a `Variant<>` containing the subscribed to type as an option.");
+
+  template <typename F, typename E>
+  static EntryResponse Dispatch(F&& f, E&& entry, idxts_t current, idxts_t last) {
+    static_assert(IsEntrySubscriber<F, TYPE_SUBSCRIBED_TO>::value, "");
+    const E& entry_cref = entry;
+    if (Exists<TYPE_SUBSCRIBED_TO>(entry_cref)) {
```
mzhurovich: As I see in the test below, `Exists<>` should be able to accept rvalue, no?

TypeSystem/helpers.h
```diff
@@ -58,14 +58,33 @@ SOFTWARE.
 
 namespace current {
 
+template <typename TEST, typename T, bool B>
```
mzhurovich: `bool B` -> `bool IS_VARIANT`?

mzhurovich:
@dkorolev How do the changes you've made to `PowerfulValueImplCaller` actually affect the behavior of `Value()`? Maybe add some tests?

Blocks/SS/pubsub.h
dkorolev: It would accept it, yes, but semantically I'm worried about using `entry` an rvalue reference twice. In this case it would work, but to an unsavvy observer it might look like `x; f(std::move(x)); g(std::move(x));`, or even `x; y = std::move(x); z = std::move(x);`, which does not seem right.

TypeSystem/helpers.h
dkorolev: :+1: cc18d9a

dkorolev:
@mzhurovich To the best of my understanding, I've changed no existing behavior of to `Exists<>`/`Value<>`, but added two:

1. `Value<T>(T& x)` would now return `T& x`.
2. `Exists` and `Value` would both work for a `Variant<A, B, C> x; Value<Variant<A, B, C>>(x);`: when the passed in object is a `Variant`, and the requested type is the same `Variant`, not a type contained within this `Variant`.

dkorolev:
@mzhurovich Ah, and `Exists<T>(T)` is now true for primitive types and `CURRENT_STRUCT`-s.

(`Exists<Variant<A, B, C>(null Variant<A, B, C>())` is still false though.)

PR #441, opened 2016-03-30
Minor `TypeSystem/Schema` cleanup.

mzhurovich:

PR #442, opened 2016-03-30
Sherlock HTTP pubsub tweaks.

mzhurovich:
@dkorolev Incorporated your proposal on HTTP PubSub interface.


dkorolev:
Merging https://github.com/C5T/Current/pull/441 and https://github.com/C5T/Current/pull/442.

PR #448, opened 2016-04-02
Matrix-style storage container with one to one relationships

grixa:
CC @dkorolev and @mzhurovich

### DIFF
--- a/Storage/container/one2one.h
+++ b/Storage/container/one2one.h
@@ -73,35 +73,34 @@ class GenericOne2One {
       const auto it_col = transposed_.find(col);
       bool row_occupied = it_row != forward_.end();
       bool col_occupied = it_col != transposed_.end();
-      // TODO: need to simplify, this is too complex
-      if (!row_occupied && !col_occupied) {
-        journal_.LogMutation(T_UPDATE_EVENT(object),
-                             [this, row_col]() { DoErase(row_col); });
-      } else if (row_occupied && col_occupied) {
+      if (row_occupied && col_occupied) {
         const T prev_object1 = *(it_row->second);
         const T prev_object2 = *(it_col->second);
         const auto row_col1 = std::make_pair(row, sfinae::GetCol(prev_object1));
         const auto row_col2 = std::make_pair(sfinae::GetRow(prev_object2), col);
-        journal_.LogMutation(T_UPDATE_EVENT(object),
-                             [this, row_col, row_col1, prev_object1, row_col2, prev_object2] {
-                               DoErase(row_col);
+        journal_.LogMutation(T_DELETE_EVENT(prev_object1),
+                             [this, row_col1, prev_object1]() {
                                DoAdd(row_col1, prev_object1);
+                             });
+        journal_.LogMutation(T_DELETE_EVENT(prev_object2),
+                             [this, row_col2, prev_object2]() {
                                DoAdd(row_col2, prev_object2);
                              });
         DoErase(row_col1);
         DoErase(row_col2);
-      } else {
+      } else if (row_occupied || col_occupied) {
         const T prev_object = row_occupied ? *(it_row->second) : *(it_col->second);
         const auto prev_row_col = std::make_pair(sfinae::GetRow(prev_object),
                                                  sfinae::GetCol(prev_object));
-        journal_.LogMutation(T_UPDATE_EVENT(object),
-                             [this, row_col, prev_row_col, prev_object]() {
-                               DoErase(row_col);
+        journal_.LogMutation(T_DELETE_EVENT(prev_object),
+                             [this, prev_row_col, prev_object]() {
                                DoAdd(prev_row_col, prev_object);
                              });
         DoErase(prev_row_col);
       }
     }
+    journal_.LogMutation(T_UPDATE_EVENT(object),
+                         [this, row_col]() { DoErase(row_col); });
     DoAdd(row_col, object);
   }
   
@@ -109,7 +108,7 @@ class GenericOne2One {
     auto it = map_.find(row_col);
     if (it != map_.end()) {
       const T previous_object = *(it->second);
-      journal_.LogMutation(T_UPDATE_EVENT(previous_object),
+      journal_.LogMutation(T_DELETE_EVENT(previous_object),
                            [this, row_col, previous_object]() {
                              DoAdd(row_col, previous_object);
                            });
@@ -131,7 +130,7 @@ class GenericOne2One {
   void operator()(const T_UPDATE_EVENT& e) {
     const auto row = sfinae::GetRow(e.data);
     const auto col = sfinae::GetCol(e.data);
-    DoAdd(std::make_pair(row, col), e.data); // ??? is it enough? What about integrity?
+    DoAdd(std::make_pair(row, col), e.data);
   }
   void operator()(const T_DELETE_EVENT& e) { DoErase(std::make_pair(e.key.first, e.key.second)); }
   
--- a/Storage/container/one2one.h
+++ b/Storage/container/one2one.h
@@ -46,8 +46,8 @@ class GenericOne2One {
  public:
   using T_ROW = sfinae::ENTRY_ROW_TYPE<T>;
   using T_COL = sfinae::ENTRY_COL_TYPE<T>;
-  using T_RCPAIR = std::pair<T_ROW, T_COL>;
-  using T_WHOLE_MAP = std::unordered_map<T_RCPAIR, std::unique_ptr<T>, CurrentHashFunction<T_RCPAIR>>;
+  using T_KEY = std::pair<T_ROW, T_COL>;
+  using T_ELEMENTS_MAP = std::unordered_map<T_KEY, std::unique_ptr<T>, CurrentHashFunction<T_KEY>>;
   using T_FORWARD_MAP = ROW_MAP<T_ROW, const T*>;
   using T_TRANSPOSED_MAP = COL_MAP<T_COL, const T*>;
   using T_REST_BEHAVIOR = rest::behavior::Matrix;
@@ -57,50 +57,57 @@ class GenericOne2One {
   bool Empty() const { return map_.empty(); }
   size_t Size() const { return map_.size(); }
 
+  // Adds specified object, removes and/or overwrites all existing entries
+  // with row and/or col value matching the corresponding values of the new entry
   void Add(const T& object) {
     const auto row = sfinae::GetRow(object);
     const auto col = sfinae::GetCol(object);
-    const auto row_col = std::make_pair(row, col);
-    const auto it = map_.find(row_col);
+    const auto key = std::make_pair(row, col);
+    const auto it = map_.find(key);
     if (it != map_.end()) {
-      const T prev_object = *(it->second);
+      const T previous_object = *(it->second);
       journal_.LogMutation(T_UPDATE_EVENT(object),
-                           [this, row_col, prev_object]() { DoAdd(row_col, prev_object); });
+                           [this, key, previous_object]() { DoAdd(key, previous_object); });
     } else {
       const auto it_row = forward_.find(row);
       const auto it_col = transposed_.find(col);
-      bool row_occupied = it_row != forward_.end();
-      bool col_occupied = it_col != transposed_.end();
+      const bool row_occupied = (it_row != forward_.end());
+      const bool col_occupied = (it_col != transposed_.end());
       if (row_occupied && col_occupied) {
-        const T prev_object1 = *(it_row->second);
-        const T prev_object2 = *(it_col->second);
-        const auto row_col1 = std::make_pair(row, sfinae::GetCol(prev_object1));
-        const auto row_col2 = std::make_pair(sfinae::GetRow(prev_object2), col);
-        journal_.LogMutation(T_DELETE_EVENT(prev_object1),
-                             [this, row_col1, prev_object1]() { DoAdd(row_col1, prev_object1); });
-        journal_.LogMutation(T_DELETE_EVENT(prev_object2),
-                             [this, row_col2, prev_object2]() { DoAdd(row_col2, prev_object2); });
-        DoErase(row_col1);
-        DoErase(row_col2);
+        const T previous_object_same_row = *(it_row->second);
+        const T previous_object_same_col = *(it_col->second);
+        const auto key_same_row = std::make_pair(row, sfinae::GetCol(previous_object_same_row));
+        const auto key_same_col = std::make_pair(sfinae::GetRow(previous_object_same_col), col);
+        journal_.LogMutation(T_DELETE_EVENT(previous_object_same_row),
+                             [this, key_same_row, previous_object_same_row]() {
+                               DoAdd(key_same_row, previous_object_same_row);
+                             });
+        journal_.LogMutation(T_DELETE_EVENT(previous_object_same_col),
+                             [this, key_same_col, previous_object_same_col]() {
+                               DoAdd(key_same_col, previous_object_same_col);
+                             });
+        DoErase(key_same_row);
+        DoErase(key_same_col);
       } else if (row_occupied || col_occupied) {
-        const T prev_object = row_occupied ? *(it_row->second) : *(it_col->second);
-        const auto prev_row_col = std::make_pair(sfinae::GetRow(prev_object), sfinae::GetCol(prev_object));
-        journal_.LogMutation(T_DELETE_EVENT(prev_object),
-                             [this, prev_row_col, prev_object]() { DoAdd(prev_row_col, prev_object); });
-        DoErase(prev_row_col);
+        const T previous_object = row_occupied ? *(it_row->second) : *(it_col->second);
+        const auto previous_key =
+            std::make_pair(sfinae::GetRow(previous_object), sfinae::GetCol(previous_object));
+        journal_.LogMutation(T_DELETE_EVENT(previous_object),
+                             [this, previous_key, previous_object]() { DoAdd(previous_key, previous_object); });
+        DoErase(previous_key);
       }
     }
-    journal_.LogMutation(T_UPDATE_EVENT(object), [this, row_col]() { DoErase(row_col); });
-    DoAdd(row_col, object);
+    journal_.LogMutation(T_UPDATE_EVENT(object), [this, key]() { DoErase(key); });
+    DoAdd(key, object);
   }
 
-  void Erase(const T_RCPAIR& row_col) {
-    auto it = map_.find(row_col);
+  void Erase(const T_KEY& key) {
+    const auto it = map_.find(key);
     if (it != map_.end()) {
-      const T prev_object = *(it->second);
-      journal_.LogMutation(T_DELETE_EVENT(prev_object),
-                           [this, row_col, prev_object]() { DoAdd(row_col, prev_object); });
-      DoErase(row_col);
+      const T previous_object = *(it->second);
+      journal_.LogMutation(T_DELETE_EVENT(previous_object),
+                           [this, key, previous_object]() { DoAdd(key, previous_object); });
+      DoErase(key);
     }
   }
   void Erase(sfinae::CF<T_ROW> row, sfinae::CF<T_COL> col) { Erase(std::make_pair(row, col)); }
@@ -108,47 +115,55 @@ class GenericOne2One {
   void EraseRow(sfinae::CF<T_ROW> row) {
     const auto it = forward_.find(row);
     if (it != forward_.end()) {
-      const T prev_object = *(it->second);
-      const auto row_col = std::make_pair(row, sfinae::GetCol(prev_object));
-      journal_.LogMutation(T_DELETE_EVENT(prev_object),
-                           [this, row_col, prev_object]() { DoAdd(row_col, prev_object); });
-      DoErase(row_col);
+      const T previous_object = *(it->second);
+      const auto key = std::make_pair(row, sfinae::GetCol(previous_object));
+      journal_.LogMutation(T_DELETE_EVENT(previous_object),
+                           [this, key, previous_object]() { DoAdd(key, previous_object); });
+      DoErase(key);
     }
   }
 
   void EraseCol(sfinae::CF<T_COL> col) {
     const auto it = transposed_.find(col);
     if (it != transposed_.end()) {
-      const T prev_object = *(it->second);
-      const auto row_col = std::make_pair(sfinae::GetRow(prev_object), col);
-      journal_.LogMutation(T_DELETE_EVENT(prev_object),
-                           [this, row_col, prev_object]() { DoAdd(row_col, prev_object); });
-      DoErase(row_col);
+      const T previous_object = *(it->second);
+      const auto key = std::make_pair(sfinae::GetRow(previous_object), col);
+      journal_.LogMutation(T_DELETE_EVENT(previous_object),
+                           [this, key, previous_object]() { DoAdd(key, previous_object); });
+      DoErase(key);
     }
   }
 
-  ImmutableOptional<T> operator[](const T_RCPAIR& row_col) const {
-    const auto it = map_.find(row_col);
-    if (it == map_.end()) return nullptr;
-    return ImmutableOptional<T>(FromBarePointer(), it->second.get());
+  ImmutableOptional<T> operator[](const T_KEY& key) const {
+    const auto it = map_.find(key);
+    if (it != map_.end()) {
+      return ImmutableOptional<T>(FromBarePointer(), it->second.get());
+    } else {
+      return nullptr;
+    }
   }
   ImmutableOptional<T> Get(sfinae::CF<T_ROW> row, sfinae::CF<T_COL> col) const {
     return operator[](std::make_pair(row, col));
   }
   ImmutableOptional<T> GetRow(sfinae::CF<T_ROW> row) const {
     const auto it = forward_.find(row);
-    if (it == forward_.end()) return nullptr;
-    return ImmutableOptional<T>(FromBarePointer(), it->second);
+    if (it != forward_.end()) {
+      return ImmutableOptional<T>(FromBarePointer(), it->second);
+    } else {
+      return nullptr;
+    }
   }
   ImmutableOptional<T> GetCol(sfinae::CF<T_COL> col) const {
     const auto it = transposed_.find(col);
-    if (it == transposed_.end()) return nullptr;
-    return ImmutableOptional<T>(FromBarePointer(), it->second);
+    if (it != transposed_.end()) {
+      return ImmutableOptional<T>(FromBarePointer(), it->second);
+    } else {
+      return nullptr;
+    }
   }
 
-  bool CanAdd(const T_RCPAIR& row_col) const {
-    return map_.find(row_col) == map_.end() && forward_.find(row_col.first) == forward_.end() &&
-           transposed_.find(row_col.second) == transposed_.end();
+  bool CanAdd(const T_KEY& key) const {
+    return forward_.find(key.first) == forward_.end() && transposed_.find(key.second) == transposed_.end();
   }
   bool CanAdd(sfinae::CF<T_ROW> row, sfinae::CF<T_COL> col) const { return CanAdd(std::make_pair(row, col)); }
 
@@ -193,13 +208,13 @@ class GenericOne2One {
   const MapAccessor<T_TRANSPOSED_MAP> Cols() const { return MapAccessor<T_TRANSPOSED_MAP>(transposed_); }
 
   struct Iterator final {
-    using T_ITERATOR = typename T_WHOLE_MAP::const_iterator;
+    using T_ITERATOR = typename T_ELEMENTS_MAP::const_iterator;
     T_ITERATOR iterator_;
     explicit Iterator(T_ITERATOR iterator) : iterator_(iterator) {}
     void operator++() { ++iterator_; }
     bool operator==(const Iterator& rhs) const { return iterator_ == rhs.iterator_; }
     bool operator!=(const Iterator& rhs) const { return !operator==(rhs); }
-    const T_RCPAIR key() const { return iterator_->first; }
+    const T_KEY key() const { return iterator_->first; }
     const T& operator*() const { return *iterator_->second; }
     const T* operator->() const { return iterator_->second; }
   };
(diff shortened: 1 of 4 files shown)


Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+
+template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
+using UnorderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Unordered, Unordered>;
+
+template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
+using OrderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Ordered, Ordered>;
+
+}  // namespace container
+
+template <typename T, typename E1, typename E2>  // Entry, update event, delete event.
```
sompylasar: `E1` -> `T_UPDATE_EVENT`, `E2` -> `T_DELETE_EVENT` ?

grixa: I used existing containers as examples (dictionary, matrix), and they also declare StorageFieldTypeSelector using short template parameters' names. I think it’s because these particular structs in fact don’t care about parameters' meaning.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+namespace current {
+namespace storage {
+namespace container {
+
+template <typename T,
+          typename T_UPDATE_EVENT,
+          typename T_DELETE_EVENT,
+          template <typename...> class ROW_MAP,
+          template <typename...> class COL_MAP>
+class GenericOne2One {
```
mzhurovich: I'd prefer `OneToOne` instead of `One2One`. @dkorolev ?

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+ public:
+  using T_ROW = sfinae::ENTRY_ROW_TYPE<T>;
+  using T_COL = sfinae::ENTRY_COL_TYPE<T>;
+  using T_RCPAIR = std::pair<T_ROW, T_COL>;
```
mzhurovich: @dkorolev what do you think about declaring `pair<ROW, COL>` as `T_KEY` for matrices everywhere in the code? AFAIR I used it already in persistence types definition.

grixa: I agree, T_KEY looks much better.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+      const auto row_col = std::make_pair(sfinae::GetRow(prev_object), col);
+      journal_.LogMutation(T_DELETE_EVENT(prev_object),
+                           [this, row_col, prev_object]() { DoAdd(row_col, prev_object); });
+      DoErase(row_col);
+    }
+  }
+
+  ImmutableOptional<T> operator[](const T_RCPAIR& row_col) const {
+    const auto it = map_.find(row_col);
+    if (it == map_.end()) return nullptr;
```
mzhurovich: Here and below, pls, use the same style as in `matrix.h`:
```
if (...) {
  return x;
} else {
  return y;
}
```

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+    return ImmutableOptional<T>(FromBarePointer(), it->second);
+  }
+  ImmutableOptional<T> GetCol(sfinae::CF<T_COL> col) const {
+    const auto it = transposed_.find(col);
+    if (it == transposed_.end()) return nullptr;
+    return ImmutableOptional<T>(FromBarePointer(), it->second);
+  }
+
+  bool CanAdd(const T_RCPAIR& row_col) const {
+    return map_.find(row_col) == map_.end() && forward_.find(row_col.first) == forward_.end() &&
```
mzhurovich: If everything is consistent, the first comparison is redundant, right?

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+    const auto row = sfinae::GetRow(object);
+    const auto col = sfinae::GetCol(object);
+    const auto row_col = std::make_pair(row, col);
+    const auto it = map_.find(row_col);
+    if (it != map_.end()) {
+      const T prev_object = *(it->second);
+      journal_.LogMutation(T_UPDATE_EVENT(object),
+                           [this, row_col, prev_object]() { DoAdd(row_col, prev_object); });
+    } else {
+      const auto it_row = forward_.find(row);
```
mzhurovich: A few comments about the overwrite policy won't hurt.

Storage/container/one2one.h
grixa: You are right, it's redundant

Storage/test.cc
```diff
@@ -184,6 +186,40 @@ TEST(TransactionalStorage, SmokeTest) {
       }).Go();
       EXPECT_TRUE(WasCommitted(result));
     }
+    {
```
mzhurovich: Let's add comment here and in the blocks above in this test, what we are actually testing in each case.

Storage/test.cc
```diff
@@ -184,6 +186,40 @@ TEST(TransactionalStorage, SmokeTest) {
...
+        EXPECT_TRUE(fields.o.Empty());
+        EXPECT_EQ(0u, fields.o.Size());
+        EXPECT_TRUE(fields.o.Rows().Empty());
+        EXPECT_TRUE(fields.o.Cols().Empty());
+        fields.o.Add(Cell{1, "one", 1});
+        fields.o.Add(Cell{2, "two", 2});
+        fields.o.Add(Cell{2, "too", 3});
+        fields.o.Add(Cell{3, "too", 4});
+        EXPECT_FALSE(fields.o.Empty());
+        EXPECT_EQ(2u, fields.o.Size());
```
mzhurovich: And as `OneToOne` overwrites are a bit less obvious, some comments explaining what's going on here could increase readability :)

Storage/test.cc
```diff
@@ -213,7 +249,31 @@ TEST(TransactionalStorage, SmokeTest) {
       }).Go();
     }
 
-    // Rollback a transaction involving a `Matrix`.
+    // Iterate over the One2One
+    {
+      storage.Transaction([](MutableFields<Storage> fields) {
```
mzhurovich: `ImmutableFields` here and below?

Storage/test.cc
```diff
@@ -213,7 +249,31 @@ TEST(TransactionalStorage, SmokeTest) {
...
+        EXPECT_FALSE(fields.o.Empty());
+        std::multiset<std::string> data;
+        for (const auto& element : fields.o.Rows()) {
+          data.insert(current::ToString(current::storage::sfinae::GetRow(element)) + ',' +
+                      current::ToString(current::storage::sfinae::GetCol(element)) + '=' +
+                      current::ToString(element.phew));
+        }
+        EXPECT_EQ("1,one=1 3,too=4", current::strings::Join(data, ' '));
+      }).Go();
```
mzhurovich: Let's check the result of every transaction in the test. Because it could fail so that nobody will notice (as it were in `docu` test that you fixed) - `EXPECT_*` just won't be called. And yes, `docu` tests should be also fixed more thoroughly :)

Storage/test.cc
```diff
@@ -283,6 +354,47 @@ TEST(TransactionalStorage, SmokeTest) {
       }).Go()));
     }
 
+    // Iterate over the One2One with deleted elements, confirm the integrity of `forward_` and `transposed_`.
+    {
+      EXPECT_FALSE(WasCommitted(storage.Transaction([](MutableFields<Storage> fields) {
```
mzhurovich: Let's use here the same style as above (it looks more readable to me):
```
const auto result = ...
EXPECT_FALSE(WasCommitted(result));
```

Storage/test.cc
```diff
@@ -283,6 +354,47 @@ TEST(TransactionalStorage, SmokeTest) {
...
+
+        fields.o.EraseCol("two");
+        EXPECT_FALSE(fields.o.Rows().Has(3));
+        EXPECT_FALSE(fields.o.Cols().Has("too"));
+        EXPECT_FALSE(fields.o.Cols().Has("two"));
+        EXPECT_EQ(1u, fields.o.Rows().Size());
+        EXPECT_EQ(1u, fields.o.Cols().Size());
+
+        std::multiset<std::string> rows;
+        std::multiset<std::string> cols;
```
mzhurovich: Just curious, why `multiset`?

Storage/container/one2one.h
grixa: Right here or above the whole method?

// Remove and/or overwrite existing objects if they use the same row and/or column
// TODO(grixa): add more flexible behavior with different policies of handling integrity conflicts
Something like this?

mzhurovich: A couple of sentences before the method declaration, explaining that all existing entries with `row` or `col` value matching the corresponding values of the new entry will be deleted.
@dkorolev Do we want to introduce overwrite policy template parameter now?

Storage/container/one2one.h
dkorolev: The number of events may eventually change. For instance, a `deque`, which Redis supports, requires four types of {push/pop} * {front/back} events.

Thus, yes, keeping it as `E1` and `E2` is probably desirable to highlight that the generic code does not have to know what their meanings are.

@grixa @sompylasar @mzhurovich

Storage/container/one2one.h
dkorolev: `OneToOne`, yes, please.

Storage/container/one2one.h
dkorolev: +1 for `T_KEY` (or `key_t`, as we're about to slowly transition to this syntax). I never really liked `T_RCPAIR` =)

Storage/container/one2one.h
dkorolev: Yes please. It's from [Google Style Guide](http://google.github.io/styleguide/cppguide.html).

"... some projects require that an if must always always have an accompanying brace."

```
if (condition)
  DoSomething();  // 2 space indent.

if (condition) {
  DoSomething();  // 2 space indent.
}
```

(ignore the "some projects" part, it's only because they merged a couple loosely related C++ guides together.)

Storage/container/one2one.h
dkorolev: As we're inconclusive about the value of the policy template parameter for overwrites, it's too early to add it.

Too open-ended and too unclear what exactly to implement for a `TODO(grixa)` to make sense. A one- or two-sentences comment before the method, yes, please. :-)

Storage/test.cc
dkorolev: +1

Storage/test.cc
dkorolev: ( without looking at the context ) I recall I used `multiset` in the Matrix test before, to keep counters, as the relationship was two-to-three.

Storage/container/one2one.h
sompylasar: OK. I was thinking of some sort of code completion that suggests E1 while it could suggest what's really should be passed in there. The detailed compilation errors would also be more readable with semantic names. My $0.02, probably you've got other use cases in mind.

dkorolev: Oh, trust me, the detailed compilation error would not be more readable with `T_UPDATE_EVENT` instead of `E1`. ;-)

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+          typename T_UPDATE_EVENT,
+          typename T_DELETE_EVENT,
+          template <typename...> class ROW_MAP,
+          template <typename...> class COL_MAP>
+class GenericOne2One {
+ public:
+  using T_ROW = sfinae::ENTRY_ROW_TYPE<T>;
+  using T_COL = sfinae::ENTRY_COL_TYPE<T>;
+  using T_RCPAIR = std::pair<T_ROW, T_COL>;
+  using T_WHOLE_MAP = std::unordered_map<T_RCPAIR, std::unique_ptr<T>, CurrentHashFunction<T_RCPAIR>>;
```
dkorolev: Perhaps `T_WHOLE_MAP` -> `T_ELEMENTS_MAP` ? "Whole" used to make sense for a dense N-to-N case, as a `T_WHOLE_MATRIX_MAP`; one-to-one is not dense by itself.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+
+  explicit GenericOne2One(MutationJournal& journal) : journal_(journal) {}
+
+  bool Empty() const { return map_.empty(); }
+  size_t Size() const { return map_.size(); }
+
+  void Add(const T& object) {
+    const auto row = sfinae::GetRow(object);
+    const auto col = sfinae::GetCol(object);
+    const auto row_col = std::make_pair(row, col);
```
dkorolev: Nit: If you collapse these three lines into `const auto key = std::make_pair(sfinae::GetRow(object), sfinae::GetCol(object))`, there would be no confusion with respect to `copy_free`, whether `row` has a value type or a const reference type.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+    const auto it = map_.find(row_col);
+    if (it != map_.end()) {
+      const T prev_object = *(it->second);
+      journal_.LogMutation(T_UPDATE_EVENT(object),
+                           [this, row_col, prev_object]() { DoAdd(row_col, prev_object); });
+    } else {
+      const auto it_row = forward_.find(row);
+      const auto it_col = transposed_.find(col);
+      bool row_occupied = it_row != forward_.end();
```
dkorolev: `const bool` here and for `col_occupied`.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+      bool col_occupied = it_col != transposed_.end();
+      if (row_occupied && col_occupied) {
+        const T prev_object1 = *(it_row->second);
```
dkorolev: Nit: why {1/2} if can name them some `same_{row,col}`?

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+        const T prev_object2 = *(it_col->second);
```
dkorolev: And another nit: in Google style guide it's generally not recommended to shorten otherwise short variable names. So, `previous_*` would always be preferred to `prev_*`.

Storage/container/one2one.h
dkorolev: Two readability nits, as you asked for them :-)

1. in `bool a = (b == c);` the parentheses around `(b == c)` help.
2. in `const X x = *y->second`, we usually don't put parentheses around `y->second` (although with this one I actually like your way!)

CC @mzhurovich

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+                             [this, prev_row_col, prev_object]() { DoAdd(prev_row_col, prev_object); });
+        DoErase(prev_row_col);
+      }
+    }
+    journal_.LogMutation(T_UPDATE_EVENT(object), [this, row_col]() { DoErase(row_col); });
+    DoAdd(row_col, object);
+  }
+
+  void Erase(const T_RCPAIR& row_col) {
+    auto it = map_.find(row_col);
```
dkorolev: const auto ?

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+        journal_.LogMutation(T_DELETE_EVENT(prev_object),
+                             [this, prev_row_col, prev_object]() { DoAdd(prev_row_col, prev_object); });
+        DoErase(prev_row_col);
+      }
+    }
+    journal_.LogMutation(T_UPDATE_EVENT(object), [this, row_col]() { DoErase(row_col); });
+    DoAdd(row_col, object);
+  }
+
+  void Erase(const T_RCPAIR& row_col) {
```
dkorolev: All `RCPAIR`, `ROW_COL` and `row_col` are indeed better off as `key`-s.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+  void operator()(const T_UPDATE_EVENT& e) {
+    const auto row = sfinae::GetRow(e.data);
+    const auto col = sfinae::GetCol(e.data);
+    DoAdd(std::make_pair(row, col), e.data);
+  }
+  void operator()(const T_DELETE_EVENT& e) { DoErase(std::make_pair(e.key.first, e.key.second)); }
+
+  template <typename T_MAP>
+  struct MapAccessor final {
+    using T_KEY = typename T_MAP::key_type;
```
dkorolev: @mzhurovich , do you think it's a good time to start migrating to `key_t`, and also add top-level template types as `using map_t = T_MAP` ?

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+  Iterator End() const { return Iterator(map_.end()); }
+
+ private:
+  void DoErase(const T_RCPAIR& row_col) {
+    forward_.erase(row_col.first);
+    transposed_.erase(row_col.second);
+    map_.erase(row_col);
+  }
+
+  void DoAdd(const T_RCPAIR& row_col, const T& object) {
```
dkorolev: Perhaps:
1) another signature with `std::unique_ptr<T>&&` object to `std::move()` / `.swap()` into `map_`, and
2) another signature with `T&& object` to `std::move()` into the `unique_ptr<>` contained in the `map_` ?

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  T_WHOLE_MAP map_;
+  T_FORWARD_MAP forward_;
+  T_TRANSPOSED_MAP transposed_;
+  MutationJournal& journal_;
+};
+
+template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
+using UnorderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Unordered, Unordered>;
```
dkorolev: `2` -> `To` here too pls.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+
+template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
+using OrderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Ordered, Ordered>;
+
+}  // namespace container
+
+template <typename T, typename E1, typename E2>  // Entry, update event, delete event.
+struct StorageFieldTypeSelector<container::UnorderedOne2One<T, E1, E2>> {
+  static const char* HumanReadableName() { return "UnorderedOne2One"; }
```
dkorolev: And `2` -> `To` here and pls.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+};
+
+template <typename T, typename E1, typename E2>  // Entry, update event, delete event.
+struct StorageFieldTypeSelector<container::OrderedOne2One<T, E1, E2>> {
+  static const char* HumanReadableName() { return "OrderedOne2One"; }
```
dkorolev: And `2` -> `To` here pls.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
+using UnorderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Unordered, Unordered>;
+
+template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
+using OrderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Ordered, Ordered>;
+
+}  // namespace container
+
+template <typename T, typename E1, typename E2>  // Entry, update event, delete event.
+struct StorageFieldTypeSelector<container::UnorderedOne2One<T, E1, E2>> {
```
dkorolev: Strictly speaking, in this particular case, a `template... TS` would do it. Probably breaking Visual Studio though, w/o a helper `struct` to wrap the `using` into. Suggest we postpone this until the time we add a contained with more than two mutation event types.

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+
+template <typename T, typename E1, typename E2>  // Entry, update event, delete event.
+struct StorageFieldTypeSelector<container::OrderedOne2One<T, E1, E2>> {
+  static const char* HumanReadableName() { return "OrderedOne2One"; }
+};
+
+}  // namespace storage
+}  // namespace current
+
+using current::storage::container::UnorderedOne2One;
```
dkorolev: And `2` -> `To` here pls, 2x. :-)

Storage/container/one2one.h
```diff
@@ -0,0 +1,253 @@
...
+using current::storage::container::OrderedOne2One;
+
+#endif  // CURRENT_STORAGE_CONTAINER_ONE2ONE_H
```
dkorolev: On, and `2` -> `_TO_` here too pls. And in the file name :-)

Storage/test.cc
```diff
@@ -283,6 +354,47 @@ TEST(TransactionalStorage, SmokeTest) {
       }).Go()));
     }
 
+    // Iterate over the One2One with deleted elements, confirm the integrity of `forward_` and `transposed_`.
```
dkorolev: over the -> over a

Storage/test.cc
dkorolev: So here I would assume a `multiset` makes sure it really is **one** to **one** :)

dkorolev:
@grixa The code LGTM, see a gazillion minor comments inline as you asked. Thanks!

Storage/test.cc
```diff
@@ -308,6 +420,21 @@ TEST(TransactionalStorage, SmokeTest) {
...
+      EXPECT_EQ(4, Value(fields.o.GetRow(3)).phew);
+      EXPECT_FALSE(Exists(fields.o.Get(2, "two")));
+      EXPECT_FALSE(Exists(fields.o.GetRow(2)));
+      EXPECT_FALSE(Exists(fields.o.GetCol("two")));
+      EXPECT_FALSE(Exists(fields.o.Get(2, "too")));
+      EXPECT_TRUE(Exists(fields.o.GetCol("too")));
+      EXPECT_FALSE(Exists(fields.o.GetCol("three")));
+      EXPECT_FALSE(fields.o.Cols().Has("three"));
+      EXPECT_TRUE(fields.o.Cols().Has("too"));
     }).Wait();
```
grixa: @mzhurovich @dkorolev Should I change `.Wait()` to `.Go()` and use the same style as above? Like:
`const auto result = ...`
`EXPECT_FALSE(WasCommitted(result));`

Storage/container/one2one.h
grixa: No problem, I'll rename `*One2One*` to `*OneToOne*` everywhere a bit later in a separate commit

Storage/test.cc
grixa: To be honest, I didn't think about that.. :) But yeah, `multiset` can help us to detect problems with integrity or wrong iteration.

Storage/container/one2one.h
grixa: You mean, `previous_object_same_{row,col}` ? Isn't it too long? If it is, it could be `same_{row,col}_object`, for example. And what about `row_col{1/2}`? `same_{row,col}_key`?

Storage/test.cc
dkorolev: Yes, this works.

Storage/container/one2one.h
dkorolev: This is too minor to spend another comment on this, so use your judgement. My instinct would probably be `existing_per_{row/col}`.

And, to be honest, `previous_object_same_row` is not long at all. It's a four-line-long code block, and worst case scenario it would become a five-line-long block due to line break. Better invest into making sure it doesn't surprise the reader, not into nice- and short-looking names :+1:

Storage/container/one_to_one.h
```diff
@@ -42,65 +42,72 @@ template <typename T,
...
-        const T prev_object2 = *(it_col->second);
-        const auto row_col1 = std::make_pair(row, sfinae::GetCol(prev_object1));
-        const auto row_col2 = std::make_pair(sfinae::GetRow(prev_object2), col);
-        journal_.LogMutation(T_DELETE_EVENT(prev_object1),
-                             [this, row_col1, prev_object1]() { DoAdd(row_col1, prev_object1); });
-        journal_.LogMutation(T_DELETE_EVENT(prev_object2),
-                             [this, row_col2, prev_object2]() { DoAdd(row_col2, prev_object2); });
-        DoErase(row_col1);
-        DoErase(row_col2);
+        const T previous_object_same_row = *(it_row->second);
```
dkorolev: Why aren't these const references? They are captured by value in the lambda regardless.

Storage/container/one_to_one.h
```diff
@@ -108,47 +115,55 @@ class GenericOne2One {
...
+      return ImmutableOptional<T>(FromBarePointer(), it->second);
+    } else {
+      return nullptr;
+    }
   }
 
-  bool CanAdd(const T_RCPAIR& row_col) const {
-    return map_.find(row_col) == map_.end() && forward_.find(row_col.first) == forward_.end() &&
-           transposed_.find(row_col.second) == transposed_.end();
+  bool CanAdd(const T_KEY& key) const {
```
dkorolev: Thinking further, perhaps `CanAdd` should be called `CanAddWithoutConflicts` or `DoesNotConflict`?

Storage/container/one_to_one.h
```diff
@@ -193,13 +208,13 @@ class GenericOne2One {
...
   struct Iterator final {
-    using T_ITERATOR = typename T_WHOLE_MAP::const_iterator;
+    using T_ITERATOR = typename T_ELEMENTS_MAP::const_iterator;
     T_ITERATOR iterator_;
     explicit Iterator(T_ITERATOR iterator) : iterator_(iterator) {}
     void operator++() { ++iterator_; }
     bool operator==(const Iterator& rhs) const { return iterator_ == rhs.iterator_; }
     bool operator!=(const Iterator& rhs) const { return !operator==(rhs); }
-    const T_RCPAIR key() const { return iterator_->first; }
+    const T_KEY key() const { return iterator_->first; }
```
dkorolev: `CF<T_KEY>`?

Storage/container/one_to_one.h
```diff
@@ -207,47 +222,47 @@ class GenericOne2One {
...
   MutationJournal& journal_;
 };
 
 template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
-using UnorderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Unordered, Unordered>;
+using UnorderedOneToOne = GenericOneToOne<T, T_UPDATE_EVENT, T_DELETE_EVENT, Unordered, Unordered>;
 
 template <typename T, typename T_UPDATE_EVENT, typename T_DELETE_EVENT>
-using OrderedOne2One = GenericOne2One<T, T_UPDATE_EVENT, T_DELETE_EVENT, Ordered, Ordered>;
+using OrderedOneToOne = GenericOneToOne<T, T_UPDATE_EVENT, T_DELETE_EVENT, Ordered, Ordered>;
```
dkorolev: This one needs a simple smoke test, confirming that the `Ordered` one does order what the `Unordered` one does not.

Storage/container/one_to_one.h
grixa: Good point. And I don't like `GetRow` and `GetCol` names either. They sound like the methods do directly the opposite. May be, `GetRowEntry` and `GetColEntry` would be better. What do you think?

Storage/test.cc
mzhurovich: If you use `vector`, `EXPECT_EQ("...", Join(...))` will perfectly do the job :) The only question is sort order.

Storage/container/one_to_one.h
```diff
@@ -0,0 +1,268 @@
...
+  using T_FORWARD_MAP = ROW_MAP<T_ROW, const T*>;
+  using T_TRANSPOSED_MAP = COL_MAP<T_COL, const T*>;
+  using T_REST_BEHAVIOR = rest::behavior::Matrix;
+
+  explicit GenericOneToOne(MutationJournal& journal) : journal_(journal) {}
+
+  bool Empty() const { return map_.empty(); }
+  size_t Size() const { return map_.size(); }
+
+  // Adds specified object, removes and/or overwrites all existing entries
```
mzhurovich: Why not write two or three sentences describing the behavior a bit more clearly? `removes and/or overwrites` - so it removes, or overwrites, or both? Does it depend on something? Short comments are awesome, but let's not sacrifice clarity.

Storage/container/one_to_one.h
```diff
@@ -0,0 +1,268 @@
...
+  // with row and/or col value matching the corresponding values of the new entry
```
mzhurovich: Period in the end of the sentence, pls.

Storage/test.cc
```diff
@@ -127,9 +129,10 @@ TEST(TransactionalStorage, SmokeTest) {
   const auto persistence_file_remover = current::FileSystem::ScopedRmFile(persistence_file_name);
 
   {
-    EXPECT_EQ(2u, Storage::FIELDS_COUNT);
+    EXPECT_EQ(3u, Storage::FIELDS_COUNT);
     Storage storage(persistence_file_name);
 
+    // Fill a `Dictionary` container
```
mzhurovich: Period.

Storage/test.cc
```diff
@@ -162,6 +165,8 @@ TEST(TransactionalStorage, SmokeTest) {
       }).Go();
       EXPECT_TRUE(WasCommitted(result));
     }
+
+    // Fill a `Matrix` container
```
mzhurovich: And here.

Storage/test.cc
```diff
@@ -185,9 +190,45 @@ TEST(TransactionalStorage, SmokeTest) {
       EXPECT_TRUE(WasCommitted(result));
     }
 
-    // Iterate over the matrix.
+    // Fill a `OneToOne` container
```
mzhurovich: And here.

Storage/test.cc
```diff
@@ -280,7 +361,50 @@ TEST(TransactionalStorage, SmokeTest) {
         EXPECT_EQ("1,one=1", current::strings::Join(data2, ' '));
 
         CURRENT_STORAGE_THROW_ROLLBACK();
-      }).Go()));
+      }).Go();
+      EXPECT_FALSE(WasCommitted(result));
+    }
+
+    // Iterate over a OneToOne with deleted elements, confirm the integrity of `forward_` and `transposed_`.
```
mzhurovich: s/OneToOne/\`OneToOne\`/

Storage/container/one_to_one.h
grixa:
```
// Adds specified object and overwrites existing one if it has the same row and col.
// Removes all other existing objects with the same row or col.
```
?

Storage/test.cc
grixa: @mzhurovich Good idea :) I used it to compare ordered and unordered versions of the containers, see corresponding commit below.

Storage/container/one_to_one.h
grixa: Done, see [this commit](https://github.com/C5T/Current/pull/448/commits/463d9e6b73378855917e11d78e0175fb818cd1c8). May be, it's not enough, but still better than nothing

PR #449, opened 2016-04-03
Removed redundant std::move uses

grixa:
Removed redundant `std::move`s, because they cause clang warnings like: `warning: moving a temporary object prevents copy elision`. And because they are redundant :)

CC @dkorolev @mzhurovich


dkorolev:
LGTM++

PR #450, opened 2016-04-04
Persistence fix: `vector`->`deque`.

mzhurovich:


Blocks/Persistence/file.h
```diff
@@ -116,7 +117,7 @@ class FilePersister {
 
     // `offset.size() == end.index`, and `offset[i]` is the offset in bytes where the line for index `i` begins.
     std::mutex mutex;
-    std::vector<std::streampos> offset;
+    std::deque<std::streampos> offset;
```
dkorolev: Looks like `offset` is used only from within a mutex-locked section, no?

mzhurovich: You think I should revert it for `file.h`?

PR #453, opened 2016-04-04
Removed redundant update_event in OneToOne in case of existing element overwriting

grixa:
CC @dkorolev @mzhurovich


dkorolev:
@grixa LGTM && merging. Would be great if the complete test for this includes validating the contents of the underlying stream / raw log.

PR #455, opened 2016-04-10
Used one Iterator type in OneToOne container instead of two almost identical ones

grixa:
CC @dkorolev @mzhurovich

### DIFF
--- a/Storage/container/one_to_one.h
+++ b/Storage/container/one_to_one.h
@@ -177,40 +177,9 @@ class GenericOneToOne {
   void operator()(const T_DELETE_EVENT& e) { DoErase(std::make_pair(e.key.first, e.key.second)); }
 
   template <typename T_MAP>
-  struct MapAccessor final {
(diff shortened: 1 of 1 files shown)


Storage/container/one_to_one.h
```diff
@@ -201,27 +203,16 @@ class GenericOneToOne {
 
     bool Has(const T_KEY& x) const { return map_.find(x) != map_.end(); }
 
-    Iterator begin() const { return Iterator(map_.cbegin()); }
-    Iterator end() const { return Iterator(map_.cend()); }
+    Iterator<T_MAP> begin() const { return Iterator<T_MAP>(map_.cbegin()); }
```
dkorolev: :+1:

@grixa , I'll merge it now. As a nit, a slightly cleaner way is to rename this now templated `Iterator` into some `IteratorImpl` or `MetaIterator`, and add two respective `using`-s, just in case these inner type names are publicly exposed and in use by the outside code.

Thanks!

PR #462, opened 2016-04-13
`ScopeOwned` begins pwning.

dkorolev:
Hi Max,

I started looking into making our Sherlock safer and more error-resilient by replacing `shared_ptr`-s by `ScopeOwned`-s. First small step: iterators over our persisters had `ScopeOwned`-s, but they were stubs. I've added the corresponding logic and tested it.

A small yet open-ended question: exceptions hierarchy. In this change I've added `GracefulShutdownException`into `Blocks`, by creating a top-level `exceptions.h` in `Blocks`. On the grand scheme of things it sounds right, as `Bricks` are supposed to be trivial pieces, while `Blocks` are the big ones. This, however, raises a question whether the exceptions `ScopeOwned` itself throws are trivial- or sophisticated-level, and whether `Bricks/sync` should be in `Blocks`.

For the record, I'm thinking of the following dependencies to be controlled by ScopeOwned. X > Y means Y should be contained in the scope of X.

- [x] Persister > Persister Iterable Range
- [x] Persister Iterable Range > Persister Iterator

- [ ] Stream ( + its Persister) > each Stream Subscriber (here and below I'm assuming Stream simply has Persister as its member)
- [ ] Master Stream ( + its Persister ) > Stream Publisher (the Publisher which can be `std::move()`-d around.)

- [ ] Storage -> Stream::Persister::IterableRange (during the initial replay phase)
- [ ] Following Storage -> (Stream or other) Subscriber (following Storage can be explicitly left with no Subscriber, but if the Subscriber is being destructed without being detached from a Following Storage, the latter should probably die)

- [ ] Master Storage > Persister (Stream or other)

Thanks,
Dima

### DIFF
--- a/Blocks/exceptions.h
+++ b/Blocks/exceptions.h
@@ -1,0 +1,40 @@
+/*******************************************************************************
+The MIT License (MIT)
+
+Copyright (c) 2016 Maxim Zhurovich <zhurovich@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPREPERSISTENCE OR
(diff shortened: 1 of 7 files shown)


Blocks/exceptions.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef BLOCKS_EXCEPTIONS_H
+#define BLOCKS_EXCEPTIONS_H
+
+#include "../Bricks/exception.h"
+
+namespace current {
+
+// The base class for exceptions originated from the fact that the desired resource is inaccessble
+// as the scope controlling it has already entered its graceful shutdown mode.
+struct GracefulShutdownException : current::Exception {
```
mzhurovich: Hm. What about `Blocks/GracefulShutdown/exceptions.h`?

PR #466, opened 2016-04-14
Unify matrix containers impls

grixa:
CC @dkorolev @mzhurovich

### DIFF
--- a/Storage/container/many_to_many.h
+++ b/Storage/container/many_to_many.h
@@ -186,7 +186,7 @@ class GenericManyToMany {
 
   OuterAccessor<transposed_map_t> Cols() const { return OuterAccessor<transposed_map_t>(transposed_); }
 
-  // For REST, iterate over all the elemnts of the ManyToMany, in no particular order.
+  // For REST, iterate over all the elements of the ManyToMany, in no particular order.
   using iterator_t = IteratorImpl<whole_matrix_map_t>;
   iterator_t begin() const { return iterator_t(map_.begin()); }
   iterator_t end() const { return iterator_t(map_.end()); }
--- a/Storage/container/many_to_many.h
+++ b/Storage/container/many_to_many.h
@@ -69,7 +69,7 @@ class GenericManyToMany {
     const auto key = std::make_pair(row, col);
     const auto it = map_.find(key);
     if (it != map_.end()) {
-      const T previous_object = *(it->second);
+      const T& previous_object = *(it->second);
       journal_.LogMutation(UPDATE_EVENT(object),
                            [this, key, previous_object]() { DoAdd(key, previous_object); });
     } else {
@@ -78,30 +78,27 @@ class GenericManyToMany {
     DoAdd(key, object);
   }
 
-  void Erase(sfinae::CF<row_t> row, sfinae::CF<col_t> col) {
-    const auto key = std::make_pair(row, col);
+  void Erase(const key_t& key) {
     const auto it = map_.find(key);
     if (it != map_.end()) {
-      const T previous_object = *(it->second);
+      const T& previous_object = *(it->second);
       journal_.LogMutation(DELETE_EVENT(previous_object),
-                           [this, row, col, key, previous_object]() {
-                             auto& placeholder = map_[key];
-                             placeholder = std::make_unique<T>(previous_object);
-                             forward_[row][col] = placeholder.get();
-                             transposed_[col][row] = placeholder.get();
-                           });
+                           [this, key, previous_object]() { DoAdd(key, previous_object); });
       DoErase(key);
     }
   }
-  void Erase(const key_t& key) { Erase(key.first, key.second); }
+  void Erase(sfinae::CF<row_t> row, sfinae::CF<col_t> col) { Erase(std::make_pair(row, col)); }
 
-  ImmutableOptional<T> Get(sfinae::CF<row_t> row, sfinae::CF<col_t> col) const {
-    const auto it = map_.find(std::make_pair(row, col));
+  ImmutableOptional<T> operator[](const key_t& key) const {
+    const auto it = map_.find(key);
     if (it != map_.end()) {
       return ImmutableOptional<T>(FromBarePointer(), it->second.get());
     } else {
       return nullptr;
     }
+  }
+  ImmutableOptional<T> Get(sfinae::CF<T_ROW> row, sfinae::CF<T_COL> col) const {
+    return operator[](std::make_pair(row, col));
   }
 
   void operator()(const UPDATE_EVENT& e) {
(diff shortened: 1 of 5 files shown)


Storage/container/many_to_many.h
```diff
@@ -238,27 +220,27 @@ class GenericMatrix {
...
+  static const char* HumanReadableName() { return "OrderedManyToMany"; }
 };
 
 }  // namespace storage
 }  // namespace current
 
-using current::storage::container::UnorderedMatrix;
-using current::storage::container::OrderedMatrix;
+using current::storage::container::UnorderedManyToMany;
+using current::storage::container::OrderedManyToMany;
```
dkorolev: Let's be backwards-compatible for a while here, could you add two more `using`-s to externally expose two old names as well?

Storage/container/many_to_many.h
```diff
@@ -110,24 +108,27 @@ class GenericMatrix {
   }
   void operator()(const DELETE_EVENT& e) { DoErase(std::make_pair(e.key.first, e.key.second)); }
 
+  template <typename MAP>
```
dkorolev: This generic Iterator class is too generic for `Storage/`, IMHO, how about moving it into `Bricks/util/iterator.h` (and adding a small test for it too)?

Storage/container/many_to_many.h
dkorolev: Perhaps even not here but into a `matrix.h` file (which should just be one `#include` and two `using`-s for a few commits from now on)?

Storage/container/one_to_many.h
```diff
@@ -137,78 +137,81 @@ class GenericOneToMany {
...
 
-  const OuterAccessor<forward_map_t> Rows() const { return OuterAccessor<forward_map_t>(forward_); }
+  const RowsAccessor<forward_map_t> Rows() const { return RowsAccessor<forward_map_t>(forward_); }
 
-  const InnerAccessor<transposed_map_t> Cols() const { return InnerAccessor<transposed_map_t>(transposed_); }
+  const ElementsAccessor<transposed_map_t> Cols() const { return ElementsAccessor<transposed_map_t>(transposed_); }
 
-  Iterator<elements_map_t> begin() const { return Iterator<elements_map_t>(map_.begin()); }
-  Iterator<elements_map_t> end() const { return Iterator<elements_map_t>(map_.end()); }
+  // For REST, iterate over all the elemnts of the OneToMany, in no particular order.
```
dkorolev: Typo, `elemnts` -> `elements`.

Storage/container/many_to_many.h
```diff
@@ -185,50 +186,31 @@ class GenericMatrix {
...
-    void operator++() { ++iterator; }
-    bool operator==(const WholeMatrixIterator& rhs) const { return iterator == rhs.iterator; }
-    bool operator!=(const WholeMatrixIterator& rhs) const { return !operator==(rhs); }
-    const std::pair<row_t, col_t> key() const { return iterator->first; }
-    const T& operator*() const { return *iterator->second; }
-    const T* operator->() const { return iterator->second; }
-  };
-  WholeMatrixIterator WholeMatrixBegin() const { return WholeMatrixIterator(map_.begin()); }
-  WholeMatrixIterator WholeMatrixEnd() const { return WholeMatrixIterator(map_.end()); }
+  // For REST, iterate over all the elemnts of the ManyToMany, in no particular order.
```
dkorolev: Typo, `elemnts` -> `elements`.

Storage/container/one_to_one.h
```diff
@@ -203,16 +204,18 @@ class GenericOneToOne {
...
+    iterator_t end() const { return iterator_t(map_.cend()); }
   };
 
   const MapAccessor<forward_map_t> Rows() const { return MapAccessor<forward_map_t>(forward_); }
 
   const MapAccessor<transposed_map_t> Cols() const { return MapAccessor<transposed_map_t>(transposed_); }
 
-  Iterator<elements_map_t> begin() const { return Iterator<elements_map_t>(map_.begin()); }
-  Iterator<elements_map_t> end() const { return Iterator<elements_map_t>(map_.end()); }
+  // For REST, iterate over all the elemnts of the OneToMany, in no particular order.
```
dkorolev: Typo, `elemnts` -> `elements`.

Storage/container/many_to_many.h
```diff
@@ -57,51 +58,48 @@ class GenericMatrix {
...
-                             forward_[row][col] = placeholder.get();
-                             transposed_[col][row] = placeholder.get();
-                           });
-      DoErase(row_col);
+                           [this, key, previous_object]() { DoAdd(key, previous_object); });
+      DoErase(key);
     }
   }
-  void Erase(const std::pair<row_t, col_t>& key) { Erase(key.first, key.second); }
+  void Erase(sfinae::CF<row_t> row, sfinae::CF<col_t> col) { Erase(std::make_pair(row, col)); }
```
dkorolev: Interestingly, the other way is more efficient: do the logic in a two-parameter version accepting `CF<row_t>, CF<col_t>` as parameters, and have the `CF<key_t>` version call the two-parameter one.

If my understanding is right, with the way it is right now, `make_pair` would create a temporary object by creating copies of both row and col. Undesirable.

(Not to mention, `CF<pair<X, Y>>` is equivalent to `pair<X, Y>` :) )

Storage/container/many_to_many.h
mzhurovich: Is it really used anywhere? I mean, do we really need backwards compatibility?

Storage/container/one_to_many.h
```diff
@@ -137,78 +137,81 @@ class GenericOneToMany {
...
 
-  template <typename OUTER_MAP>
-  struct OuterAccessor final {
-    using OUTER_KEY = typename OUTER_MAP::key_type;
-    using INNER_MAP = typename OUTER_MAP::mapped_type;
-    const OUTER_MAP& map_;
+  template <typename ROWS_MAP>
+  struct RowsAccessor final {
+    using key_t = typename ROWS_MAP::key_type;
+    using ELEMENTS_MAP = typename ROWS_MAP::mapped_type;
```
mzhurovich: s/ELEMENTS_MAP/elements_map_t/?

Storage/container/many_to_many.h
dkorolev: Good point. No need to keep `Matrix`!

Storage/container/many_to_many.h
grixa: Yes, but the previous version of `Erase(row, col)` also use make_pair to create temporary object on stack. So, both old Erases creates temporary copy of `row` and `col`, but the new `Erase(key)` version is not. Am I wrong?

dkorolev: You're right. If there's a `std::pair<row_t, col_t>`as a key in `std::[unordered_]map`, the `Erase(row, col)` method with two signatures would inevitably create that `std::pair<row_t, col_t>` on the stack.

Good point. Confusion resolved. :-)

mzhurovich:
LGTM, merging.

PR #472, opened 2016-04-18
Replaced `inet_ntoa` with `inet_ntop`.

mzhurovich:
This is supposed to fix a problem with non-multithreading-friendly `inet_ntoa`.

### DIFF
--- a/Bricks/net/tcp/impl/posix.h
+++ b/Bricks/net/tcp/impl/posix.h
@@ -183,7 +183,9 @@ struct IPAndPort {
 };
 
 inline std::string InetAddrToString(const struct in_addr* in) {
-  const char* result = ::inet_ntoa(*in);
+  // 16 bytes buffer for IPv4.
+  char buffer[16];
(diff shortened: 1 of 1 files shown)


Bricks/net/tcp/impl/posix.h
```diff
@@ -183,7 +183,9 @@ struct IPAndPort {
 };
 
 inline std::string InetAddrToString(const struct in_addr* in) {
-  const char* result = ::inet_ntoa(*in);
+  // 16 bytes buffer for IPv4.
```
dkorolev: I'll merge it to avoid opening a can of worms.

That said:
1) Won't this just work for IPv6 if the buffer is increased (i.e., won't our connect methods just work?)
2) 
```
/usr/include/netinet/in.h:#define INET6_ADDRSTRLEN 46
/usr/include/netinet/in.h:#define INET_ADDRSTRLEN 16
```

Perhaps pick one of the constants (and `#define` it in `port.h` if it's not there on Windows or somewhere else?)

PR #475, opened 2016-04-22
Graphviz binding.

dkorolev:


Bricks/dot/test.cc
```diff
@@ -0,0 +1,147 @@
...
+}
+
+TEST(GraphViz, TextEscaping) {
+  using namespace current::graphviz;
+
+  Graph g;
+  g += Node("plain");
+  g += Node("with whitespace");
+  g += Node("multi\nline");
+  g += Node("contains 'double-quoted' text");
```
mzhurovich: s/double/single/?

dkorolev: :+1: 7a5d5a4

PR #480, opened 2016-04-28
Simple Nginx manager + config generator.

mzhurovich:


3rdparty/gtest/gtest.h
```diff
@@ -10,4 +10,26 @@ inline std::string CurrentTestName() {
   return testing::GetUnitTestImpl()->current_test_info()->name();
 }
 
+// Effectively returns `argv[0]`.
```
dkorolev: License header for this file? Perhaps even more it to `Bricks/` ?

Utils/Nginx/config.h
```diff
@@ -0,0 +1,207 @@
...
+
+#include "../../port.h"
+
+#include "exceptions.h"
+
+#include "../../Bricks/strings/util.h"
+#include "../../TypeSystem/struct.h"
+
+namespace current {
+namespace utils {
```
dkorolev: `utils` look too much to my taste. `current::nginx::config` is perfect.

Utils/Nginx/config.h
```diff
@@ -0,0 +1,207 @@
...
+  CURRENT_CONSTRUCTOR(BlockDirective)(const std::string& name, const std::string& parameters)
+      : name(name), parameters(parameters) {}
+
+  SimpleDirective& AddSimpleDirective(const std::string& name, const std::string& parameters) {
+    directives.push_back(SimpleDirective(name, parameters));
+    return Value<SimpleDirective>(directives.back());
+  }
+
+  template <typename T>
+  T& AddDirective(T && directive) {
```
dkorolev: Format.

Utils/Nginx/config.h
```diff
@@ -0,0 +1,207 @@
...
+  CURRENT_CONSTRUCTOR(LocationDirective)(const std::string& location) : SUPER("location", location) {}
+
+  LocationDirective& AddNestedLocation(const std::string& location) {
+    return SUPER::AddDirective(LocationDirective(location));
+  }
+};
+// clang-format on
+
+inline LocationDirective DefaultLocationDirective(const std::string& location) {
+  LocationDirective result(location);
```
dkorolev: Dot notation?

Utils/Nginx/config.h
```diff
@@ -0,0 +1,207 @@
...
+  result.AddSimpleDirective("proxy_set_header", "X-Real-IP $remote_addr");
+  result.AddSimpleDirective("proxy_set_header", "X-Forwarded-For $proxy_add_x_forwarded_for");
+  result.AddSimpleDirective("proxy_pass_request_headers", "on");
```
dkorolev: :+1: for keeping this one in mind.

Utils/Nginx/config.h
```diff
@@ -0,0 +1,207 @@
...
+
+struct ConfigPrinter {
+  struct DirectivesPrinter {
+    DirectivesPrinter(std::ostream& os, size_t indent) : os_(os), indent_(indent) {}
+
+    void operator()(const SimpleDirective& simple) const {
+      PrintIndent();
+      os_ << simple.name;
+      if (!simple.parameters.empty()) {
+        os_ << ' ' << simple.parameters;
```
dkorolev: No quotes and escaping?

Utils/Nginx/config.h
```diff
@@ -0,0 +1,207 @@
...
+        directive.Call(*this);
+      }
+      --indent_;
+      PrintIndent();
+      os_ << "}\n";
+    }
+
+    void PrintIndent() const {
+      for (size_t i = 0; i < indent_; ++i) {
+        os_ << '\t';
```
dkorolev: Two spaces? :-/

Utils/Nginx/config.h
mzhurovich: Could be confusing if `AddSimpleDirective` returns a reference to a parent, since `AddBlockDirective` should return a reference to a newly created object.

Utils/Nginx/config.h
```diff
@@ -0,0 +1,207 @@
...
+      }
+    }
+    return SUPER::AddDirective(ServerDirective(port));
+  }
+};
+// clang-format on
+
+using full_config_directives_t = Variant<SimpleDirective, BlockDirective>;
+
+CURRENT_STRUCT(FullConfig) {
```
dkorolev: Could `ServerConfig` or `ServiceConfig` or `HTTPServiceConfig` be a better name?

dkorolev: Oh. Take it back after another read. Talk in half an hour :)

Utils/Nginx/config.h
mzhurovich: By default they use tabs. Switching to double space?

Utils/Nginx/config.h
mzhurovich: Don't get it.

Utils/Nginx/nginx.h
```diff
@@ -0,0 +1,72 @@
...
+    return ExecNginx("-t -c " + config_full_path) == 0;
+  }
+
+  bool ReloadConfig() {
+    return ExecNginx("-s reload") == 0;
+  }
+
+ private:
+  int ExecNginx(const std::string& args) {
+#ifdef CURRENT_WINDOWS
```
dkorolev: This part belongs higher up, perhaps even in `base/port.h`.

Utils/Nginx/test.cc
```diff
@@ -0,0 +1,76 @@
...
+TEST(Nginx, FullConfig) {
+  using current::FileSystem;
+  using namespace current::utils::nginx::config;
+
+  FullConfig config;
+  config.AddSimpleDirective("worker_processes", "1");
+  config.AddBlockDirective("events", "").AddSimpleDirective("worker_connections", "1024");
+  config.http.AddSimpleDirective("default_type", "application/octet-stream");
+  config.http.AddSimpleDirective("keepalive_timeout", "100");
+  auto& server = config.http.AddServer(8910);
```
dkorolev: Why not `PickPortForUnitTest()` ?

Utils/Nginx/nginx.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <cstdlib>
+
+#include "config.h"
+
+namespace current {
+namespace utils {
+namespace nginx {
+
+class NginxManager {
```
dkorolev: After reading the test, perhaps `NginxInvoker` is a better name? `Manager` doesn't convey the message that this class makes external shell calls.

Utils/Nginx/test.cc
mzhurovich: Because of troubles with the golden file :-/

Utils/Nginx/config.h
dkorolev: What you're saying is more confusing. Perhaps add a comment header with what returns what?

If `.Add()` calls can't be dot-chained, how about making the following syntax legal: `return LocationDirective(location) + SimpleDirective(...) + SimpleDirective(...)`? It only takes adding an `operator+=`, which should by definition return `*this`, and then an `operator+`, which creates a copy of `*this` first and then returns it.

Utils/Nginx/config.h
dkorolev: Yes please.

Utils/Nginx/config.h
dkorolev: What if `simple_parameters` contains spaces or quotes? Perhaps `os_ << " \"" << Escaped(simple.parameters) << '\";` is the right way? A link to the docu would be nice if it's not.

Utils/Nginx/test.cc
dkorolev: Ah. /me stupid.

3rdparty/gtest/gtest.h
mzhurovich: Then move `gtest-main.h` and `gtest-main-with-dflags.h` to `Bricks/gtest` => massive replace. Maybe not now?

dkorolev: Yes, certainly not now.

mzhurovich:
@dkorolev PTAL!

Utils/Nginx/test.cc
```diff
@@ -35,20 +35,19 @@ DEFINE_bool(nginx_overwrite_golden_files, false, "Set to true to have Nginx gold
...
-  config.http.AddSimpleDirective("keepalive_timeout", "100");
-  auto& server = config.http.AddServer(8910);
-  EXPECT_THROW(config.http.AddServer(8910), current::utils::nginx::PortAlreadyUsedException);
-  server.AddSimpleDirective("server_name", "myserver");
-  server.AddDefaultLocation("/").AddSimpleDirective("proxy_pass", "http://127.0.0.1:8888/");
-  auto& complex_location = server.AddLocation("~ /a|/b");
-  complex_location.AddNestedLocation("~ /a");
-  complex_location.AddNestedLocation("~ /b");
+  config.Add(SimpleDirective("worker_processes", "1"));
+  config.Add(BlockDirective("events", {SimpleDirective("worker_connections", "1024")}));
```
dkorolev: A test doing `config.Add(BlockDirective("foo").Add(SimpleDirective("bar", "baz")).Add(SimpleDirective("meh", "nah")));` would be nice too.

Great code, LGTM++ && merging.

PR #489, opened 2016-05-02
Fixed the flakiness.

dkorolev:
It was the first number in the `Generator` thread calling `SetNow()` zero, making it non-monotonic. Fixed by starting from 1'000'000 instead.


dkorolev:
Passed all with `--gtest_repeat=1000` and `Generator` with `--gtest_repeat=1000`.

PR #492, opened 2016-05-07
`CURRENT_FIELD_DESCRIPTION`, and fixing the top-level `make test`.

dkorolev:


TypeSystem/base.h
```diff
@@ -47,13 +47,18 @@ struct CountFieldsImplementationType {
...
 class FieldTypeAndName {};
+class FieldTypeAndNameAndIndex {};
 
 template <typename SELF>
 class FieldNameAndPtr {};
 
 class FieldNameAndImmutableValue {};
 class FieldNameAndMutableValue {};
 
+// Simple index. TODO(dkorolev): Retire the copmlex `Index` below, it may speed up compilation.
```
sompylasar: `copmlex` -> `complex`

TypeSystem/Schema/smoke_test_struct.h
```diff
@@ -20,6 +21,7 @@ CURRENT_STRUCT(Primitives) {
   CURRENT_FIELD(m, bool);
   CURRENT_FIELD(n, std::chrono::microseconds);
   CURRENT_FIELD(o, std::chrono::milliseconds);
+  CURRENT_FIELD_DESCRIPTION(b, "Field descriptions can be set in any order.");
```
sompylasar: Test newlines in the descriptions?

examples/EventStore/test.cc
```diff
@@ -41,7 +41,7 @@ DEFINE_string(event_store_test_tmpdir, ".", "Local path for the test to create t
 DEFINE_int32(event_store_test_port, PickPortForUnitTest(), "Local port to run the test against.");
 
 TEST(EventStore, SmokeWithInMemoryEventStore) {
-  current::time::SetNow(std::chrono::microseconds(0), std::chrono::microseconds(100));
+  current::time::ResetToZero();
```
mzhurovich: Is it safe to replace range `SetNow()` with just a `ResetToZero()` call?

TypeSystem/base.h
dkorolev: :+1: b2d46ce

examples/EventStore/test.cc
dkorolev: Yes, by design `ResetToZero()` initializes high range upper limit as well.

TypeSystem/Schema/smoke_test_struct.h
dkorolev: Let me see how plausible this would be. Keep in mind, we have F# and Markdown export formats, where C++-grade escaping is probably not the wisest choice.

dkorolev: Oh, Markdown, at least the GitHub one, supports `<br>`. Doing it.

dkorolev: Done, 493832f.
A bit messy in plain C struct output format, but should do it for now.

dkorolev:
@mzhurovich, incorporated comments from @sompylasar, PTAL.

TypeSystem/Schema/smoke_test_struct.h
sompylasar: For F# and C, I'd suggest to place the description on the line right before the field, and separate such described fields from the other ones with empty lines. BTW, escaping of `'` is not needed in the comments.
```
type Primitives = {
  // It's the order of fields that matters.
  a : byte

  // Field descriptions can be set in any order.
  b : uint16

  c : uint32
  d : uint64
  e : sbyte
  f : int16
  g : int32
  h : int64
  i : char
  j : string
  k : float
  l : double

  // Multiline
  // descriptions
  // can be used.
  m : bool

  // microseconds.
  n : int64

  // milliseconds.
  o : int64
}
```

Bricks/strings/test.cc
```diff
@@ -642,9 +643,16 @@ TEST(TimeIntervalAsString, SmokeTest) {
 
 TEST(Util, Escape) {
   EXPECT_EQ("Smoke.", EscapeForCPlusPlus("Smoke."));
+  EXPECT_EQ("vertical|bar", EscapeForCPlusPlus("vertical|bar"));
   EXPECT_EQ("new\\nline", EscapeForCPlusPlus("new\nline"));
   EXPECT_EQ("t\\ts\\tv", EscapeForCPlusPlus("t\ts\tv"));
   EXPECT_EQ("Testing \\'quote\\' and \\\"quote\\\".", EscapeForCPlusPlus("Testing 'quote' and \"quote\"."));
```
mzhurovich: Pretty strange escaping imho :-/

TypeSystem/Schema/smoke_test_struct.h
dkorolev: Great idea! Doing now. Nit: the `microseconds` and `milliseconds` are not user-level comments but type comments (annotations for `int64` as their C++ types are `std::chrono::{micro/milli}seconds` respectively). They would stay the way it is right now.

Bricks/strings/test.cc
sompylasar: I'd rename `EscapeForCPlusPlus` to `EscapeForCPlusPlusLikeLineComment` and remove the quote escaping.

dkorolev: It's used to generate the `CURRENT_FIELD_DESCRIPTION(name, "blah")`, where `"blah"` should compile.

TypeSystem/Schema/smoke_test_struct.h
dkorolev: :+1: e103e47

dkorolev:
@sompylasar @mzhurovich, please take a look at e103e47.

No real corner case checking is performed (ex. a trailing `\` at the end of C++ comment making it to the next line, or something in F# I'm not aware of), but should be good for all our practical purposes for the time being.

Thanks!
Dima

mzhurovich:
@dkorolev LGTM!

PR #494, opened 2016-05-07
Generalized map iterator and accessor

grixa:
CC @dkorolev @mzhurovich

### DIFF
--- a/Bricks/util/iterator.h
+++ b/Bricks/util/iterator.h
@@ -1,18 +1,18 @@
 /*******************************************************************************
  The MIT License (MIT)
- 
+
  Copyright (c) 2016 Grigory Nikolaenko <nikolaenko.grigory@gmail.com>
- 
+
  Permission is hereby granted, free of charge, to any person obtaining a copy
  of this software and associated documentation files (the "Software"), to deal
  in the Software without restriction, including without limitation the rights
  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  copies of the Software, and to permit persons to whom the Software is
  furnished to do so, subject to the following conditions:
- 
+
  The above copyright notice and this permission notice shall be included in all
  copies or substantial portions of the Software.
- 
+
  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


Bricks/util/iterator.h
```diff
@@ -0,0 +1,68 @@
...
+  const value_t* operator->() const { return is_unique_ptr<mapped_t>::pointer(iterator_->second); }
+};
+
+template <typename MAP>
+struct GenericMapAccessor final {
+  using iterator_t = GenericMapIterator<MAP>;
+  using key_t = typename MAP::key_type;
+  const MAP& map_;
+
+  GenericMapAccessor(const MAP& map) : map_(map) {}
```
dkorolev: `explicit`.

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
   EXPECT_EQ("400:7", bar_y_q.InstantiateAsUniquePtrWithExtraParameter(400)->AsString());
 }
 
+TEST(Util, Iterator) {
+  typedef std::map<int, std::string> map1_t;
+  map1_t map1;
+  auto accessor1 = current::GenericMapAccessor<map1_t>(map1);
```
dkorolev: `const auto`?

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
...
+  EXPECT_TRUE(accessor1.Empty());
+  EXPECT_EQ(0u, accessor1.Size());
+  EXPECT_EQ(accessor1.begin(), accessor1.end());
```
dkorolev: This probably best as `EXPECT_TRUE(accessor1.begin() == accessor1.end())`. What the `EXCEPT_EQ` would print as `Actual:` and `Expected:` would be unreadable anyway.

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
...
+  for (char c = 'a'; c <= 'z'; ++c) {
+    map1[c - 'a'] = c;
+  }
+  EXPECT_FALSE(accessor1.Empty());
+  EXPECT_EQ(26u, accessor1.Size());
+  EXPECT_FALSE(accessor1.begin() == accessor1.end());
+  std::string str1;
+  for (const auto& element : accessor1) {
+    str1 += current::ToString(element);
```
dkorolev: Isn't `current::ToString()` redundant here?

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
   EXPECT_EQ("400:7", bar_y_q.InstantiateAsUniquePtrWithExtraParameter(400)->AsString());
 }
 
+TEST(Util, Iterator) {
+  typedef std::map<int, std::string> map1_t;
```
dkorolev: `typedef` -> `using`?

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
...
+  EXPECT_FALSE(accessor1.Empty());
+  EXPECT_EQ(26u, accessor1.Size());
+  EXPECT_FALSE(accessor1.begin() == accessor1.end());
+  std::string str1;
+  for (const auto& element : accessor1) {
+    str1 += current::ToString(element);
+  }
+  EXPECT_EQ("abcdefghijklmnopqrstuvwxyz", str1);
+
+  typedef std::map<int, std::unique_ptr<std::string>> map2_t;
```
dkorolev: `typedef` -> `using`?

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
...
+  map2_t map2;
+  auto accessor2 = current::GenericMapAccessor<map2_t>(map2);
+  EXPECT_TRUE(accessor2.Empty());
+  EXPECT_EQ(0u, accessor2.Size());
+  EXPECT_EQ(accessor2.begin(), accessor2.end());
+  for (auto it = accessor1.begin(), eit = accessor1.end(); it != eit; ++it) {
```
dkorolev: @sompylasar you gonna love @grixa's for-loop declaration style. :-)

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
...
+    map2[it.key()] = std::make_unique<std::string>(*it);
+  }
+  EXPECT_FALSE(accessor2.Empty());
+  EXPECT_EQ(26u, accessor2.Size());
+  EXPECT_FALSE(accessor2.begin() == accessor2.end());
+  std::string str2;
+  for (const auto& element : accessor2) {
+    str2 += current::ToString(element);
```
dkorolev: Intuitively, `current::ToString` should not be required here too, right?

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
...
+  }
+  EXPECT_EQ("abcdefghijklmnopqrstuvwxyz", str2);
```
dkorolev: Perhaps initialize the 2nd string with `(c - 'a' + 'A')` (or the 1st one with capital and the 2nd one with lowercase), to reduce room for human error?

Very clean test BTW.

dkorolev:
@grixa LGTM++, comments are mostly style nits. Also, perhaps `Pointer` and `Extract` capitalized, instead of `Pointer` and `Extract`?

Bricks/util/test.cc
sompylasar: `for (var ic = arr.length, i = 0; i < ic; ++i)` – I still prefer to define end before begin to be able to reference it in the `i = 0` or `i = ic - 1` initializer clause.

Bricks/util/test.cc
```diff
@@ -517,6 +518,44 @@ TEST(Util, LazyInstantiation) {
   EXPECT_EQ("400:7", bar_y_q.InstantiateAsUniquePtrWithExtraParameter(400)->AsString());
 }
 
+TEST(Util, Iterator) {
```
mzhurovich: `Iterator`->`GenericMapAccessor`?

mzhurovich:
@grixa LGTM and ready to merge!
Minor nit: let's rename the test to be more specific on what's going on there.

Bricks/util/test.cc
dkorolev: You and your JavaScript style in C++. Still remember the suggestion to format ternary operator on three lines with an extra pair of parentheses :)

sompylasar: On five lines, actually :)
```
const x = (
  condition
    ? true
    : false
);
```
Can be shortened to four, but looks a little condensed.
```
const x = (condition
    ? true
    : false
);
```

dkorolev: Let's just agree an average C++ programmer knows a bit more about operator precedence compared to an average JavaScript one. The compiler reminding us to put semicolons where they belong kinda helps too. ;-)

sompylasar: Knowing more about operator precedence doesn't help much reading ternaries, especially nested :)

PR #498, opened 2016-05-09
Added partially ordered versions of matrix-style storage containers

grixa:
CC @dkorolev @mzhurovich


dkorolev:
As my only concern with this change has been that it may break something before yesterday's release, I'll merge it in and see if everything compiles :-)

PR #502, opened 2016-05-15
Who could have possibly imagined?

dkorolev:


sompylasar:
Strict typing strikes back 🎱

PR #503, opened 2016-05-16
Type evolution prototyping.

mzhurovich:


TypeSystem/struct.h
```diff
@@ -438,13 +438,28 @@ struct SuperTypeImpl {
...
+  };
+};
+
+template <FieldCounterPolicy POLICY>
+struct FieldCounter<CurrentStruct, POLICY> {
+  enum { value = 0 };
 };
 
+template <typename T>
+using TotalFieldCounter = FieldCounter<T, FieldCounterPolicy::IncludingSuperFields>;
```
dkorolev: :+1:

dkorolev:
With the one-to-one evolution model we converged to, it's unclear to a fresh eye what's the benefit of populating and returning `into` from a function accepting `from` over the solution where the evolution evolution function takes two arguments: `from` and `into`.

Merging nonetheless. Great job!

dkorolev:
`copy_free<>` works for me. Let me look into build time. Now it's `4.3s g++`, `2.9s clang++` for `make test` from `Evolution/`...

PR #504, opened 2016-05-17
Type evolution prototype follow-up after an offline chat.

dkorolev:


dkorolev:
Running "perf test" for the compiler on Hetz.

**Pre this PR**

```
export CPLUSPLUS=g++ && time for i in `seq 10`; do touch test.cc ; make ./.current/test; done > /dev/null
```

```
real    0m17.227s
user    0m15.793s
sys     0m1.051s
```
```
export CPLUSPLUS=clang++ && time for i in `seq 10`; do touch test.cc ; make ./.current/test; done > /dev/null`
```
```
real    0m14.902s
user    0m14.099s
sys     0m0.612s
```

**Past this PR**

```
export CPLUSPLUS=g++ && time for i in `seq 10`; do touch test.cc ; make ./.current/test; done > /dev/null
```

```
real    0m37.726s
user    0m35.422s
sys     0m1.796s
```

```
export CPLUSPLUS=clang++ && time for i in `seq 10`; do touch test.cc ; make ./.current/test; done > /dev/null
```

```
real    0m19.694s
user    0m18.567s
sys     0m0.800s
```

Nicht gut. :-( And the old code even used to give some compilation warnings.

mzhurovich:
@dkorolev :(
And I see the error:
```
test.cc:107:15: error: non-type template argument evaluates to 18446744073709551615, which cannot be narrowed to type 'int' [-Wc++11-narrowing]
static_assert(IS_VALID_CURRENT_STRUCT(Empty), "`Empty` must pass `IS_VALID_CURRENT_STRUCT` check.");
```

dkorolev:
The error has been fixed. The cleanest path forward is to be found.

PR #506, opened 2016-05-22
Renamed `current_build.h` into `current.build`.

dkorolev:
Hi Max,

Chinese coding. I figured if we rename `current_build.h` into `current.build`, the double-remake trouble would go away. After all, it's the time of the last build of a specific binary *as of when the code related to this binary has changed*, not the last `make` invocation.

Btw, this change, if you like it, makes most of the logic in `scripts/gen-current-build.sh` obsolete. :-)

Thanks,
Dima


mzhurovich:
@dkorolev 👍 
What do you think about `current_build.inc`?

dkorolev:
If .inc is for include, i don't think it sends the right message. Some
.header or .autogen or .h_autogen would probably be cleaner.

Or we could read the docs for Makefiles and exclude this current_build.h
header from a *.h dependencies list.

On Sunday, May 22, 2016, Maxim Zhurovich <notifications@github.com> wrote:

> @dkorolev <https://github.com/dkorolev> 👍
> What do you think about current_build.inc?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/C5T/Current/pull/506#issuecomment-220855723>
>


-- 
Dima

dkorolev:
@mzhurovich PTAL!

PR #517, opened 2016-05-30
Custom storage can be used in Karl.

mzhurovich:
@dkorolev Moar metaprogramming!

### DIFF
--- a/Karl/test.cc
+++ b/Karl/test.cc
@@ -72,8 +72,9 @@ DEFINE_bool(karl_run_test_forever, false, "Set to `true` to run the Karl test fo
 
 DEFINE_bool(karl_overwrite_golden_files, false, "Set to true to have SVG golden files created/overwritten.");
 
-using unittest_karl_t =
-    current::karl::GenericKarl<current::karl::default_user_status::status, karl_unittest::is_prime>;
+using unittest_karl_t = current::karl::GenericKarl<current::karl::UseOwnStorage,
+                                                   current::karl::default_user_status::status,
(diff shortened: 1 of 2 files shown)


Karl/test.cc
```diff
@@ -905,6 +906,82 @@ TEST(Karl, EndToEndTest) {
...
+    const custom_karl_t karl(storage, UnittestKarlParameters());
+    const current::karl::Locator karl_locator(Printf("http://localhost:%d/", FLAGS_karl_test_keepalives_port));
+    const karl_unittest::ServiceGenerator generator(
+        FLAGS_karl_generator_test_port, std::chrono::microseconds(1000), karl_locator);
+    generator_codename = generator.ClaireCodename();
+
+    const auto result =
+        karl.InternalExposeStorage().ReadOnlyTransaction([&](ImmutableFields<custom_storage_t> fields) {
+          ASSERT_TRUE(Exists(fields.claires[generator_codename]));
+          EXPECT_EQ(current::karl::ClaireRegisteredState::Active,
```
dkorolev: These would be unreadable in the console output if the test fails, right?

dkorolev:
LGTM++!

Karl/test.cc
mzhurovich: Readable enough IMHO:
```
Value of: Value(fields.claires[generator_codename]).registered_state
  Actual: 1-byte object <01>
Expected: current::karl::ClaireRegisteredState::Active
Which is: 1-byte object <00>
```

dkorolev: Merging!

PR #522, opened 2016-06-02
Karl's callbacks.

dkorolev:
Hi Max,

I don't like two `Exists` calls for the `Optional<Variant<TS...>>` call, but should do the job for now.

Thanks,
Dima


mzhurovich:
@dkorolev This new test has revealed some multithreading issues in Claire. Fix is on the way.

PR #524, opened 2016-06-07
Added constructors for iOSDeviceInfo to fix Midichlorians' compilation

grixa:
CC @dkorolev @mzhurovich

### DIFF
--- a/Midichlorians/Client/iOS/MidichloriansImpl.mm
+++ b/Midichlorians/Client/iOS/MidichloriansImpl.mm
@@ -321,7 +321,7 @@ using namespace current::midichlorians::ios::impl;
     }
     
     if (!info.empty()) {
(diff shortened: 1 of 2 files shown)


dkorolev:
:+1:

Midichlorians/Client/iOS/MidichloriansImpl.mm
```diff
@@ -321,7 +321,7 @@ + (void)setup:(NSString *)serverUrl withLaunchOptions:(NSDictionary *)options {
     }
     
     if (!info.empty()) {
-        [MidichloriansImpl emit:iOSDeviceInfo(info)];
+        [MidichloriansImpl emit:iOSDeviceInfo(std::move(info))];
```
mzhurovich: Why `std::move` if there is no move-constructor in the `iOSDeviceInfo`?

PR #528, opened 2016-06-11
Type evolved storage.

dkorolev:
Hi Max!

TL;DR:

```
 942 TEST(TypeEvolutionTest, StorageInnerTypes) {
 943   // Test the plain `Name` type goes through.
 944   typename SchemaOriginalStorage::Name foo;
 945   foo.first = "Ayn";
 946   foo.last = "Rand";
 947   {
 948     typename SchemaOriginalStorage::Name bar;
 949     current::type_evolution::Evolve<SchemaOriginalStorage, typename SchemaOriginalStorage::Name>::template Go<
 950         SchemaOriginalStorage>(foo, bar);
 951     EXPECT_EQ("{\"first\":\"Ayn\",\"last\":\"Rand\"}", JSON(bar));
 952   }
 953   {
 954     typename SchemaOriginalStorage::Name baz;
 955     current::type_evolution::Evolve<SchemaOriginalStorage,
 956                                     typename SchemaOriginalStorage::Name,
 957                                     current::type_evolution::OriginalStorageToOriginalStorageEvolutor>::
 958         template Go<SchemaOriginalStorage>(foo, baz);
 959     EXPECT_EQ("{\"first\":\"AYN\",\"last\":\"RAND\"}", JSON(baz));
 960   }
 961   {
 962     typename SchemaModifiedStorage::Name meh;
 963     current::type_evolution::Evolve<SchemaOriginalStorage,
 964                                     typename SchemaOriginalStorage::Name,
 965                                     current::type_evolution::OriginalStorageToModifiedStorageEvolutor>::
 966         template Go<SchemaModifiedStorage>(foo, meh);
 967     EXPECT_EQ("{\"full\":\"Rand, A\"}", JSON(meh));
 968   }
 969 }
```

```
$ cat golden/storage_pre_evolution.txt 
{"index":0,"us":1}	{"meta":{"timestamp":0,"fields":{}},"mutations":[{"PersistedUserUpdated":{"data":{"first":"KARL","last":"MARX","key":"karl"}},"":"T9201121062548426382"}]}
{"index":1,"us":3}	{"meta":{"timestamp":2,"fields":{}},"mutations":[{"PersistedUserUpdated":{"data":{"first":"DIMA","last":"KOROLEV","key":"dima"}},"":"T9201121062548426382"},{"PersistedUserUpdated":{"data":{"first":"MAX","last":"ZHUROVICH","key":"max"}},"":"T9201121062548426382"}]}

$ cat golden/storage_post_evolution.txt 
[...]

PR #529, opened 2016-06-13
Hardening type evolution.

dkorolev:
(Note: This PR contains all the changes from https://github.com/C5T/Current/pull/528).

Hi @mzhurovich , CC @grixa.

**TL;DR:**

1. The implementation of our type evolution is officially robust to the degree it can handle CTFO's ID migration. Specifically, I've added and tested `CURRENT_ENUM`, `std::pair<>`, and `Optional<>` support.
2. Sherlock streams, and, by extension, our Storage subsystem, not only expose all their types in a format ready for type evolution, but also "create" a `CURRENT_NAMESPACE` (actually, a `CURRENT_DERIVED_NAMESPACE`) of a user-defined name, exposing the `TopLevelTransaction` type. (*) Thus, as long as while initializing the `Storage` from the user code (CTFO++) we keep this namespace different from version to version, most of data&schema migration would would be about `#include`-ing that one `curl`-ed "header".

But of course:

```
[==========] 376 tests from 57 test cases ran. (5857 ms total)
[  PASSED  ] 376 tests.
```

PTAL && thanks!
Dima

(*) I've realized too late `TopLevelTransaction` makes little sense for the Stream itself, it's a Storage-centric name. Accepting suggestions.


Sherlock/sherlock.h
```diff
@@ -483,7 +483,10 @@ class StreamImpl {
     explicit FillPerLanguageSchema(SherlockSchema& schema) : schema_ref(schema) {}
     template <current::reflection::Language language>
     void PerLanguage() {
-      schema_ref.language[current::ToString(language)] = schema_ref.type_schema.Describe<language>();
+      current::reflection::NamespaceToExpose expose("Sherlock");
+      expose.template AddType<entry_t>("TopLevelTransaction");
```
mzhurovich: `StreamEntry(Type)`?

mzhurovich:
@dkorolev I doubt, I can do something more constructive at this point than just merge this pull request :) This is really huge step and I really appreciate your effort to convert our ideas into this solid prototype. Так что, помолясь, применимъ!

dkorolev:
Yay! Thanks @mzhurovich 

@grixa -- This is a pretty big change, perhaps now is a good time to re-clone a fresh repo altogether on your end, confirm the top-level `make test` and `make check` pass, and take it from there? Thanks!

PR #532, opened 2016-06-14
Allow only GET method for the Nginx-proxied Claire status pages.

mzhurovich:


mzhurovich:
[Reviewed by @dkorolev]

PR #537, opened 2016-06-28
Update MakefileImpl

trueaywee:
"find -perm +111" doesn't work at least in find 4.7.0 and +mode is considered deprecated. Manual suggests using /mode instead (and it works).

### DIFF
--- a/scripts/MakefileImpl
+++ b/scripts/MakefileImpl
@@ -49,7 +49,7 @@ default: all
 	if [ -f test.cc ] ; then \
 		make test ;\
 	else \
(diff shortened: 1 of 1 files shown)


mzhurovich:
This won't work on OS X: `find: -perm: /111: illegal mode string`.
Need more cross-platform way.

dkorolev:
@trueaywee welcome to the dark side!

Appreciate you want it to work on your machine out of the box. Conditional invocation of `find` would probably be best here.

Thanks!
Dima

PR #541, opened 2016-06-30
`FromString`: universal references work fine.

mzhurovich:
I've removed additional template parameters for static functions, as I proved them to be redundant.
TL;DR: when we call `FromString` template deduction kicks in and we deal with universal references, so at the moment we call `FromStringImpl<INPUT....>`, `INPUT` argument itself has already the proper type (const or rvalue ref). Thus, coupled with `std::forward` and reference collapsing rules, our `static` functions do their job absolutely fine without extra templatization.

### DIFF
--- a/Bricks/strings/test.cc
+++ b/Bricks/strings/test.cc
@@ -140,8 +140,8 @@ TEST(Util, FromString) {
   EXPECT_EQ(65535, static_cast<int>(current::FromString<uint16_t>("65535")));
 
   double tmp_double;
-  EXPECT_EQ(0.1, current::FromString("0.1", tmp_double));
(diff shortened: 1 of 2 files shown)


Bricks/strings/test.cc
```diff
@@ -140,8 +140,8 @@ TEST(Util, FromString) {
   EXPECT_EQ(65535, static_cast<int>(current::FromString<uint16_t>("65535")));
 
   double tmp_double;
-  EXPECT_EQ(0.1, current::FromString("0.1", tmp_double));
-  EXPECT_EQ(0.1, tmp_double);
+  EXPECT_EQ(0.25, current::FromString("0.25", tmp_double));
```
dkorolev: :+1:

PR #563, opened 2016-07-18
Used input arguments for sherlock schema top level name instead of hardcoded value.

grixa:
CC @dkorolev @mzhurovich


dkorolev:
:+1:, thank you! And glad to know Travis helped.

PR #564, opened 2016-07-20
Fixed requesting Sherlock schema using "/schema.<language>".

grixa:
СС @dkorolev @mzhurovich

### DIFF
--- a/Storage/test.cc
+++ b/Storage/test.cc
@@ -1613,6 +1613,15 @@ TEST(TransactionalStorage, ReplicationViaHTTP) {
   const auto master_storage_file_remover = current::FileSystem::ScopedRmFile(master_storage_file_name);
   Storage master_storage(master_storage_file_name);
   master_storage.ExposeRawLogViaHTTP(FLAGS_transactional_storage_test_port, "/raw_log");
+  const std::string base_url = Printf("http://localhost:%d/raw_log", FLAGS_transactional_storage_test_port);
+
+  // Check that the Storage is available via HTTP.
+  {
+    const auto result = HTTP(GET(base_url + "?schema=json"));
+    EXPECT_EQ(200, static_cast<int>(result.code));
+    const auto result2 = HTTP(GET(base_url + "/schema.json"));
+    EXPECT_EQ(200, static_cast<int>(result2.code));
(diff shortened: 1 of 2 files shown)


dkorolev:
@grixa perhaps add a simple test?

grixa:
@dkorolev There are such tests in the [https://github.com/C5T/Current/blob/master/Sherlock/test.cc](url), but they are not use ExposeRawLogViaHTTP method (as it is a part of Storage). Should I add a similar test to the [https://github.com/C5T/Current/blob/master/Storage/test.cc](url) then?

dkorolev:
Your links don't open by themselves, I had to copy-paste them. :-)

Something as simple as https://github.com/C5T/Current/blob/master/Sherlock/test.cc#L459-L463; no need to compare the body to the golden body, just confirm it returns the right HTTP code after your change, and the wrong HTTP code before it.

Storage/test.cc
```diff
@@ -1613,6 +1613,15 @@ TEST(TransactionalStorage, ReplicationViaHTTP) {
...
   Storage master_storage(master_storage_file_name);
   master_storage.ExposeRawLogViaHTTP(FLAGS_transactional_storage_test_port, "/raw_log");
+  const std::string base_url = Printf("http://localhost:%d/raw_log", FLAGS_transactional_storage_test_port);
+
+  // Check that the Storage is available via HTTP.
+  {
+    const auto result = HTTP(GET(base_url + "?schema=json"));
+    EXPECT_EQ(200, static_cast<int>(result.code));
+    const auto result2 = HTTP(GET(base_url + "/schema.json"));
+    EXPECT_EQ(200, static_cast<int>(result2.code));
```
dkorolev: :+1: could you also add an EXPECT_EQ that the `.body` responses of two GET requests match?

grixa:
Done. Sorry for the bad links.

Storage/test.cc
grixa: Yes, good point

dkorolev:
LGTM++, thanks @grixa! See my one-line comment about another `EXPECT_EQ`, and let's merge this thing.

PR #565, opened 2016-07-20
RapidJSON is no longer killing us.

dkorolev:
(Deliberately behind Master -- D.K.)


dkorolev:
Actually, no point in merging this one. Let's do it right with The Master tomorrow. Closing.

PR #567, opened 2016-07-22
RapidJSON is no longer killing us (up to date).

dkorolev:
Replacing https://github.com/C5T/Current/pull/565 which was a hotfix.


mzhurovich:
@dkorolev LGTM!

PR #575, opened 2016-07-24
Remote stream subscriber and replicator

grixa:
CC @dkorolev @mzhurovich

### DIFF
--- a/Sherlock/exceptions.h
+++ b/Sherlock/exceptions.h
@@ -47,6 +47,30 @@ struct PublisherAlreadyOwnedException : SherlockException {
   using SherlockException::SherlockException;
 };
 
+struct PublisherDoesNotExistException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
+struct PublisherIsUsedException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
+struct SubscriptionIsActiveException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
+struct SubscriptionIsNotActiveException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
+struct RemoteStreamDoesNotRespondException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
+struct RemoteStreamSchemaInvalidException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
 struct StreamInGracefulShutdownException : InGracefulShutdownException {
   using InGracefulShutdownException::InGracefulShutdownException;
 };
--- a/Sherlock/exceptions.h
+++ b/Sherlock/exceptions.h
@@ -47,22 +47,6 @@ struct PublisherAlreadyOwnedException : SherlockException {
   using SherlockException::SherlockException;
 };
 
-struct PublisherDoesNotExistException : SherlockException {
-  using SherlockException::SherlockException;
-};
-
-struct PublisherIsUsedException : SherlockException {
-  using SherlockException::SherlockException;
-};
-
-struct SubscriptionIsActiveException : SherlockException {
-  using SherlockException::SherlockException;
-};
-
-struct SubscriptionIsNotActiveException : SherlockException {
-  using SherlockException::SherlockException;
-};
-
 struct RemoteStreamDoesNotRespondException : SherlockException {
   using SherlockException::SherlockException;
 };
--- a/Sherlock/exceptions.h
+++ b/Sherlock/exceptions.h
@@ -67,7 +67,7 @@ struct RemoteStreamDoesNotRespondException : SherlockException {
   using SherlockException::SherlockException;
 };
 
-struct RemoteStreamSchemaInvalidException : SherlockException {
+struct RemoteStreamInvalidSchemaException : SherlockException {
   using SherlockException::SherlockException;
 };
 
--- a/Sherlock/exceptions.h
+++ b/Sherlock/exceptions.h
@@ -71,6 +71,10 @@ struct RemoteStreamInvalidSchemaException : SherlockException {
   using SherlockException::SherlockException;
 };
 
+struct StreamTerminatedBySubscriber : SherlockException {
+  using SherlockException::SherlockException;
+};
+
 struct StreamInGracefulShutdownException : InGracefulShutdownException {
   using InGracefulShutdownException::InGracefulShutdownException;
 };
--- a/Sherlock/replicator.h
+++ b/Sherlock/replicator.h
@@ -229,32 +229,40 @@ class SubscribableRemoteStream final {
 };
 
 template <typename T_STREAM>
-struct RemoteStreamReplicatorImpl {
+struct StreamReplicatorImpl {
   using EntryResponse = current::ss::EntryResponse;
   using TerminationResponse = current::ss::TerminationResponse;
   using entry_t = typename T_STREAM::entry_t;
   using publisher_t = typename T_STREAM::publisher_t;
 
-  T_STREAM& stream_;
-  std::unique_ptr<publisher_t> publisher_;
-
-  RemoteStreamReplicatorImpl(T_STREAM& stream) : stream_(stream) { stream.MovePublisherTo(*this); }
-  ~RemoteStreamReplicatorImpl() { stream_.AcquirePublisher(std::move(publisher_)); }
+  StreamReplicatorImpl(T_STREAM& stream) : stream_(stream) { stream.MovePublisherTo(*this); }
+  ~StreamReplicatorImpl() { stream_.AcquirePublisher(std::move(publisher_)); }
 
   void AcceptPublisher(std::unique_ptr<publisher_t> publisher) { publisher_ = std::move(publisher); }
 
   EntryResponse operator()(entry_t&& entry, idxts_t current, idxts_t) {
+    assert(publisher_);
     publisher_->Publish(std::move(entry), current.us);
+    return EntryResponse::More;
+  }
+
+  EntryResponse operator()(const entry_t& entry, idxts_t current, idxts_t) {
+    assert(publisher_);
+    publisher_->Publish(entry, current.us);
     return EntryResponse::More;
   }
 
   EntryResponse EntryResponseIfNoMorePassTypeFilter() const { return EntryResponse::More; }
   TerminationResponse Terminate() const { return TerminationResponse::Terminate; }
+
+ private:
+  T_STREAM& stream_;
+  std::unique_ptr<publisher_t> publisher_;
 };
 
 template <typename T_STREAM>
-using RemoteStreamReplicator =
-    current::ss::StreamSubscriber<RemoteStreamReplicatorImpl<T_STREAM>, typename T_STREAM::entry_t>;
+using StreamReplicator =
+    current::ss::StreamSubscriber<StreamReplicatorImpl<T_STREAM>, typename T_STREAM::entry_t>;
 
 }  // namespace sherlock
 }  // namespace current
--- a/Sherlock/replicator.h
+++ b/Sherlock/replicator.h
@@ -151,7 +151,8 @@ class SubscribableRemoteStream final {
       std::unique_lock<std::mutex> lock(mutex_);
       event_.wait(lock, [this]() { return !subscription_id_.empty(); });
       external_stop_ = true;
-      const std::string terminate_url = remote_stream_.ObjectAccessorDespitePossiblyDestructing().url_ + "?terminate=" + subscription_id_;
+      const std::string terminate_url =
+          remote_stream_.ObjectAccessorDespitePossiblyDestructing().url_ + "?terminate=" + subscription_id_;
       try {
         const auto result = HTTP(GET(terminate_url));
       } catch (current::net::NetworkException&) {
@@ -180,7 +181,6 @@ class SubscribableRemoteStream final {
       : stream_(remote_stream_url),
         schema_(
             Value<reflection::ReflectedTypeBase>(reflection::Reflector().ReflectType<T_STREAM_ENTRY>()).type_id,
-            reflection::CurrentTypeName<T_STREAM_ENTRY>(),
             sherlock::constants::kDefaultTopLevelName,
             sherlock::constants::kDefaultNamespaceName) {
     CheckRemoteSchema();
@@ -192,12 +192,11 @@ class SubscribableRemoteStream final {
       : stream_(remote_stream_url),
         schema_(
             Value<reflection::ReflectedTypeBase>(reflection::Reflector().ReflectType<T_STREAM_ENTRY>()).type_id,
-            reflection::CurrentTypeName<T_STREAM_ENTRY>(),
             top_level_name,
             namespace_name) {
     CheckRemoteSchema();
   }
-  ~SubscribableRemoteStream(){}
+  ~SubscribableRemoteStream() {}
 
   template <typename F>
   RemoteSubscriberScope<F, T_STREAM_ENTRY> Subscribe(F& subscriber,
--- a/Sherlock/replicator.h
+++ b/Sherlock/replicator.h
@@ -22,8 +22,8 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 *******************************************************************************/
 
-#ifndef REMOTE_STREAM_REPLICATOR_H
-#define REMOTE_STREAM_REPLICATOR_H
+#ifndef CURRENT_SHERLOCK_REPLICATOR_H
+#define CURRENT_SHERLOCK_REPLICATOR_H
 
 #include <functional>
 #include <string>
@@ -43,9 +43,11 @@ SOFTWARE.
 namespace current {
 namespace sherlock {
 
-template <typename T_STREAM_ENTRY>
+template <typename STREAM_ENTRY>
 class SubscribableRemoteStream final {
  public:
+  using entry_t = STREAM_ENTRY;
+
   struct RemoteStream final {
     RemoteStream(const std::string& url) : url_(url) {}
 
@@ -54,25 +56,6 @@ class SubscribableRemoteStream final {
 
   template <typename F, typename TYPE_SUBSCRIBED_TO>
   class RemoteSubscriberThread final : public current::sherlock::SubscriberScope::SubscriberThread {
-   private:
-    bool valid_;
-    std::function<void()> done_callback_;
-    std::atomic_bool external_stop_;
-    std::atomic_bool internal_stop_;
-    ScopeOwnedBySomeoneElse<RemoteStream> remote_stream_;
-    F& subscriber_;
-    uint64_t index_;
-    std::thread thread_;
-    std::mutex mutex_;
-    std::condition_variable event_;
-    std::string subscription_id_;
-
-    RemoteSubscriberThread() = delete;
-    RemoteSubscriberThread(const RemoteSubscriberThread&) = delete;
-    RemoteSubscriberThread(RemoteSubscriberThread&&) = delete;
-    void operator=(const RemoteSubscriberThread&) = delete;
-    void operator=(RemoteSubscriberThread&&) = delete;
-
    public:
     RemoteSubscriberThread(ScopeOwned<RemoteStream>& remote_stream,
                            F& subscriber,
@@ -82,15 +65,10 @@ class SubscribableRemoteStream final {
           done_callback_(done_callback),
           external_stop_(false),
           internal_stop_(false),
-          remote_stream_(remote_stream,
-                         [this]() {
-                           TerminateSubscription();
-                           thread_.join();
-                         }),
+          remote_stream_(remote_stream, [this]() { TerminateSubscription(); }),
           subscriber_(subscriber),
           index_(start_idx),
           thread_([this]() { Thread(); }) {
-      static_assert(current::ss::IsEntrySubscriber<F, TYPE_SUBSCRIBED_TO>::value, "");
       valid_ = true;
     }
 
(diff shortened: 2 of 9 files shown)


Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+#ifndef REMOTE_STREAM_REPLICATOR_H
```
mzhurovich: s/REMOTE_STREAM_REPLICATOR_H/CURRENT_SHERLOCK_REPLICATOR_H/

Sherlock/exceptions.h
```diff
@@ -47,6 +47,30 @@ struct PublisherAlreadyOwnedException : SherlockException {
...
+
+struct SubscriptionIsNotActiveException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
+struct RemoteStreamDoesNotRespondException : SherlockException {
+  using SherlockException::SherlockException;
+};
+
+struct RemoteStreamSchemaInvalidException : SherlockException {
```
dkorolev: s/`SchemaInvalid`/`InvalidSchema`/g?

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+          internal_stop_(false),
+          remote_stream_(remote_stream,
+                         [this]() {
+                           TerminateSubscription();
+                           thread_.join();
+                         }),
+          subscriber_(subscriber),
+          index_(start_idx),
+          thread_([this]() { Thread(); }) {
+      static_assert(current::ss::IsEntrySubscriber<F, TYPE_SUBSCRIBED_TO>::value, "");
```
dkorolev: This is a class invariant, not constructor invariant -- best to hoist it.

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+      bool terminate_sent = false;
+      while (!internal_stop_) {
+        if (!terminate_sent && external_stop_) {
+          terminate_sent = true;
+          if (subscriber_.Terminate() != ss::TerminationResponse::Wait) {
+            return;
+          }
+        }
+        try {
+          HTTP(ChunkedGET(
```
dkorolev: Probably better to `CheckRemoteSchema()` here, prior to each `ChunkedGET` -- the source of data may restart.

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+      internal_stop_ =
+          (subscriber_(std::move(entry), idx, idxts_t() /*unused "last" arg*/) == ss::EntryResponse::Done);
+      std::unique_lock<std::mutex> lock(mutex_);
+      ++index_;
+      event_.notify_one();
+    }
+
+    void TerminateSubscription() {
+      std::unique_lock<std::mutex> lock(mutex_);
+      event_.wait(lock, [this]() { return !subscription_id_.empty(); });
```
dkorolev: `|| internal_stop_` in the condition?

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+      external_stop_ = true;
+      const std::string terminate_url =
```
dkorolev: And hide this into an `if (!subscription_id_.empty())`.

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+
+    void OnHeader(const std::string& header, const std::string& value) {
+      if (header == "X-Current-Stream-Subscription-Id") {
+        std::unique_lock<std::mutex> lock(mutex_);
+        subscription_id_ = value;
+        event_.notify_one();
+      }
+    }
+
+    void OnChunk(const std::string& chunk) {
```
mzhurovich: This implementation relies on the fact that each chunk is exactly one log entry, which in general is not true. I'd suggest to at least add a comment about this.

Sherlock/replicator.h
mzhurovich: What happens to the `!subscription_id_.empty()` logic if the connection has been terminated abnormally?

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+        return;
+      }
+
+      const auto split = current::strings::Split(chunk, '\t');
+      assert(split.size() == 2u);
+      const auto idx = ParseJSON<idxts_t>(split[0]);
+      assert(idx.index == index_);
+      auto entry = ParseJSON<TYPE_SUBSCRIBED_TO>(split[1]);
+      internal_stop_ =
+          (subscriber_(std::move(entry), idx, idxts_t() /*unused "last" arg*/) == ss::EntryResponse::Done);
```
dkorolev: `/* */` comments are generally undesirable -- perhaps just `const idxts_t unused_idxts;` up there?

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+      std::unique_lock<std::mutex> lock(mutex_);
+      ++index_;
+      event_.notify_one();
```
mzhurovich: What's the purpose of this notification?

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+                                    const std::string& top_level_name,
+                                    const std::string& namespace_name)
+      : stream_(remote_stream_url),
+        schema_(
+            Value<reflection::ReflectedTypeBase>(reflection::Reflector().ReflectType<T_STREAM_ENTRY>()).type_id,
+            top_level_name,
+            namespace_name) {
+    CheckRemoteSchema();
+  }
+  ~SubscribableRemoteStream() {}
```
dkorolev: This must have had some value :)

Sherlock/replicator.h
grixa: Right. By the way, if the schema is changed, there will be a busy loop, is it OK?

dkorolev: A comment would do, for now. :-)

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+    try {
+      return RemoteSubscriberScope<F, T_STREAM_ENTRY>(stream_, subscriber, start_idx, done_callback);
+    } catch (const current::sync::InDestructingModeException&) {
+      CURRENT_THROW(StreamInGracefulShutdownException());
+    }
+  }
+
+ private:
+  void CheckRemoteSchema() {
+    const auto response = HTTP(GET(stream_.ObjectAccessorDespitePossiblyDestructing().url_ + "/schema.simple"));
```
dkorolev: Design suggestion for outside this pull request. Perhaps it's best to have a single "pre-subscribe" endpoint, would would:
1) Accept the desired schema as a header.
2) If the actual schema matches the desired one, pre-generate a subscription ID and return it, so that the actual subscription that follows next can request using the pre-generated ID instead of waiting for one in the header.
3) If the actual schema does not match, return both the 3-tuple schema signature and the actual schema, so that if we need to know what exactly is the difference between the schemas, the data to diff is right there.

Sherlock/replicator.h
grixa: Then may be we should check `subscriber_thread_done_` here. But it also won't help if the thread would continuously fail to connect, for example.

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+    if (static_cast<int>(response.code) == 200) {
+      SubscribableSherlockSchema remote_schema =
```
dkorolev: Don't these two fit one line?

Sherlock/replicator.h
dkorolev: Yes, see the suggestion to pre-generate the subscription ID on the schema-validating request.

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+
+template <typename T_STREAM>
+struct StreamReplicatorImpl {
+  using EntryResponse = current::ss::EntryResponse;
+  using TerminationResponse = current::ss::TerminationResponse;
+  using entry_t = typename T_STREAM::entry_t;
+  using publisher_t = typename T_STREAM::publisher_t;
+
+  StreamReplicatorImpl(T_STREAM& stream) : stream_(stream) { stream.MovePublisherTo(*this); }
+  ~StreamReplicatorImpl() { stream_.AcquirePublisher(std::move(publisher_)); }
```
dkorolev: `virtual` ?

Sherlock/replicator.h
```diff
@@ -0,0 +1,285 @@
...
+    assert(publisher_);
+    publisher_->Publish(entry, current.us);
+    return EntryResponse::More;
+  }
+
+  EntryResponse EntryResponseIfNoMorePassTypeFilter() const { return EntryResponse::More; }
+  TerminationResponse Terminate() const { return TerminationResponse::Terminate; }
+
+ private:
+  T_STREAM& stream_;
```
dkorolev: Nit: `using stream_t` above as well, and `stream_t` here.

Also, upon introducing the `type_alias_t` semantics, we've moved from `T_STREAM` to just `STREAM` in template parameter names. Thanks!

Storage/test.cc
```diff
@@ -1722,57 +1724,35 @@ TEST(TransactionalStorage, ReplicationViaHTTP) {
...
-  while (std::getline(body, line)) {
-    const size_t tab_pos = line.find('\t');
-    ASSERT_FALSE(tab_pos == std::string::npos);
-    const auto idx_ts = ParseJSON<idxts_t>(line.substr(0, tab_pos));
-    EXPECT_EQ(expected_index, idx_ts.index);
-    auto transaction = ParseJSON<Storage::transaction_t>(line.substr(tab_pos + 1));
-    ASSERT_NO_THROW(stream_publisher_owner.publisher_->Publish(std::move(transaction), idx_ts.us));
-    ++expected_index;
+  {
+    auto subscriber_scope = remote_stream.Subscribe(*replicator);
```
dkorolev: `const`

Storage/test.cc
```diff
@@ -1722,57 +1724,35 @@ TEST(TransactionalStorage, ReplicationViaHTTP) {
...
+    while (replicated_stream.InternalExposePersister().Size() < 2u) {
+      std::this_thread::yield();
```
dkorolev: :+1:

grixa:
It's not the final change yet

grixa:
Almost done

mzhurovich:
Looks good enough to merge, @dkorolev ?

grixa:
Two more changes left: break ChunkedGET from itself using exception and use WaitableAtomic instead of condition_variable

dkorolev:
Got it. Staying tuned!

grixa:
Done.

Sherlock/exceptions.h
```diff
@@ -47,6 +47,34 @@ struct PublisherAlreadyOwnedException : SherlockException {
   using SherlockException::SherlockException;
 };
 
+struct PublisherDoesNotExistException : SherlockException {
```
dkorolev: Unused?

Sherlock/exceptions.h
```diff
@@ -47,6 +47,34 @@ struct PublisherAlreadyOwnedException : SherlockException {
...
+  using SherlockException::SherlockException;
+};
+
+struct PublisherIsUsedException : SherlockException {
```
dkorolev: Unused?

Sherlock/exceptions.h
```diff
@@ -47,6 +47,34 @@ struct PublisherAlreadyOwnedException : SherlockException {
...
+  using SherlockException::SherlockException;
+};
+
+struct SubscriptionIsActiveException : SherlockException {
```
dkorolev: Unused too. What am I missing?

Sherlock/exceptions.h
```diff
@@ -47,6 +47,34 @@ struct PublisherAlreadyOwnedException : SherlockException {
...
+  using SherlockException::SherlockException;
+};
+
+struct SubscriptionIsNotActiveException : SherlockException {
```
dkorolev: Unused :)

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+   public:
+    RemoteStream(const std::string& url, const std::string& top_level_name, const std::string& namespace_name)
+        : url_(url),
+          schema_(Value<reflection::ReflectedTypeBase>(reflection::Reflector().ReflectType<entry_t>()).type_id,
+                  top_level_name,
+                  namespace_name) {}
+
+    void CheckSchema() const {
+      const auto response = HTTP(GET(url_ + "/schema.simple"));
+      if (static_cast<int>(response.code) == 200) {
```
dkorolev: `static_cast<int>` is useful for unit tests to dump the "wrong" value as an integer. In real code, if (response.code == HTTPResponseCode.OK) is cleaner.

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+      } else {
+        CURRENT_THROW(RemoteStreamDoesNotRespondException());
+      }
+    }
+
+    const std::string GetURLToSubscribe(uint64_t index) const {
+      return url_ + "?i=" + current::ToString(index);
+    }
+
+    const std::string GetURLToTerminate(const std::string& subscription_id) {
```
dkorolev: `) const {` ?

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+          remote_stream_(remote_stream, [this]() { TerminateSubscription(); }),
+          subscriber_(subscriber),
+          index_(start_idx),
+          thread_([this]() { Thread(); }) {
+      valid_ = true;
+    }
+
+    ~RemoteSubscriberThread() {
+      if (valid_) {
+        if (!subscriber_thread_done_) {
```
dkorolev: This needs a mutex, here and where it's accessed.

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+          TerminateSubscription();
+        }
+        thread_.join();
```
dkorolev: Effectively, `!subscriber_thread_done_` is an invariant for "thread has not been joined yet", which is `thread_.joinable()`. Might be worth a comment or an assert.

Sherlock/replicator.h
mzhurovich: What's the point of returning `const`?

Sherlock/replicator.h
dkorolev: (Sorry for being picky here, just spent the morning making `Sherlock` and `Karl` tests stable. Thanks for understanding! :-) )

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+      } else {
+        if (done_callback_) {
+          done_callback_();
+        }
+      }
+    }
+
+   private:
+    static_assert(current::ss::IsEntrySubscriber<F, TYPE_SUBSCRIBED_TO>::value, "");
```
dkorolev: Nit: a `static_assert` on a class-level invariant better be the first line of the body of the lass.

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+          }
+        }
+        try {
+          bare_stream.CheckSchema();
+          HTTP(ChunkedGET(
+              bare_stream.GetURLToSubscribe(index_),
+              [this](const std::string& header, const std::string& value) { OnHeader(header, value); },
+              [this](const std::string& chunk_body) { OnChunk(chunk_body); },
+              [this]() {}));
+        } catch (StreamTerminatedBySubscriber&) {
```
dkorolev: `OnHeader()` here would save `subscription_id_`. Won't it make sense to erase it after `HTTP(ChunkedGET(...))` returns, as its value is no longer valid? (Or perhaps combine `subscription_id_` and `subscription_established_` in one `WaitableAtomic<>` would be a good idea?)

Sherlock/exceptions.h
grixa: Ah, right. These are reminders of an old version of the replicator.

Sherlock/replicator.h
```diff
@@ -155,12 +155,12 @@ class SubscribableRemoteStream final {
...
         std::unique_lock<std::mutex> lock(mutex_);
         subscription_id_ = value;
-        event_.notify_one();
+        *subscription_established_.MutableScopedAccessor() = true;
       }
     }
 
     void OnChunk(const std::string& chunk) {
-      if (external_stop_ ) {
+      if (external_stop_) {
```
dkorolev: +1 to some `_requested` in the name.

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+
+    void OnChunk(const std::string& chunk) {
+      if (external_stop_) {
+        return;
+      }
+
+      // Expected one entire entry in each chunk,
+      // won't work in case of partial or multiple entries per chunk
+      const auto split = current::strings::Split(chunk, '\t');
+      assert(split.size() == 2u);
```
dkorolev: These should certainly not be `assert`-s -- why kill the whole binary when all it warrants is just HTTP connection re-initiation? After all, a TCP connection could be lost in the meantime, etc., etc. -- better be safe.

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+      const auto idxts = ParseJSON<idxts_t>(split[0]);
+      assert(idxts.index == index_);
+      auto entry = ParseJSON<TYPE_SUBSCRIBED_TO>(split[1]);
+      const idxts_t unused_idxts;
```
dkorolev: Makes me wonder if it's default-initialized correctly -- I know it is, but looks unsafe. Why not make it a `const` member variable and initialize in constructor?

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+    void TerminateSubscription() {
+      std::unique_lock<std::mutex> lock(mutex_);
+      subscription_established_.Wait(
+          [this](bool established) { return established || subscriber_thread_done_; });
+      external_stop_ = true;
+      if (!subscription_id_.empty()) {
+        const std::string terminate_url =
+            remote_stream_.ObjectAccessorDespitePossiblyDestructing().GetURLToTerminate(subscription_id_);
+        try {
+          const auto result = HTTP(GET(terminate_url));
```
dkorolev: lonely, lonely `result`, no one needs it :(

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+        } catch (current::net::NetworkException&) {
```
dkorolev: `current::Exception` instead of `current::net::NetworkException`?

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+        }
+      }
+    }
+
+   private:
+    bool valid_;
+    std::function<void()> done_callback_;
```
dkorolev: looks `const` to me

Sherlock/replicator.h
```diff
@@ -0,0 +1,296 @@
...
+    using base_t = current::sherlock::SubscriberScope;
+
+   public:
+    using subscriber_thread_t = RemoteSubscriberThread<F, TYPE_SUBSCRIBED_TO>;
+
+    RemoteSubscriberScope(ScopeOwned<RemoteStream>& remote_stream,
+                          F& subscriber,
+                          uint64_t start_idx,
+                          std::function<void()> done_callback)
+        : base_t(std::move(
```
dkorolev: Just a thought: doesn't `base_t` have a perfect forwarding constructor (with `template<typename... ARGS>`)?

dkorolev:
Looks great! Мы в офлайн часа на два с половиной, до вечера :)

Sherlock/replicator.h
grixa: It is declared as `std::atomic_bool`. And what the point of acquiring the mutex here and below? It doesn't prevent us from skipping `subscriber_thread_done_` setting (and it's not bad). By the way, there is exact the same code in `sherlock::Stream::~SubscriberThreadInstance`. But I'm thinking in this case we can remove this check at all, because there is one more inside `TerminateSubscription()`. What do you think?

dkorolev: Agree -- and confirming Sherlock's code is battle-tested. The design you use is in fact very clean:

* `valid_` is initialized as `true` in constructor body, and would only be `false` if the constructor did throw -- `valid_` is pretty much semantically const.
* `subscriber_thread_done_`, on the other hand, only changes once, from `true` to `false`.

I'm concerned about the `subscription_established_.Wait([this](bool established) { return established || subscriber_thread_done_; });)` construct. Looks like both bad corner cases are possible:
1. If the server is not responding at all (which needs a test too), as internal termination has been triggered, the above`.Wait(...)`  may well wait forever, as it's not being notified.
2. On the other hand, the code in `TerminateSubscription()`, right after that `Wait()` might as well get executed twice: for instance, once from `ScopeOwnedBySomeoneElse()`'s callback, and one from this destructor. Also, as you don't clear `subscription_id_`, you may end up requesting to terminate twice (which is no harm at all, and your `TerminateSubscription()` ensures it's reentry-safe thanks to the top-level mutex lock, but still).

A better solution is a (much) longer stare and extensive tests for corner cases. A good enough solution, if you ask me, is to put more state into one `WaitableAtomic`, which would serve both as single state maintainer and the locking primitive to ensure 

Disclaimer: I'm no expert in multithreading; you may well have both more experience and have put more thought into   this snippet. Would be happy to be enlightened. Thank you! :innocent:

dkorolev:
:+1:

mzhurovich:
Ok, merging this and let's move further :)

PR #585, opened 2016-07-31
CURRENT_OBJECT_INTERFACE + CURRENT_FORWARD_METHOD.

dkorolev:
`CURRENT_OBJECT_INTERFACE` + `CURRENT_FORWARD_METHOD`.

I kept writing `CURRENT_INTERFACE_METHOD` instead of `CURRENT_EXPOSE_METHOD`, which led me to thinking `EXPOSE` is not the right word. Max, what do you think of `FORWARD` instead?

Thanks,
Dima


mzhurovich:
1) `FORWARD` is fine.
2) Not sure, whether `/utils` is the best place for ObjectInterface, but let it be here for now.

PR #587, opened 2016-08-01
A bit more clean `RemoteSubscriberThread` termination.

grixa:
CC @dkorolev @mzhurovich

### DIFF
--- a/Sherlock/replicator.h
+++ b/Sherlock/replicator.h
@@ -90,11 +90,12 @@ class SubscribableRemoteStream final {
                            uint64_t start_idx,
                            std::function<void()> done_callback)
         : valid_(false),
-          done_callback_(done_callback),
-          terminate_subscription_requested_(false),
           remote_stream_(remote_stream, [this]() { TerminateSubscription(); }),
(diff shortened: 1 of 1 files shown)


Sherlock/replicator.h
```diff
@@ -171,39 +170,39 @@ class SubscribableRemoteStream final {
       const auto idxts = ParseJSON<idxts_t>(split[0]);
       assert(idxts.index == index_);
       auto entry = ParseJSON<TYPE_SUBSCRIBED_TO>(split[1]);
-      const idxts_t unused_idxts;
       ++index_;
-      if (subscriber_(std::move(entry), idxts, unused_idxts) == ss::EntryResponse::Done) {
+      if (subscriber_(std::move(entry), idxts, unused_idxts_) == ss::EntryResponse::Done) {
```
mzhurovich: Note to self: `unused_idxts_` is very dangerous concept that can cause a lot of troubles if subscriber relies on the value passed here to be valid. I.e. interfaces/behavior of universal subscribers that can be attached to the local and remote streams must be revised.
CC @dkorolev

PR #596, opened 2016-08-05
Yet another implementation of the `RemoteSubscriberThread` termination.

grixa:
CC @dkorolev @mzhurovich


mzhurovich:
@grixa It's not obvious, why this code is better than the previous version :) How about adding a test with a hundred subscribers, being terminating in different ways?

grixa:
The main improvement is that now there is no synchronisation "gap" between `Wait` and analysing  the `subscribtion_id_` value. And you are right, we need more tests, not the only one from `Storage.ReplicationViaHTTP`

PR #600, opened 2016-08-09
Sherlock / F# extensions.

dkorolev:


Sherlock/pubsub.h
```diff
@@ -154,6 +154,8 @@ struct ParsedHTTPRequestParams {
   // If set, stop serving after the response size reached/exceeded the value.
   // Controlled by `stop_after_bytes` URL parameter.
   uint64_t stop_after_bytes = 0u;
+  // If set, do not prepend each entry with its index and timestamp followed by a '\t', return data JSONs only.
+  bool data = false;
```
mzhurovich: What do you think about something more specific like `entries_only`?

mzhurovich: Or `no_idx_info`.

dkorolev: Thought it's too long for the URL. Let me make the change though. :)

sompylasar: URL parameters, being a part of a public API, should be self-explanatory and have as few shortenings and abbreviations as possible. [2K](http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers) "ought to be enough for anybody" (c)

PR #601, opened 2016-08-09
Workaround for VC++ decltype + function constexpr return bug

nyospe:
Another workaround for an MSVC++ bug.


dkorolev:
Fingers crossed && in production with it.

PR #609, opened 2016-08-15
Compilation Improvements RC4.

dkorolev:


regression_tests/type_system/struct_fields_reflection.cc
```diff
@@ -0,0 +1,49 @@
...
+namespace type_test {
+
+#include "include/struct_fields.h"
+
+}  // namespace type_test
+
+TEST(TypeTest, StructFields) {
+  using namespace type_test;
+  using current::reflection::StructSchema;
+  using current::reflection::Language;
```
mzhurovich: We don't need these two usings here, right?

regression_tests/type_system/test.sh
```diff
@@ -3,6 +3,7 @@ COUNT_LIST=${*:-10 25 100 500}
 TEST_LIST=(
     'current_struct'
```
mzhurovich: `current_structs` -> `current_structs_reflection`?

regression_tests/type_system/struct_fields_reflection.cc
dkorolev: :1: 5a86a94

regression_tests/type_system/test.sh
dkorolev: :+1: 5a86a94

TypeSystem/variant.h
```diff
@@ -43,20 +52,91 @@ namespace current {
...
+  RuntimeTypeListHelpersImpl() : types_(PopulateTypes()) {}
+  template <typename T>
+  void AssertContains() const {
+    if (types_.find(std::type_index(typeid(T))) == types_.end()) {
+      std::cerr << "Variant type mismatch.\n\tType: " << reflection::CurrentTypeNameAsConstCharPtr<T>()
+                << "\nIs not among:";
+      for (const auto& type : types_) {
+        std::cerr << ' ' << type.second;
+      }
+      std::cerr << '\n';
```
mzhurovich: Let's remove or `ifdef` this `cerr` stuff?

dkorolev: I thought of it -- the only downside it extra output of the test, and the upside is simplified debugging.

Perhaps `#ifndef NDEBUG` ?

PR #622, opened 2016-09-02
Supported partial or multiple entries per chunk in the `RemoteSubscriberThread`.

grixa:
CC @dkorolev @mzhurovich

### DIFF
--- a/Sherlock/replicator.h
+++ b/Sherlock/replicator.h
@@ -148,7 +148,7 @@ class SubscribableRemoteStream final {
           break;
         } catch (current::Exception&) {
         }
-        unprocessed_data_.clear();
+        carried_over_data_.clear();
         subscription_id_.MutableScopedAccessor()->clear();
       }
     }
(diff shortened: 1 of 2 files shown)


Sherlock/replicator.h
```diff
@@ -163,17 +164,22 @@ class SubscribableRemoteStream final {
...
-      const auto split = current::strings::Split(chunk, '\t');
-      CURRENT_ASSERT(split.size() == 2u);
-      const auto idxts = ParseJSON<idxts_t>(split[0]);
-      CURRENT_ASSERT(idxts.index == index_);
-      auto entry = ParseJSON<TYPE_SUBSCRIBED_TO>(split[1]);
-      ++index_;
-      if (subscriber_(std::move(entry), idxts, unused_idxts_) == ss::EntryResponse::Done) {
-        CURRENT_THROW(StreamTerminatedBySubscriber());
+      const std::string combined_data = unprocessed_data_ + chunk;
+      const auto lines = current::strings::Split(combined_data, '\n');
```
dkorolev: `ByLines` is cleaner, and would work with Window$ CRLF style.

Sherlock/replicator.h
```diff
@@ -163,17 +164,22 @@ class SubscribableRemoteStream final {
...
+      size_t processed_bytes = 0;
+      for (size_t i = 0, sz = combined_data.back() == '\n' ? lines.size() : lines.size() - 1; i < sz; ++i) {
+        const auto split = current::strings::Split(lines[i], '\t');
+        CURRENT_ASSERT(split.size() == 2u);
+        const auto idxts = ParseJSON<idxts_t>(split[0]);
+        CURRENT_ASSERT(idxts.index == index_);
+        auto entry = ParseJSON<TYPE_SUBSCRIBED_TO>(split[1]);
+        ++index_;
+        processed_bytes += lines[i].length() + 1;
```
dkorolev: :-(

Sherlock/replicator.h
```diff
@@ -163,17 +164,22 @@ class SubscribableRemoteStream final {
...
+        if (subscriber_(std::move(entry), idxts, unused_idxts_) == ss::EntryResponse::Done) {
+          CURRENT_THROW(StreamTerminatedBySubscriber());
+        }
       }
+      unprocessed_data_ = combined_data.substr(processed_bytes);
```
dkorolev: Factoring out the `sz = ...` condition to above the for-loop would be cleaner.

Sherlock/replicator.h
```diff
@@ -204,6 +210,7 @@ class SubscribableRemoteStream final {
     current::WaitableAtomic<std::string> subscription_id_;
     std::atomic_bool terminate_subscription_requested_;
     std::thread thread_;
+    std::string unprocessed_data_;
```
dkorolev: Nit: some `carried_over_data_` could be a better name.

dkorolev:
Thanks! Minor comments, and a test would be nice too.

grixa:
@dkorolev, do you have an idea how a simple test for this should look like?

dkorolev:
Just an HTTP endpoint returning two or more records as two or more chunks broken at the wrong boundary, for example, [this one](https://github.com/C5T/Current/blob/216f55589bd1b3a2c179e773e01c8759c58302ac/Blocks/HTTP/test.cc#L454-L463).

I'd test `'\n'`, `'\r\n`, and data broken between `\r` and `\n`. And `Split<ByLines>` ([example](https://github.com/C5T/Current/blob/216f55589bd1b3a2c179e773e01c8759c58302ac/Bricks/strings/test.cc#L232)) would do the trick wrt to line breaks.

dkorolev:
LGTus!

PR #623, opened 2016-09-03
Regression tests to support both g++ and clang++.

dkorolev:
Compilation failed on `g++-4.9` on Hetz, here's the fix, confirmed to work with `clang++` as well.


regression_tests/type_system/test.sh
```diff
@@ -22,6 +22,7 @@ TEST_LIST=(
 
 CPLUSPLUS=${CPLUSPLUS:-g++}
 CPPFLAGS=${CPPFLAGS:- -std=c++11 -ftemplate-backtrace-limit=0  -ftemplate-depth=10000}
+LDFLAGS=${LDFLAGS:- -pthread}
```
mzhurovich: Line 28 adds platform-dependent LDFLAGS, why do you need it here?

dkorolev: :+1: f0d25a8

PR #633, opened 2016-09-09
Integers are Doubles too.

dkorolev:
The fix for:
```

{"success":false,"message":null,"error":{"name":"ParseJSONError","message":"Invalid JSON in request body.","details":{"error_details":"Expected double for `avg_blah_blah`, got: 0"}}}

```

### DIFF
--- a/TypeSystem/Serialization/test.cc
+++ b/TypeSystem/Serialization/test.cc
@@ -60,6 +60,14 @@ CURRENT_STRUCT(Serializable) {
   bool operator<(const Serializable& rhs) const { return i < rhs.i; }
 };
 
+CURRENT_STRUCT(Int) {
+  CURRENT_FIELD(x, int32_t, 0);
+};
+
+CURRENT_STRUCT(Double) {
+  CURRENT_FIELD(x, double, 0);
+};
+
 CURRENT_STRUCT(ComplexSerializable) {
   CURRENT_FIELD(j, uint64_t);
   CURRENT_FIELD(q, std::string);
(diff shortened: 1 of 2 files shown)


TypeSystem/Serialization/test.cc
```diff
@@ -60,6 +60,14 @@ CURRENT_STRUCT(Serializable) {
   bool operator<(const Serializable& rhs) const { return i < rhs.i; }
 };
 
+CURRENT_STRUCT(Int) {
+  CURRENT_FIELD(x, int32_t, 0);
```
sompylasar: Shouldn't you check for `float` as well, as this is where the change happened as well.

dkorolev: You're lucky to poke me at the right time (a few whiskeys, not asleep yet): 934fc6c

sompylasar: @dkorolev,  you seem to repeatedly show me that you don't need my reviews with such "keep Ivan happy" and "whiskey" comments. I'm not a telepath to see you as being drunk so that I don't accidentally interfere with that at the wrong time. Please express yourself clearly and I will unwatch this repo so that we're both happy.

dkorolev: Good catch!

Just realized this CR has already been merged. The fix is going live with https://github.com/C5T/Current/pull/634.

Telepathy and good engineering are orthogonal. And it's never a wrong time for the right comment. Thanks!

mzhurovich: @sompylasar Why so serious? :) Your comments are really appreciated, and Dima being drunk doesn't mean he won't fix what should be fixed :)

dkorolev: I take issue with the above two comments. Whiskey and drunk, while not necessarily orthogonal, are not identical.

sompylasar: Okay then :) happy to help.

PR #637, opened 2016-09-10
The top-level `make test` fix.

dkorolev:


dkorolev:
@mzhurovich A very valid point is that `using namespace current` should not break anything in Current. We'll address it separately.

dkorolev:
Notes:

1. The issue is with the macro expanding templated syntax as part of its arguments.
2. The issue is easy to reproduce by adding `using namespace current;` to the top of `RipCurrent/test.cc`
3. The top-level issue is easy to reproduce by adding `using namespace current;` to the top of `everything.cc` generated by the top-level `make test`.

I'll get there. Eventually.

dkorolev:
And the top-level `make test` from this pull request does pass.

mzhurovich:
@dkorolev 👍

PR #640, opened 2016-09-16
Be more tolerant to header names.

dkorolev:


dkorolev:
@mzhurovich Pls post-review. ;)

PR #647, opened 2016-09-19
Attacking HTTP RAM usage.

dkorolev:
Hi Max,

Here's my attempt to keep RAM usage of chunked HTTP responses at bay.

Thanks,
Dima


mzhurovich:
@dkorolev Looks like it works and solves the problem for now. Thanks!

PR #654, opened 2016-09-28
Patch? PATCH!

mzhurovich:
@dkorolev Here is your PATCH code with minor REST API tweaks and couple of additional tests.


dkorolev:
Patch.

PR #656, opened 2016-10-01
RipCurrent syntactic sugar tweaks.

dkorolev:
Hi Max!

A few tweaks from the airplane. Cleaner syntax, and, looks like, I've fixed the global `using namespace current;` top-level `make test` issue in the meantime.

Worth testing on Windows as I factored out the `REMOVE_PARENTHESES` macro into a dedicated header, and extended it with `__VA_ARGS__`.

Thanks,
Dima

### DIFF
--- a/TypeSystem/remove_parentheses.h
+++ b/TypeSystem/remove_parentheses.h
@@ -1,0 +1,52 @@
+/*******************************************************************************
+The MIT License (MIT)
+
+Copyright (c) 2015 Maxim Zhurovich <zhurovich@gmail.com>
+          (c) 2015 Dmitry "Dima" Korolev <dmitry.korolev@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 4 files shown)


dkorolev:
And confirmed the top-level `make test` does the job both in vanilla build, and if `using namespace current;` is done from the global scope.

The top-level `make check` also passes.

TypeSystem/remove_parentheses.h
```diff
@@ -0,0 +1,51 @@
...
+#define EMPTY_CF_TYPE_EXTRACT
+#define CF_TYPE_PASTE(x, ...) x##__VA_ARGS__
+#define CF_TYPE_PASTE2(x, ...) CF_TYPE_PASTE(x, __VA_ARGS__)
+#define CF_TYPE_EXTRACT(...) CF_TYPE_EXTRACT __VA_ARGS__
+
+// Was. -- D.K. TODO: Test on Windows.
+// #define REMOVE_PARENTHESES(x) CF_TYPE_PASTE2(EMPTY_, CF_TYPE_EXTRACT x)
+
+// Now. -- D.K.
+#define REMOVE_PARENTHESES(...) CF_TYPE_PASTE2(EMPTY_, CF_TYPE_EXTRACT __VA_ARGS__)
```
sompylasar: Shouldn't `#define` macros always be namespace-prefixed to avoid clashing in the global namespace of a program? I remember blaming `winapi.h` for defining something too generic without any prefix...

dkorolev: :+1: 1632d94.

PR #659, opened 2016-10-08
The "priority MMQ" to allow `PublishIntoTheFuture` and `UpdateHead`.

dkorolev:
Hi @mzhurovich and CC @grixa.

I've put together a piece of logic to allow joining events from the future with events from the present using timestamps as sort keys.

The last two tests should be self-explanatory. Requires comments are:

* I've kept MMQ (and MMPQ) indexes 1-based, as historically zero return code from MMQ's `Publish()` is used to indicate the message has been dropped.
* Obviously, `std::priority_queue<>` would outperform `std::set<>`, but sadly it doesn't allow extracting the top element by rvalue reference. This change is just to prove the semantics though; the production code would use `std::make_heap()` etc.
* @grixa, don't be quick to use this piece, as a) (although I generally like it) it's just an architectural suggestion which may evolve, and b) more cool stuff is coming soon.

Thanks,
Dima

### DIFF
--- a/Blocks/MMQ/mmpq.h
+++ b/Blocks/MMQ/mmpq.h
@@ -31,7 +31,7 @@ SOFTWARE.
 #include <condition_variable>
 #include <mutex>
 #include <thread>
-#include <queue>
+#include <set>
 
 #include "../SS/ss.h"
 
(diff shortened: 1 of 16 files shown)


Blocks/MMQ/mmpq.h
```diff
@@ -0,0 +1,171 @@
...
+        std::lock_guard<std::mutex> lock(mutex_);
+        destructing_ = true;
+        condition_variable_.notify_all();
+      }
+      consumer_thread_.join();
+    }
+  }
+
+  // Adds a message to the buffer. Supports both copy and move semantics. THREAD SAFE.
+  using MutexLockStatus = current::locks::MutexLockStatus;
```
mzhurovich: I'm not against `using`, it's perfectly fine. I was speaking about flipping the lines to make the comment belong to the function it describes.

Karl/test.cc
```diff
@@ -472,10 +472,10 @@ TEST(Karl, DeregisterWithNginx) {
   {
     // Must wait for Nginx config reload to take effect.
     while (HTTP(GET(is_prime_proxied_status_url)).code != HTTPResponseCode.NotFound) {
-      std::this_thread::sleep_for(std::chrono::milliseconds(10));  // Spin lock.
```
mzhurovich: AFAIR we've found that `sleep_for` works better in this test, no?

Blocks/SS/idx_ts.h
```diff
@@ -44,6 +44,26 @@ CURRENT_STRUCT(IndexAndTimestamp) {
...
+};
+
+struct InconsistentTimestampException : InconsistentIndexOrTimestampException {
+  using InconsistentIndexOrTimestampException::InconsistentIndexOrTimestampException;
+  InconsistentTimestampException(std::chrono::microseconds expected, std::chrono::microseconds found) {
+    using current::strings::Printf;
+    SetWhat(Printf("Expecting timestamp >= %llu, seeing %llu.", expected.count(), found.count()));
+  }
+};
+
```
mzhurovich: Why not in `exceptions.h`?

Blocks/MMQ/mmpq.h
```diff
@@ -0,0 +1,169 @@
...
+      queue_.erase(it);
+    }
+  }
+
+  bool consumer_thread_created_ = false;
+
+  // The instance of the consuming side of the FIFO buffer.
+  consumer_t& consumer_;
+
+  // The `Entry` struct keeps the entries along with their completion status.
```
mzhurovich: There is no completion status in this struct.

Karl/test.cc
dkorolev: Good catch, e228158.

Blocks/SS/idx_ts.h
dkorolev: :+1: 98ad1ca

Blocks/MMQ/mmpq.h
dkorolev: :+1: dd2edd7 :+1:

Blocks/MMQ/mmpq.h
dkorolev: Let's keep them in their present, no-using, form? :)

dkorolev:
@mzhurovich Thx for the review && PTAL.

PR #663, opened 2016-10-15
Do not throw the `InconsistentTimestampException` in case of publishing with zero timestamp into a fresh and empty MMQ.

grixa:
CC @dkorolev @mzhurovich


dkorolev:
Aren't the initial timestamps initialized by `std::chrono::microseconds(-1)` in both MMQ and MMPQ?

PR #664, opened 2016-10-15
UpdateHead for sherlock and HEAD directive for the file persister.

grixa:
CC @dkorolev @mzhurovich


Blocks/Persistence/file.h
```diff
@@ -69,32 +69,34 @@ class IteratorOverFileOfPersistedEntries {
...
+        if (current.us < next_.us) {
+          // Timestamps must monotonically increase.
+          CURRENT_THROW(ss::InconsistentTimestampException(next_.us, current.us));
+        }
+        onEntry(current, line_.c_str() + tab_pos + 1);
+        next_ = current;
+        ++next_.index;
+        ++next_.us;
+      } else {
+        onDirective(line_.substr(1, tab_pos).c_str(), line_.c_str() + tab_pos + 1);
```
dkorolev: `line_.c_str() + ` :-)

Blocks/Persistence/file.h
```diff
@@ -158,52 +160,40 @@ class FilePersister {
...
+              }
+              offset.push_back(current_offset);
+              timestamp.push_back(current.us);
+              current_offset = fi.tellg();
+              head = current.us;
+              head_offset = 0;
+            },
+            [&](const char* key, const char* value) {
+              if (key == constants::kHeadDirective) {
+                const auto us = std::chrono::microseconds(current::FromString<int64_t>(value));
```
dkorolev: Perhaps `typedef` this underlying `int64` type, and put it next to the `"%020lld` constant?

Blocks/Persistence/file.h
grixa: You mean, use the trick for the first arg or don't use for the second one? :)

Blocks/Persistence/file.h
```diff
@@ -131,38 +144,54 @@ class FilePersister {
...
+                if (!(us > head)) {
+                  CURRENT_THROW(ss::InconsistentTimestampException(head + std::chrono::microseconds(1), us));
+                }
+                head = us;
+                head_offset = std::streamoff(fi.tellg()) - strlen(value);
```
grixa: This `strlen` looks pretty ugly. May be better declare a `constants::kHeadValueLength` and write here:
```
CURRENT_ASSERT(strlen(value) == constants::kHeadValueLength);
head_offset = std::streamoff(fi.tellg()) - constants::kHeadValueLength;
```
BTW, this version also ensures that we could rewrite the value later.

Blocks/Persistence/file.h
```diff
@@ -62,26 +69,30 @@ class IteratorOverFileOfPersistedEntries {
...
       }
-      const auto current = ParseJSON<idxts_t>(line_.substr(0, tab_pos));
-      if (current.index != next_.index) {
-        // Indexes must be strictly continuous.
-        CURRENT_THROW(ss::InconsistentIndexException(next_.index, current.index));
-      }
-      if (current.us < next_.us) {
-        // Timestamps must monotonically increase.
-        CURRENT_THROW(ss::InconsistentTimestampException(next_.us, current.us));
+      if (line_[0] != constants::kDirectiveMarker) {
```
mzhurovich: We definitely need some comments on directive lines format.

Blocks/Persistence/file.h
mzhurovich: Or you can write `\0` instead of tab char :)

Blocks/Persistence/file.h
```diff
@@ -131,38 +144,54 @@ class FilePersister {
...
+                CURRENT_THROW(ss::InconsistentTimestampException(head + std::chrono::microseconds(1), current.us));
+              }
+              offset.push_back(current_offset);
+              timestamp.push_back(current.us);
+              current_offset = fi.tellg();
+              head = current.us;
+              head_offset = 0;
+            },
+            [&](const char* key, const char* value) {
+              if (key == constants::kHeadDirective) {
```
mzhurovich: Just compare the pointers? Shouldn't this be a `strcmp(...)`?

Sherlock/replicator.h
```diff
@@ -145,7 +145,7 @@ class SubscribableRemoteStream final {
                           [this]() {}));
         } catch (StreamTerminatedBySubscriber&) {
           break;
-        } catch (current::Exception&) {
+        } catch (current::Exception& e) {
```
mzhurovich: ?

Sherlock/test.cc
```diff
@@ -750,13 +762,15 @@ TEST(Sherlock, HTTPSubscriptionCanBeTerminated) {
...
-    "{\"index\":0,\"u",
-    "s\":100}\t{\"x\":1}\r",
-    "\n{\"index\":1,\"us\":200}\t{\"x\":2}\n\r\n{\"index\":2,\"us\":300}\t{\"x",
-    "\":3}\n"};
+const std::string sherlock_golden_data_chunks[] = {"{\"index\":0,\"u",
+                                                   "s\":100}\t{\"x\":1}\r",
+                                                   "\n{\"index\":1,\"us\":200}\t{\"x\":2}\n\r\n{\"us\"",
+                                                   ":3",
+                                                   "00}\n{\"index\":2,\"us\":400}\t{\"x",
+                                                   "\":3}\n"};
```
mzhurovich: Let's format it to be human-readable :)

Blocks/Persistence/file.h
grixa: Wow! It means that actually there is no test of parsing the head directive..

mzhurovich: Yeah, and it also means that we need a proper test for this part of the code :-/

Blocks/Persistence/file.h
```diff
@@ -133,36 +150,56 @@ class FilePersister {
...
+              }
+              offset.push_back(current_offset);
+              timestamp.push_back(current.us);
+              current_offset = fi.tellg();
+              head = current.us;
+              head_offset = 0;
+            },
+            [&](const char* key, const char* value) {
+              if (!strcmp(key, constants::kHeadDirective)) {
+                const auto us = std::chrono::microseconds(current::FromString<constants::head_value_t>(value));
```
grixa: this `constants::head_value_t` looks a bit strange. Shouldn't it be `using head_value_t = int64_t` inside the `FilePersister` itself?

Blocks/Persistence/file.h
```diff
@@ -62,26 +71,33 @@ class IteratorOverFileOfPersistedEntries {
     }
   }
 
-  template <typename F>
-  bool ProcessNextEntry(F&& f) {
+  template <typename F1, typename F2>
+  bool ProcessNextEntry(F1&& onEntry, F2&& onDirective) {
```
dkorolev: Nit: `on_entry/on_directive` or `f_entry/f_directive` pls, no Java :-)

Blocks/SS/idx_ts.h
```diff
@@ -48,9 +49,38 @@ CURRENT_STRUCT(IndexAndTimestamp) {
   CURRENT_CONSTRUCTOR(IndexAndTimestamp)(uint64_t index, std::chrono::microseconds us) : index(index), us(us) {}
 };
 
+// The same as `IndexAndTimestamp`, but with optional index member.
```
dkorolev: Pls add a comment re. the value, not the meaning. Along the lines of: The timestamp is either the timestamp of the last entry published, or of the last call to `UpdateHead()`. In the former case, `index` is set to the index of the last published entry, and in the latter case `index` is unset.

Blocks/SS/idx_ts.h
```diff
@@ -48,9 +49,38 @@ CURRENT_STRUCT(IndexAndTimestamp) {
...
+CURRENT_STRUCT(TimestampAndOptionalIndex) {
+  CURRENT_FIELD(us, std::chrono::microseconds);
+  CURRENT_FIELD(index, Optional<uint64_t>);
+  CURRENT_USE_FIELD_AS_TIMESTAMP(us);
+  CURRENT_DEFAULT_CONSTRUCTOR(TimestampAndOptionalIndex) : us(0) {}
+  CURRENT_CONSTRUCTOR(TimestampAndOptionalIndex)(std::chrono::microseconds us) : us(us) {}
+  CURRENT_CONSTRUCTOR(TimestampAndOptionalIndex)(std::chrono::microseconds us, uint64_t index) : us(us), index(index) {}
+};
+
+CURRENT_STRUCT(HeadAndOptionalIndexAndTimestamp) {
```
dkorolev: Same comment here pls.

Sherlock/test.cc
```diff
@@ -750,13 +767,19 @@ TEST(Sherlock, HTTPSubscriptionCanBeTerminated) {
...
+    "#head\t00000000000000000500\n";
 
+// clang-format off
 const std::string sherlock_golden_data_chunks[] = {
-    "{\"index\":0,\"u",
-    "s\":100}\t{\"x\":1}\r",
-    "\n{\"index\":1,\"us\":200}\t{\"x\":2}\n\r\n{\"index\":2,\"us\":300}\t{\"x",
-    "\":3}\n"};
+  "{\"index\":0,\"u","s\":100}\t{\"x\":1}\r",
+  "\n{\"index\":1,\"us\":200}\t{\"x\":2}\n\r\n"
```
dkorolev: Bad line wrap. :)

Sherlock/test.cc
```diff
@@ -773,7 +796,13 @@ TEST(Sherlock, PersistsToFile) {
   current::time::SetNow(std::chrono::microseconds(200u));
   persisted.Publish(2);
   current::time::SetNow(std::chrono::microseconds(300u));
+  persisted.UpdateHead();
+  current::time::SetNow(std::chrono::microseconds(400u));
```
dkorolev: Why the `u` suffixes on the numbers here?

Sherlock/test.cc
grixa: @mzhurovich asked to make this data more human-readable... and what do you suggest? :)

Sherlock/test.cc
grixa: Sorry, it's just a copy-paste from above :)

Blocks/SS/idx_ts.h
grixa: For now the `TimestampAndOptionalIndex` is used only to pass head updates to remote subscribers. So it's not really about the _last_ entry or head.. But you are right, there should be more informative comment.

Blocks/Persistence/file.h
```diff
@@ -133,36 +152,58 @@ class FilePersister {
...
+            },
+            [&](const char* key, const char* value) {
+              if (!strcmp(key, constants::kHeadDirective)) {
+                const auto us = std::chrono::microseconds(current::FromString<constants::head_value_t>(value));
+                if (!(us > head)) {
+                  CURRENT_THROW(ss::InconsistentTimestampException(head + std::chrono::microseconds(1), us));
+                }
+                head = us;
+                CURRENT_ASSERT(constants::kHeadValueLength == strlen(value));
+                head_offset = std::streamoff(fi.tellg()) - constants::kHeadValueLength;
```
grixa: You are right, and this code is completely wrong. I temporarily changed it to `head_offset = std::streamoff(current_offset) + strlen(key) + 2;`, but it's also looks non-obvious. I am thinking of passing the whole `line_` to the `on_directive` and parse it inside..

Blocks/Persistence/file.h
```diff
@@ -302,37 +345,68 @@ class FilePersister {
...
-    iterator.us += std::chrono::microseconds(1);
+    ++iterator.next_index;
+    file_persister_impl_->head_offset = 0;
     file_persister_impl_->end.store(iterator);
     return current;
   }
 
-  bool Empty() const noexcept { return !file_persister_impl_->end.load().index; }
-  uint64_t Size() const noexcept { return file_persister_impl_->end.load().index; }
+  void DoUpdateHead(const std::chrono::microseconds timestamp) {
```
grixa: You mean to explain why the mutex isn't used here? And I don't understand your words about "the two method in two classes". I thought I added `DoUpdateHead` and `UpdateHead` methods completely symmetrical to the existing `DoPublish` and `Publish` ones. Perhaps, I should look through that code again :)

Sherlock/test.cc
dkorolev: Moving `'\n`' one line above, I'd assume. Also, keep in mind "\r\n" is Windows-only.

dkorolev:
@grixa re. two methods, there's no clear invariant on whether `Do..()` and `"no-Do"` methods are run from mutex-locked or mutex-unlocked sections. Would be nice to clear this up.

Sherlock/test.cc
grixa: Ah, right. As for "\r\n", yes, it's Windows-only, but one of the purposes of this test was to make sure that the remote subscriber parses entries, split by a single '\n', as well as split by any combination of '\r' and '\n'. (see #622)

mzhurovich: I meant fixing this:

Blocks/SS/idx_ts.h
```diff
@@ -59,6 +60,9 @@ CURRENT_STRUCT(TimestampAndOptionalIndex) {
   CURRENT_CONSTRUCTOR(TimestampAndOptionalIndex)(std::chrono::microseconds us, uint64_t index) : us(us), index(index) {}
 };
 
+// The `head` is either the timestamp of the last published entry or of the last `UpdateHead()` call,
```
dkorolev: Better, `is the timestamp of either ... or ...`.

Blocks/SS/idx_ts.h
```diff
@@ -59,6 +60,9 @@ CURRENT_STRUCT(TimestampAndOptionalIndex) {
...
+// depending on whether the last entry was published after the last `UpdateHead()` call or not.
```
dkorolev: Why not `depending on whether the most recent update was '.Publish()' or '.UpdateHead()'`?

Sherlock/pubsub.h
```diff
@@ -336,11 +336,12 @@ class PubSubHTTPEndpointImpl : public AbstractSubscriberObject {
...
     }
     if (serving_) {
-      if (us > to_timestamp_) {
+      // Stop serving if the limit on timestamp is exceeded.
+      if (to_timestamp_.count() && us > to_timestamp_) {
         return ss::EntryResponse::Done;
       }
       if (!params_.array && !params_.entries_only) {
-        http_response_(JSON<J>(ts_optidx_t(us)));
+        http_response_(JSON<J>(ts_optidx_t(us)) + "\n");
```
dkorolev: Single quotes `'\n'` would do.

Blocks/Persistence/file.h
```diff
@@ -182,14 +181,17 @@ class FilePersister {
...
               head_offset = 0;
             },
-            [&](const char* key, const char* value) {
-              if (!strcmp(key, constants::kHeadDirective)) {
-                const auto us = std::chrono::microseconds(current::FromString<head_value_t>(value));
+            [&](const std::string& value) {
+              static const auto head_key_length = strlen(constants::kHeadDirective);
+              if (!value.compare(0, head_key_length, constants::kHeadDirective)) {
+                auto offset = head_key_length;
+                while (offset < value.length() && (value[offset] == '\t' || value[offset] == ' ')) ++offset;
```
dkorolev: `&& std::isspace(value[offset])` ?

Blocks/Persistence/test.cc
```diff
@@ -144,6 +144,13 @@ TEST(PersistenceLayer, MemoryExceptions) {
     current::time::ResetToZero();
     current::time::SetNow(std::chrono::microseconds(1));
     ASSERT_THROW(impl.Publish("1"), current::ss::InconsistentTimestampException);
+    ASSERT_THROW(impl.UpdateHead(), current::ss::InconsistentTimestampException);
+    current::time::SetNow(std::chrono::microseconds(4));
+    impl.UpdateHead();
+    current::time::ResetToZero();
```
dkorolev: Unnecessary, `SetNow()` won't throw.

Blocks/Persistence/test.cc
```diff
@@ -144,6 +144,13 @@ TEST(PersistenceLayer, MemoryExceptions) {
...
+    current::time::SetNow(std::chrono::microseconds(3));
+    ASSERT_THROW(impl.Publish("1"), current::ss::InconsistentTimestampException);
+    ASSERT_THROW(impl.UpdateHead(), current::ss::InconsistentTimestampException);
```
dkorolev: :+1:

Blocks/Persistence/test.cc
```diff
@@ -338,6 +350,103 @@ TEST(PersistenceLayer, File) {
...
+    current::time::ResetToZero();
+
+    current::FileSystem::WriteStringToFile(
+        "{\"index\":0,\"us\":100}\t{\"s\":\"foo\"}\n"
+        "{\"index\":1,\"us\":200}\t{\"s\":\"bar\"}\n"
+        "#head\t00000000000000000300\n"
+        "#some_other_directive\n"
+        "#head  00000000000000000400\n"
+        "{\"index\":2,\"us\":500}\t{\"s\":\"meh\"}\n"
+        "#head\t \t00000000000000000600\n",
```
dkorolev: :+1:

Blocks/Persistence/test.cc
```diff
@@ -338,6 +350,103 @@ TEST(PersistenceLayer, File) {
...
+    EXPECT_EQ(999, head_idxts.head.count());
+  }
+  EXPECT_EQ(
+      "{\"index\":0,\"us\":100}\t{\"s\":\"foo\"}\n"
+      "{\"index\":1,\"us\":200}\t{\"s\":\"bar\"}\n"
+      "#head\t00000000000000000300\n"
+      "#some_other_directive\n"
+      "#head  00000000000000000400\n"
+      "{\"index\":2,\"us\":500}\t{\"s\":\"meh\"}\n"
+      "#head\t \t00000000000000000700\n"
```
dkorolev: :+1:

Blocks/Persistence/test.cc
sompylasar: Why the directive lines aren't two-JSON-per-line-compliant? Something along `{“directive":"head","data":"0000...."}\t{}\n`
...

Sherlock/pubsub.h
grixa: Right.
BTW, `JSON(ts_optidx_t(timestamp))` produces strings like `{us:"<timestamp>", index:null}`. Is there a simple way to get rid of this `index:null` part?
And one more question: if `J` isn't `JSONFormat::Minimalistic`, `JSON<J>(ts_optidx_t(us, index))` and `JSON<J>(idxts_t(index, us))` produce different strings with different type_ids, right? If so, we need more "compatible" solution.

mzhurovich:
@dkorolev @grixa Guys, I'm worried about the syntax and functionality of FilePersister's iterators. How do we perform shallow (entries only) and deep (including all the directives) iteration? Could there be a problem with dereferencing operators? What about the possibly mutating last entry?
AFAIR the current implementation was pretty much based on the fact that the log is strictly append-only and consists of immutable records.

grixa:
@mzhurovich, the current implementation uses the same `IteratorOverFileOfPersistedEntries`, that is used for the persister initialisation, but passes an empty directive handler. Any directive is just skipped (so it is a "shallow" iteration) and, as far as we rewrite only the last directive in the file, I don't see any way to damage the iteration in dereferencing operators.

dkorolev:
Let's assume shallow-only file iteration. Proxying head updates is the task at Sherlock level, not File level.

dkorolev:
Your second sentence answers the first one -- yes, it's the minimalistic
format. :-)

And TypeIDs are only important when parsing Variant<>-s for now, and it's
intentional. When parsing CURRENT_STRUCT-s as JSONs, it's as backwards-
(and sideways-) compatible as you could imagine.

Dima

dkorolev:
Because overwriting it in-place would be a pain.

Dima

PR #673, opened 2016-10-27
More robust locking in Stream => SS => Persister.

dkorolev:
Hi guys,

A follow up after @grixa's excellent elaboration. I've tried to make it clean that:
- `Publish/UpdateHead` and `DoPublish/DoUpdateHead` all have the template `MutexLockStatus` parameter.
- The `Do*` methods have this parameter required, the non-`Do` ones have it defaulting to `MLS::NeedToLock`.
- In `Sherlock`, where the mutex locked in the stream level and then, deeper down, another mutex is locked on the persistence level, I took the liberty to make the inner call unchecked.

Making the inner call unchecked is generally unsafe if the user is calling methods such as `CurrentHead()` from `InternalExposePersister()`, and I've added a note that whoever does use `InternalExposePersister()` is playing with fire. Maybe we should revisit this decision.

Also, I've made head rewrite in `Blocks/Persistence/file.h` a bit stronger-locked. Open for a conversation re. whether it's how it should be -- perhaps another look specifically for file rewrites could be a better idea.

Thanks!
Dima

### DIFF
--- a/Blocks/Persistence/file.h
+++ b/Blocks/Persistence/file.h
@@ -41,8 +41,9 @@ SOFTWARE.
 
 #include "../../TypeSystem/Serialization/json.h"
 
-#include "../../Bricks/time/chrono.h"
+#include "../../Bricks/sync/locks.h"
 #include "../../Bricks/sync/scope_owned.h"
+#include "../../Bricks/time/chrono.h"
 #include "../../Bricks/util/atomic_that_works.h"
 
 namespace current {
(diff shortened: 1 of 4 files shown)


mzhurovich:
@dkorolev This change most likely breaks the iteration over the persistence file from a separate thread (like HTTP subscriptions). `Iterate()` functions rely on the locked access to the `offset` and `timestamp` maps via `file_persister_impl_->mutex`, which is not locked during a call to `Stream.Publish()`.
```
    std::lock_guard<std::mutex> lock(file_persister_impl_->mutex);
    CURRENT_ASSERT(file_persister_impl_->offset.size() >=
                    current_size);  // "Greater" is OK, `Iterate()` is multithreaded. -- D.K.
    return IterableRange(file_persister_impl_, begin_index, end_index,
                         file_persister_impl_->offset[begin_index]);
```

mzhurovich:
@grixa What do you think?

mzhurovich:
@dkorolev @grixa I'm closing this PR.

PR #674, opened 2016-10-27
Sherlock signature directive

grixa:
CC @dkorolev @mzhurovich


Blocks/Persistence/file.h
```diff
@@ -192,8 +199,16 @@ class FilePersister {
...
                 head_offset = std::streamoff(current_offset) + offset;
-              } else {
-                head_offset = 0;
+              } else if (!value.compare(0, signature_key_length, constants::kSignatureDirective)) {
+                // The signature could be only at the beginning of the file.
+                if (current_offset != 0) {
+                  CURRENT_THROW(InvalidSignaturePosition());
+                }
+                auto offset = signature_key_length;
+                while (offset < value.length() && std::isspace(value[offset])) ++offset;
```
mzhurovich: Braces pls.

Blocks/Persistence/exceptions.h
```diff
@@ -50,6 +50,14 @@ struct NoEntriesPublishedYet : PersistenceException {
   using PersistenceException::PersistenceException;
 };
 
+struct InvalidStreamSignature : PersistenceException {
+  using PersistenceException::PersistenceException;
+};
+
+struct InvalidSignaturePosition : PersistenceException {
```
dkorolev: Nit: `s/Position/Location/`.

Blocks/Persistence/file.h
```diff
@@ -192,8 +199,16 @@ class FilePersister {
                 }
                 head = us;
                 head_offset = std::streamoff(current_offset) + offset;
-              } else {
-                head_offset = 0;
+              } else if (!value.compare(0, signature_key_length, constants::kSignatureDirective)) {
+                // The signature could be only at the beginning of the file.
```
dkorolev: Nit: `s/could be only/could only be/`, or "should only be", or "The signature, if present, should be at ...".

Blocks/Persistence/file.h
```diff
@@ -192,8 +199,16 @@ class FilePersister {
...
+                if (current_offset != 0) {
+                  CURRENT_THROW(InvalidSignaturePosition());
+                }
+                auto offset = signature_key_length;
+                while (offset < value.length() && std::isspace(value[offset])) ++offset;
+                if (value.compare(offset, signature.length(), signature)) {
+                  CURRENT_THROW(InvalidStreamSignature());
```
dkorolev: A diff between old and new signatures would be nice here. I mean, just print them; the user would be able to run `cut` and `jq` on their own.

Blocks/Persistence/file.h
```diff
@@ -203,6 +218,9 @@ class FilePersister {
         // The `next.us` stores the closest possible next entry timestamp,
         // so the last processed entry timestamp is always 1us less.
         end.store({next.index, next.us - std::chrono::microseconds(1), head});
+        if (!current_offset) {
```
dkorolev: Need a comment here on the value of this condition. I'm *guessing* it's "if no entries have been appended yet", but it's only my guess, which is not good enough for a random reader.

Blocks/Persistence/test.cc
```diff
@@ -294,6 +299,12 @@ TEST(PersistenceLayer, File) {
   }
 
   EXPECT_EQ(
+      "#signature\t{\"exposed_namespace\":\"namespace\",\"top_level_name\":\"top_level_name\",\"schema\":{\"types\":[["
```
dkorolev: You know what? Perhaps taking the `Signature` object outside, `EXPECT_EQ`-ing its `JSON()` with a string, and then just `+`-ing that string into the expected value would be a cleaner solution?

Blocks/Persistence/test.cc
```diff
@@ -353,29 +364,35 @@ TEST(PersistenceLayer, FileDirectives) {
...
     EXPECT_EQ(0u, impl.Size());
     EXPECT_EQ(-1, impl.CurrentHead().count());
     const auto head_idxts = impl.HeadAndLastPublishedIndexAndTimestamp();
     ASSERT_FALSE(Exists(head_idxts.idxts));
     EXPECT_EQ(-1, head_idxts.head.count());
   }
+  current::reflection::StructSchema struct_schema;
+  struct_schema.AddType<StorableString>();
+  const std::string signature =
+      "#signature\t" + JSON(current::ss::StreamSignature(namespace_name, struct_schema.GetSchemaInfo())) + '\n';
```
dkorolev: I'm still wondering about `\t`. Why not a space?

Blocks/Persistence/test.cc
```diff
@@ -459,13 +476,42 @@ TEST(PersistenceLayer, FileExceptions) {
...
 
+  const auto namespace_name = current::ss::StreamNamespaceName("namespace", "top_level_name");
   const std::string persistence_file_name = current::FileSystem::JoinPath(FLAGS_persistence_test_tmpdir, "data");
+  current::reflection::StructSchema invalid_schema;
+  invalid_schema.AddType<std::string>();
+  const std::string signature =
+      "#signature\t" + JSON(current::ss::StreamSignature(namespace_name, invalid_schema.GetSchemaInfo())) + '\n';
+
+  {
+    // Invalid signature.
```
dkorolev: A dedicated test, maybe?

Blocks/SS/signature.h
```diff
@@ -0,0 +1,55 @@
...
+#define BLOCKS_SS_SIGNATURE_H
+
+#include "../../TypeSystem/struct.h"
+#include "../../TypeSystem/Schema/schema.h"
+
+namespace current {
+namespace ss {
+
+CURRENT_STRUCT(StreamNamespaceName) {
+  CURRENT_FIELD(exposed_namespace, std::string);
```
dkorolev: Now when I think of it, these two names -- `exposed_namespace` and `top_level_name` -- are a bit too long. @mzhurovich, do you have the same concern? In this case, any suggestions?

I'd go as far as saying something as simple as `"namespace_name"` and `"{struct|type}_name"` would already be better. Obviously, these two are not the suggestions per se -- just a mere observation some rethought and simplification could help us here.

Sherlock/test.cc
```diff
@@ -765,6 +765,12 @@ TEST(Sherlock, HTTPSubscriptionCanBeTerminated) {
 }
 
 const std::string sherlock_golden_data =
+    "#signature\t{\"exposed_namespace\":\"SherlockSchema\",\"top_level_name\":\"TopLevelTransaction\",\"schema\":{"
```
dkorolev: Same comment re. the structure of the golden output here.

Blocks/Persistence/test.cc
grixa: Initially I just thought that if an entry has the `JSON(idxts_t)\tJSON(entry_t)` format, it would be natural to use `\t` to divide the directive name from its value. But it's really not, I agree.

Blocks/Persistence/test.cc
grixa: Yes, but in that way it will be possible to overlook some "harmful" `StreamSignature` modification, and existing data files would become invalid. May be it's not so important..

Blocks/SS/signature.h
mzhurovich: @dkorolev +1 for `namespace_name`. Speaking of `top_level_name` - what about `entry_name`?

dkorolev:
Оставил пару комментов. Давайте уже этот пулреквест дожимать!

Blocks/SS/signature.h
dkorolev: `namespace_name` and `entry_name` :+1: :+1: :+1: 

@mzhurovich @grixa

dkorolev:
Guys, let's merge it already? After https://github.com/C5T/Current/pull/680, I'd suggest.

@mzhurovich @grixa.

mzhurovich:
@dkorolev +1. Merging.

PR #679, opened 2016-11-03
A trick to please `clang-format`.

dkorolev:


mzhurovich:
`struct RIPCURRENT_NODE` is weird if you ask me. We don't write `struct CURRENT_STRUCT` or `struct CURRENT_STORAGE`.
What is the exact problem you're trying to solve here?

dkorolev:
The problem is the need to use `// clang-format off` for each `RIPCURRENT_NODE[_T]`, as w/o it really bad things happen.

For example:

```
// `RCEmit`: The emitter of events. Emits the integers passed to its constructor.
RIPCURRENT_NODE(RCEmit, void, Integer){static std::string UnitTestClassName(){return "RCEmit";
}
RCEmit() {}  // LCOV_EXCL_LINE
RCEmit(int a) { emit<Integer>(a); }
RCEmit(int a, int b) {
  emit<Integer>(a);
  emit<Integer>(b);
}
RCEmit(int a, int b, int c) {
  emit<Integer>(a);
  emit<Integer>(b);
  emit<Integer>(c);
}
}
;
#define RCEmit(...) RIPCURRENT_MACRO(RCEmit, __VA_ARGS__)
```

```
  RCDump(std::vector<int> & ref) : ptr(&ref) {}
  RCDump(std::vector<int> & ref, std::atomic_size_t & cnt) : ptr(&ref), cnt(&cnt) {}
```

```
RIPCURRENT_NODE(EmitString, (), String){// Note: `()` is same as `void`.
                                        EmitString(){emit<String>("Answer");
}
}
;
#define EmitString(...) RIPCURRENT_MACRO(EmitString, __VA_ARGS__)
```

Extra space around `&`.
```
  DumpIntegerAndString(std::vector<std::string> & ref) : ptr(&ref) {}
```

And templates hell:
```
RIPCURRENT_NODE_T(TemplatedEmitter, void, T){
    TemplatedEmitter(){// Too bad we need this ugly syntax for now. -- D.K.
                       RIPCURRENT_TEMPLATED_EMIT((T), T, LifeUniverseAndEverything());
}
}
;
```

dkorolev:
Closing for now.

PR #683, opened 2016-11-11
The Operational Transformation parser.

dkorolev:
We haz the code to replay code pads now, as well as to grab their historical snapshots!


sompylasar:
Cool stuff!

The noun should be "transformation", wikipedia says: https://en.m.wikipedia.org/wiki/Operational_transformation

mzhurovich:
@dkorolev 
1) All the platform-related `ifdef`s should be placed after `port.h` (or `ot.h`, which includes `port.h`).
2) You write "`libstdc++` doesn't have `#include <codecvt>` yet", and "Requires `g++` or Mac OS." below, which are contradictive statements since `g++` uses `libstdc++`.
3) I suppose there are problems with clang on Linux, since without `-stdlib=libc++` clang uses some version of `libstdc++` without `codecvt` support. We may think about setting `-stdlib=libc++` for clang compiler in the Makefile.
4) `codecvt` is support by GCC since 5.1, so it's better to add smth like this
```
#if ... !defined(__clang__) && (__GNUC__ == 5 && (__GNUC_MINOR__ >= 1))
```

dkorolev:
@mzhurovich @sompylasar -- thanks for the review! Let me amend this PR (abused it too much trying to please Travis), and the future is [here](https://github.com/C5T/Current/pull/685).

PR #689, opened 2016-11-16
Fixed an issue with default values of current time arguments

grixa:
Used std::chrono::microseconds(-1) instead of current::time::Now() as default arguments values.
CC @dkorolev @mzhurovich


Blocks/MMQ/mmpq.h
```diff
@@ -74,8 +74,9 @@ class MMPQ {
   // NOTE(dkorolev): `std::enable_if_t<std::is_same<message_t, current::decay<T>>::value, idxts_t>` can't convert
   // `const char*` into an `std::string`, which is essential, as, unlike MMQ, MMPQ is not an `ss::EntryPublisher<>`.
   template <current::locks::MutexLockStatus MLS = current::locks::MutexLockStatus::NeedToLock, class T>
-  idxts_t Publish(T&& message, std::chrono::microseconds timestamp = current::time::Now()) {
+  idxts_t Publish(T&& message, std::chrono::microseconds us = std::chrono::microseconds(-1)) {
```
dkorolev: Why not use `std::chrono:microseconds(-1)` as a constant, and compare to it directly, w/o `.count()`?

Blocks/MMQ/mmq.h
```diff
@@ -103,11 +99,6 @@ class MMQImpl {
   // THREAD SAFE. Blocks the calling thread for as short period of time as possible.
   template <current::locks::MutexLockStatus MLS>
   idxts_t DoPublish(const message_t& message, std::chrono::microseconds timestamp) {
-    if (CHECK_TIMESTAMP_MONOTONICITY) {
-      if (!(timestamp > last_idx_ts_.us)) {
-        CURRENT_THROW(ss::InconsistentTimestampException(last_idx_ts_.us + std::chrono::microseconds(1), timestamp));
```
dkorolev: These checks and `CURRENT_THROW`-s are still essential.

Blocks/MMQ/mmq.h
```diff
@@ -201,13 +187,16 @@ class MMQImpl {
...
   template <bool DROP = DROP_ON_OVERFLOW>
-  typename std::enable_if<DROP, std::pair<bool, size_t>>::type CircularBufferAllocate(
-      std::chrono::microseconds timestamp) {
+  typename std::enable_if<DROP, std::pair<bool, size_t>>::type CircularBufferAllocate(std::chrono::microseconds us) {
     // Implementation that discards the message if the queue is full.
     // MUTEX-LOCKED.
     std::lock_guard<std::mutex> lock(mutex_);
     if (circular_buffer_[head_].status == Entry::FREE) {
       // Regular case.
+      const auto timestamp = us.count() >= 0 ? us : current::time::Now();
```
dkorolev: If in this method there is no default parameter, perhaps just `assert` `us.count() >= 0`?

Blocks/MMQ/mmq.h
grixa: They are inside the `CircularBufferAllocate` now.

Blocks/MMQ/mmq.h
grixa: Yes, there is no default parameter. But it doesn't mean the `CircularBufferAllocate` couldn't be called with `us == std::chrono::microseconds(-1)`, it just mean no one needs to call it without specifying `us` value explicitly. The default `-1` value from `EntryPersister::Publish` would be passed here through the `MMQImpl::DoPublish` without modification. And here, at last, we can call `Now()`.

PR #691, opened 2016-11-21
FnCAS Tweaks

dkorolev:


FnCAS/fncas/mathutil.h
```diff
@@ -135,8 +153,9 @@ inline std::vector<double> Backtracking(F&& eval_function,
       break;
     }
   }
+  OptimizerLogger().Log(std::string("Backtracking: Final value = ") + current::ToString(test_f_value));
 
-  return test_point;
+  return std::make_pair(test_f_value, test_point);
```
mzhurovich: `std::move(test_point)`?

mzhurovich: A couple of thoughts:
`OptimizationResult` has the same structure. We can unify them under some `PointAndValue` structure.
We also use `std::vector<double>` explicitly to define a point in R<sup>n</sup>. As we go further with more sophisticated linear algebra methods, a lack of proper mathematical vector/matrix notation could become troublesome. I don't suggest to spend time on linear algebra engine now, but maybe it's better to add some `Point` class (aliased to `std::vector<double>`) abstraction at this stage, so we can easily upgrade math engine later. What fo you think?

FnCAS/fncas/optimize.h
```diff
@@ -117,45 +121,86 @@ class Optimizer : noncopyable {
...
 
   OptimizationResult Optimize(const std::vector<double>& starting_point) const override {
-    size_t max_steps = 5000;   // Maximum number of optimization steps.
-    double step_factor = 1.0;  // Gradient is multiplied by this factor.
+    size_t max_steps = 5000;                           // Maximum number of optimization steps.
+    double step_factor = 1.0;                          // Gradient is multiplied by this factor.
+    double min_absolute_per_step_improvement = 1e-25;  // Terminate early if the absolute improvement is less than this.
+    double min_relative_per_step_improvement = 1e-25;  // Terminate early if the relative improvement is less than this.
+    double no_improvement_steps_to_terminate =
+        2;  // Terminate early if the improvement condition is broken for this number of iterations.
```
mzhurovich: clang-format prefer no line wraps in comments?

FnCAS/test.cc
```diff
@@ -286,39 +286,40 @@ TEST(FNCAS, OptimizationOfAMemberFunctionForwardsParameters) {
...
   const auto result = fncas::ConjugateGradientOptimizer<PolynomialFunction>().Optimize({5.0, 20.0});
   EXPECT_NEAR(0.0, result.value, 1e-6);
   ASSERT_EQ(2u, result.point.size());
   EXPECT_NEAR(0.0, result.point[0], 1e-6);
   EXPECT_NEAR(0.0, result.point[1], 1e-6);
 }
 
-TEST(FNCAS, OptimizationOfRosenbrockUsingConjugateGradient) {
+TEST(FnCAS, OptimizationOfRosenbrockUsingConjugateGradient) {
+  // fncas::ScopedLogToStderr scope;
```
mzhurovich: Debug leftovers?

mzhurovich:
@dkorolev I've found a bug causing segfault. 
`optimize.h` line 111:
`const Optional<OptimizerParameters> Parameters() const { return parameters_; }`
should be `const Optional<OptimizerParameters>&`

FnCAS/fncas/optimize.h
```diff
@@ -117,45 +121,86 @@ class Optimizer : noncopyable {
...
+          OptimizerLogger().Log("GradientDescentOptimizer: Value " + current::ToString(value) + " at step " +
+                                current::ToString(step));
+          best_candidate = std::min(best_candidate, std::make_pair(value, candidate_point));
+        }
+      }
+      if (!has_valid_candidate) {
+        CURRENT_THROW(FnCASOptimizationException("!fncas::IsNormal(value)"));
+      }
+      if (best_candidate.first / current.first > 1.0 - min_relative_per_step_improvement ||
+          current.first - best_candidate.first < min_absolute_per_step_improvement) {
```
mzhurovich: `std::abs(...)` for both statements?

FnCAS/fncas/mathutil.h
dkorolev: :+1: 5ad49be

FnCAS/test.cc
dkorolev: Sorry about that, fixes.

FnCAS/fncas/optimize.h
dkorolev: The objective function should be strictly decreasing, shouldn't it?

dkorolev:
@mzhurovich Thanks for the catch (the issue is not the return type though but assigning it to a const reference -- fixed regardless) and for the comments. Fixed everything, PTAL.

mzhurovich:
@dkorolev :+1:

PR #692, opened 2016-11-23
Fixed Karl's state update thread wakeup logic.

mzhurovich:
My latest PR introduced some problems to the Karl's state update thread waiting logic. This one should fix it.


dkorolev:
Merging!

Would you like to clear the `force_wakeup` variable from within the lambda, just before returning `true`?

PR #694, opened 2016-11-26
FnCAS refactoring to support gradient compilation.

dkorolev:
## Code

TL;DR:

* Enabled compiling the gradient.
* Added an example solving a somewhat real problem usingf `FnCAS`.
* Refactored and tested stuff through.
 
#### `In FnCAS $`:

Original: `make test`: OK
Uncommented `#define FNCAS_USE_LONG_DOUBLE`: (no JIT) `make test`: OK
With `#define FNCAS_JIT NASM`: `make test`: OK
With `#define FNCAS_JIT CLANG`: `make test`: OK

#### `In examples/Optimize $`:

Original `make`: OK.

With **NASM**: `NDEBUG=1 make && time ./.current/optimize --dim 6 --logtostderr`: OK.

```
ConjugateGradientOptimizer: 74 iterations, 161.757 per second, 1137 f() calls, 148 g() calls, 74 backtracking calls of 989 total steps.

real	0m28.697s  # Including compilation time.
user	0m28.512s
sys	0m0.164s
```

With **CLANG**: `NDEBUG=1 make && time ./.current/optimize --dim 6 --logtostderr`: OK.

```
ConjugateGradientOptimizer: 74 iterations, 142.437 per second, 1137 f() calls, 148 g() calls, 74 backtracking calls of 989 total steps.

real	0m19.509s  # Including compilation time.
user	0m18.736s
sys	0m0.752s
```

With `GradientDescentOptimizerBT` (nasm): NDEBUG=1 make && time ./.current/optimize --dim 6 --logtostderr`: OK.

```
GradientDescentOptimizerBT: 1660 iterations, 167.839 per second, 22730 f() calls, 1660 g() calls, 1660 backtracking calls of 19410 total steps.

real	0m38.514s  # Including compilation time.
user	0m38.276s
sys	0m0.212s
```

With `GradientDescentOptimizer` (nasm): `NDEBUG=1 make && time ./.current/optimize --dim 6 --logtostderr`

```
GradientDescentOptimizer: 5006 iterations, 319.903 per second, 15018 f() calls, 5006 g() calls.

real	0m46.306s  # Including compilation time.
user	0m45.996s
sys	0m0.268s
```

With `#define FNCAS_USE_LONG_DOUBLE` (intermediate mode only, no compiled functions/gradients) : OK

#### `FnCAS/deprecated_test/smoke $ make`

```
./run_smoke_test.sh
g++ --std=c++11 -O2 -DFNCAS_JIT=NASM ../eval.cc -I $PWD/autogen -o $BINARY -ldl
/tmp/ccfxuEOr.o: In function `action_gen_eval_Xeval<eval::compiled>::start()':
eval.cc:(.text._ZN21action_gen_eval_XevalIN4eval8compiledEE5startEv[_ZN21action_gen_eval_XevalIN4eval8compiledEE5startEv]+0xb5): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
arithmetics: OK OK OK OK
math: OK OK OK OK
sum: OK OK OK OK
g++ --std=c++11 -O2 -DFNCAS_JIT=CLANG ../eval.cc -I $PWD/autogen -o $BINARY -ldl
/tmp/ccv6ePjD.o: In function `action_gen_eval_Xeval<eval::compiled>::start()':
eval.cc:(.text._ZN21action_gen_eval_XevalIN4eval8compiledEE5startEv[_ZN21action_gen_eval_XevalIN4eval8compiledEE5startEv]+0xb5): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
arithmetics: OK OK OK OK
math: OK OK OK OK
sum: OK OK OK OK
OK
```

#### `FnCAS/deprecated_test/perf $ make`

Ref. https://gist.github.com/dkorolev/b04a9d7a9ae3e0f96fa92789e6839b2f

### DIFF
--- a/FnCAS/fncas/differentiate.h
+++ b/FnCAS/fncas/differentiate.h
@@ -36,10 +36,10 @@ namespace fncas {
 static const double_t APPROXIMATE_DERIVATIVE_EPS = 1e-4;
 
 template <typename F>
-inline double_t approximate_derivative(F f,
-                                       const std::vector<double_t>& x,
-                                       node_index_type i,
-                                       const double_t EPS = APPROXIMATE_DERIVATIVE_EPS) {
+double_t approximate_derivative(F f,
+                                const std::vector<double_t>& x,
+                                node_index_type i,
+                                const double_t EPS = APPROXIMATE_DERIVATIVE_EPS) {
   std::vector<double_t> x1(x);
   std::vector<double_t> x2(x);
   x1[i] -= EPS;
@@ -48,9 +48,9 @@ inline double_t approximate_derivative(F f,
 }
 
 template <typename F>
-inline std::vector<double_t> approximate_gradient(F f,
-                                                  const std::vector<double_t>& x,
-                                                  const double_t EPS = APPROXIMATE_DERIVATIVE_EPS) {
+std::vector<double_t> approximate_gradient(F f,
+                                           const std::vector<double_t>& x,
+                                           const double_t EPS = APPROXIMATE_DERIVATIVE_EPS) {
   std::vector<double_t> g(x.size());
   std::vector<double_t> xx(x);
   for (size_t i = 0; i < xx.size(); ++i) {
--- a/FnCAS/fncas/differentiate.h
+++ b/FnCAS/fncas/differentiate.h
@@ -102,7 +102,13 @@ inline node_index_type d_f(function_t function, const V& original, const V& x, c
       [](const V&, const V& x, const V& dx) { return V(-1.0) * dx / sqrt(V(1.0) - x * x); },
       // atan().
       [](const V&, const V& x, const V& dx) { return dx / (x * x + 1); },
-  };
+      // zero_or_one().
+      [](const V& o, const V&, const V&) {
+        CURRENT_THROW(FnCASZeroOrOneIsNonDifferentiable());
+        return o;
+      },
+      // zero_or_x().
+      [](const V&, const V& x, const V& dx) { return zero_or_one(x) * dx; }};
   return function < function_t::end ? differentiator[static_cast<size_t>(function)](original, x, dx).index() : 0;
 }
 
--- a/FnCAS/fncas/differentiate.h
+++ b/FnCAS/fncas/differentiate.h
@@ -33,32 +33,32 @@
 
 namespace fncas {
 
-static const double APPROXIMATE_DERIVATIVE_EPS = 1e-4;
+static const double_t APPROXIMATE_DERIVATIVE_EPS = 1e-4;
 
 template <typename F>
-fncas_value_type approximate_derivative(F f,
-                                        const std::vector<fncas_value_type>& x,
-                                        node_index_type i,
-                                        const fncas_value_type EPS = APPROXIMATE_DERIVATIVE_EPS) {
-  std::vector<fncas_value_type> x1(x);
-  std::vector<fncas_value_type> x2(x);
+double_t approximate_derivative(F f,
+                                const std::vector<double_t>& x,
+                                node_index_type i,
+                                const double_t EPS = APPROXIMATE_DERIVATIVE_EPS) {
+  std::vector<double_t> x1(x);
+  std::vector<double_t> x2(x);
   x1[i] -= EPS;
   x2[i] += EPS;
   return (f(x2) - f(x1)) / (EPS + EPS);
 }
 
 template <typename F>
-inline std::vector<fncas_value_type> approximate_gradient(F f,
-                                                          const std::vector<fncas_value_type>& x,
-                                                          const fncas_value_type EPS = APPROXIMATE_DERIVATIVE_EPS) {
-  std::vector<fncas_value_type> g(x.size());
-  std::vector<fncas_value_type> xx(x);
+inline std::vector<double_t> approximate_gradient(F f,
+                                                  const std::vector<double_t>& x,
+                                                  const double_t EPS = APPROXIMATE_DERIVATIVE_EPS) {
+  std::vector<double_t> g(x.size());
+  std::vector<double_t> xx(x);
   for (size_t i = 0; i < xx.size(); ++i) {
-    const fncas_value_type v0 = xx[i];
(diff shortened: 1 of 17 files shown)


FnCAS/fncas/differentiate.h
```diff
@@ -33,32 +33,32 @@
...
+static const double_t APPROXIMATE_DERIVATIVE_EPS = 1e-4;
 
 template <typename F>
-fncas_value_type approximate_derivative(F f,
-                                        const std::vector<fncas_value_type>& x,
-                                        node_index_type i,
-                                        const fncas_value_type EPS = APPROXIMATE_DERIVATIVE_EPS) {
-  std::vector<fncas_value_type> x1(x);
-  std::vector<fncas_value_type> x2(x);
+double_t approximate_derivative(F f,
```
mzhurovich: `inline`?

examples/Optimize/README.md
```diff
@@ -0,0 +1,48 @@
...
+* Each player has a "bag of strategies" to play, and begins with this bag containing one occurrence of each strategy. This player's mixed strategy is assumed to be the weighted sum of all the strategies in their bag.
+* Players take turns at optimizing their bags.
+* At each optimization half-step per-player, knowing the opponent's mixed strategy -- the contents of their bag -- the player under consideration picks their strategy that best counters the present state of the opponent's bag, and adds this newly picked strategy into their own bag.
+* For this exercise, to construct the solution we look at the strategies added into players' bags in the past half of the iterative process. This improves convergence slightly, as early iterations often pick the strategies that don't even present in the final solution, and, while the process would converge nonetheless, it would take longer to outweigh them than to disregard them.
+
+### Optimization-Based Solution
+
+The optimization-based solution (`optimize.cc`) solves the problem by converting it into an optimization problem. To accomplish this:
+
+* The the very linear programming constraints are used as constraits:
```
dkorolev: :+1: самые что ни на есть они, 8ea510b

FnCAS/fncas/differentiate.h
mzhurovich: It's not true, however not relevant in this particular case. 
I've suggested it for consistency reasons, especially that there's `inline std::vector<double_t> approximate_gradient` below.

dkorolev: :+1: 0b3d092

Would be interested to know the truth.

mzhurovich:
And with long double on clang:
```
./../../FnCAS/fncas/mathutil.h:61:9: warning: absolute value function 'fabs' given an argument of type 'value_type' (aka 'long double') but has parameter of
      type 'double' which may cause truncation of value [-Wabsolute-value]
    if (fabs(a[i]) < 1e-100) {
...
./../../FnCAS/fncas/mathutil.h:61:9: note: use function 'std::abs' instead
    if (fabs(a[i]) < 1e-100) {
        ^~~~
        std::abs
```

nyospe:
basic.def.odr (3.2 (6) in C++14) + dcl.fct.spec (7.1.2 note 94) make it
explicit, temp (14 (4&6)) make it clear that this applies to function
templates.

dkorolev:
@mzhurovich : I've started tracing the failing case you pointed out (`./.current/optimize --matrix '[[7,5,7,9],[2,9,9,3],[5,6,9,5],[6,1,6,7]]' --logtostderr`), and found the following.

* The gradient is all `NaN`-s.
* It is such using both intermediate and compiled formats.
* Yet, the value of the objective function is perfectly normal at the starting point, and at the points near it.

(I didn't try `g_approximate`, just added `0.01` to each coordinate and computed the objective function, and the results all made sense.)

Thus, it's the underlying math. I had one idea, and it turned out to be correct: the `sqr(v + sqrt(v * v))` magic is unstable when it comes to differentiation. This is fixed in https://github.com/C5T/Current/pull/694/commits/668877fd7a8a80ca1d915787c23de0a5d27d81dc, with a one-liner.

To actually fix it, I had to add some form of the non-linear `|x|` function into FnCAS. Well, this has been on my mind for a while, and I knew what to do -- added `zero_or_x(x) = max(0, x)` and `zero_or_one(x) = (x >= 0 ? 1 : 0)`, with the latter being the derivative of the former. Took a bit to have them run in intermediate + clang + nasm modes, but it's done now.

@nyospe I assume you're referring to http://eel.is/c++draft/basic.def.odr and http://eel.is/c++draft/dcl.dcl. Still don't quite see what difference would `inline` make in this case, would appreciate a brief elaboration.

Thanks,
Dima

nyospe:
The `inline` keyword is superfluous. It makes absolutely no difference, and the standard says it cannot make a difference. That was my point; there is no reason to decorate a function template with `inline`, ever.

mzhurovich:
@dkorolev @nyospe You're right. However, if you want to add explicit function specialization, `inline` is required, and what I tried to say is that it's easier to always write `inline` for function templates/specializations that are supposed to be de-facto inlined to improve readability.

dkorolev:
Our 'make check' would quickly catch such a blunder, right?

mzhurovich:
@dkorolev It should.

mzhurovich:
@dkorolev Looks like your function is almost `theta(x)*x`

dkorolev:
Yes, written as it's indefinite integral instead to make sure it can be
differentiated.

Suggestions on better naming for those 'zero_or_*' functions?

examples/Optimize/README.md
```diff
@@ -0,0 +1,48 @@
+## Optimize
+
+### Intro
+
+This directory contains the example of using the `FnCAS` library to solve an optimization problem.
+
+The problem considered is the two-player zero-sum matrix game: For an `N` by `N` matrix `A` where `A[i][j]` is the payoff from player two to player one as players have picked up strategies `i` and `j`respectively, find the optimal mixed strategies for both players. The mixed strategy for each player is a simplex of dimension `N` where each element is the probability of the player picking this strategy, which corresponds to the row or column of the matrix for players one and two respectively.
```
dkorolev: :+1: f28a438

PR #699, opened 2016-12-11
Injected FnCAS math into `std::` as well.

dkorolev:
CC @forwidur.


mzhurovich:
I don't like the idea of injecting into `std` namespace:
```
[C++11: 17.6.4.2.1/1]: The behavior of a C++ program is undefined if it adds declarations or definitions to namespace std or to a namespace within namespace std unless otherwise specified. A program may add a template specialization for any standard library template to namespace std only if the declaration depends on a user-defined type and the specialization meets the standard library requirements for the original template and is not explicitly prohibited.
```

dkorolev:
Follow-up after our chat w/ @mzhurovich:

1. Don't violate the standard by default, and move math functions into the `fncas::` namespace, and
2. Add a preprocessor directive that does add symbols into `std::`, as having user code compile under FnCAS right away is worth it.

Thanks,
Dima

dkorolev:
@mzhurovich -- Could you pls peek through the new `test.cc` pls? Still polishing the rest, but managed to refactor away all the `X2V/V2X` magic, making user-facing code much more like a plain `std::vector<double>`.

dkorolev:
@mzhurovich -- could you take a look please?

FnCAS/README.md
```diff
@@ -1,8 +1,131 @@
...
+EXPECT_EQ(4*4 + 3*3, reference(std::vector<double>({-5, -5})));
+ASSERT_EQ(0, number_of_calls);
+
+const fncas::function_t& jit_reference = jit;
+number_of_calls = 0;
+EXPECT_EQ(5, jit_reference(std::vector<double>({0, 0})));
+EXPECT_EQ(4*4 + 3*3, jit_reference(std::vector<double>({-5, -5})));
+ASSERT_EQ(0, number_of_calls);
+
+// Wrap the function into the approximate gradient computer.
```
mzhurovich: Would be good to explain in a couple of words, what does `approximate gradient` mean.

FnCAS/fncas/differentiate.h
```diff
@@ -23,15 +23,16 @@
...
-#define FNCAS_DIFFERENTIATE_H
+#ifndef FNCAS_FNCAS_DIFFERENTIATE_H
+#define FNCAS_FNCAS_DIFFERENTIATE_H
 
 #include <vector>
 
 #include "base.h"
 #include "node.h"
 
 namespace fncas {
```
mzhurovich: Don't we plan to wrap it into `current` namespace?

FnCAS/fncas/node.h
```diff
@@ -404,39 +478,69 @@ using X2V = typename x2v_impl<X>::type;
...
-    result.rhs_index() = rhs.index_;                                        \
-    return result;                                                          \
-  }                                                                         \
-  inline const fncas::V& operator OP2(fncas::V& lhs, const fncas::V& rhs) { \
-    lhs = lhs OP rhs;                                                       \
-    return lhs;                                                             \
+// Arithmetic operations and mathematical functions are defined outside namespace fncas::impl.
+
+#define DECLARE_OP(OP, OP2, NAME)                                                             \
+  inline fncas::impl::V operator OP(const fncas::impl::V& lhs, const fncas::impl::V& rhs) {   \
```
mzhurovich: `::fncas`?

FnCAS/test.cc
```diff
@@ -29,27 +29,49 @@ SOFTWARE.
...
-fncas::X2V<X> SimpleFunction(const X& x) {
+// Need an explicit specialization, as `SimpleFunction<double_t>` is used directly in the test.
+template <typename T>
+T SimpleFunction(const std::vector<T>& x) {
   return ParametrizedFunction(x, 2u);
 }
 
-static_assert(std::is_same<std::vector<fncas::double_t>, fncas::V2X<fncas::double_t>>::value, "");
-static_assert(std::is_same<fncas::X2V<std::vector<fncas::double_t>>, fncas::double_t>::value, "");
+// `variables_vector_t` behaves in a way that makes it a drop-in substitute of `std::vector<term_t>`.
```
mzhurovich: s/variables_vector_t/term_vector_t/?

mzhurovich: Or vice versa?

mzhurovich:
And one more nit - JIT can't be disabled now because `fncas.h` always includes `jit.h`, which contains:
```
#ifndef FNCAS_JIT
#define FNCAS_JIT AS  // Default to `as` as the JIT target for FnCAS.
#endif
```

dkorolev:
Wow, the failing test is in `Bricks/sync`:

```
[ RUN      ] WaitableAtomic.Smoke
test.cc:260: Failure
Expected: (copy_of_object.x) > (10u), actual: 7 vs 10
test.cc:262: Failure
Expected: (copy_of_object.y) > (10u), actual: 7 vs 10
[  FAILED  ] WaitableAtomic.Smoke (44 ms)
```

It's a flaky test though, OK-ish if the Travis instance was heavily overloaded.

Looking into your comments.

FnCAS/README.md
```diff
@@ -1,8 +1,131 @@
...
+// This type allows passing functions around without thinking of whether they
+// are native functions, blueprints, or JIT-compiled dynamically linked `.so`-s.
+// The `2` parameter is the dimensionality of the function.
+fncas::function_reference_t native(simple_function<double>, 2);
+const fncas::function_t& reference = native;
+
+number_of_calls = 0;
+EXPECT_EQ(5, reference(std::vector<double>({0, 0})));
+EXPECT_EQ(4*4 + 3*3, reference(std::vector<double>({-5, -5})));
+ASSERT_EQ(2, number_of_calls);  // By-reference evaluation doesn't call the function.
```
dkorolev: Cleaned up in 88a8cea. Good catch!

FnCAS/README.md
dkorolev: :+1: 88a8cea

FnCAS/fncas/base.h
```diff
@@ -34,14 +34,16 @@
...
-
 #ifndef FNCAS_USE_LONG_DOUBLE
 typedef double double_t;  // Make it simple to switch to `long double` or a custom type.
 #else
 typedef long double double_t;
 #endif
 
+namespace impl {
+
+typedef int64_t node_index_type;  // Allow 4B+ nodes on 64-bit arch, keep signed for (~i) vs. (i) index magic.
```
dkorolev: :+1: 4ff97ce

FnCAS/fncas/differentiate.h
dkorolev: Let's do it in a separate commit and/or pull request.

Yes!

FnCAS/test.cc
```diff
@@ -147,18 +171,18 @@ TEST(FnCAS, CompiledGradientsWrapper) {
 }
 
 TEST(FnCAS, CompiledSqrGradientWrapper) {
-  // The `fncas::sqr()` function is a special case, which it worth unit-testing with different `FNCAS_JIT. -- D.K.
+  // The `sqr()` function is a special case, which it worth unit-testing with different `FNCAS_JIT. -- D.K.
```
dkorolev: :+1: 11c742e

dkorolev:
@mzhurovich Thanks, took care of almost all of the comments.

Re. JIT being always on: Yes, it's always compiled into the binary. For now, when computing the functions and gradients manually, it's in the users' hands to not use JIT. And during the optimization phase, it can be controlled in the parameters.

Looking deeper, perhaps we could simply enable in all the JITs, and let the user choose which one to use as a template parameter, at their own compile time? I.e., get rid of the `FNCAS_JIT` macro altogether, and replace it with an `enum class` that's expected as a template parameter?

Re. `s/fncas::/current::/g` -- Most certainly, and let's do it in a separate commit or pull request. It should be done before FnCAS goes out live, but I considered the refactoring of this pull request major enough to ask for your comments first.

Thanks,
Dima

mzhurovich:
@dkorolev Merging this PR, having in mind our offline chat about controlling JIT.

PR #700, opened 2016-12-13
The EJS demo.

dkorolev:
Gentlemen,

Here's a relatively clean and simple way I have in mind to keep the "source of truth" of the "frontend" of the "app" on the Current side, while allowing to render it somewhat nicely for the end user, with one line change time on the frontend being, well, under one second.

Comments would be much appreciated.

Thanks,
Dima


examples/EJS/cpp/README.md
```diff
@@ -0,0 +1,30 @@
...
+<h1>Primes from 99 to 101</h1>
+<ul>
+   <li>999</li> 
+</ul>
+```
+
+Once both servers are running, "HTML" "beautification" would work just like this:
+
+```
+$ curl -H "Accept: text/html" "localhost:3000?a=10&b=20"
```
sompylasar: 3001?

examples/EJS/cpp/README.md
```diff
@@ -0,0 +1,30 @@
...
+<h1>Primes from 10 to 20</h1>
+<ul>
+   <li>11</li>  <li>13</li>  <li>17</li>  <li>19</li> 
+</ul>
+```
+
+And if that's too nerdy, just open [http://localhost:3000/?a=10&b=20] from the browser.
```
sompylasar: 3001?

examples/EJS/cpp/README.md
```diff
@@ -0,0 +1,30 @@
+# Current + EJS Demo
+
+Just `make` in this directory would build and start a simple server on port 3000:
```
sompylasar: Which port serves what? Not obvious, confusing.

examples/EJS/nodejs/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "CurrentEJS",
```
sompylasar: Packages are generally lowercase and dash-separated. So, `current-ejs` or `current-demo-renderer-ejs` or `c5t-demo-renderer-ejs`.

examples/EJS/nodejs/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "keywords": [
+    "current",
+    "ejs",
+    "cpp",
+    "backend"
+  ],
+  "main": "server.js",
+  "dependencies": {
+    "body-parser": "*",
+    "express": "*",
```
sompylasar: I'd rather not use wildcard versions, stick to the current major at least.

examples/EJS/nodejs/template.ejs
```diff
@@ -0,0 +1,4 @@
+<h1>Primes from <%= a %> to <%= b %></h1>
+<ul>
+  <% for(var i = 0; i < x.length; ++i) { %> <li><%= x[i] %></li> <% } %>
```
sompylasar: You don't need whitespace between `%>` and `<`. Whitespace is sometimes meaningful in HTML (if the blocks are inline, the space gets a nonzero width).

sompylasar: And `> <% ` of course.

examples/EJS/cpp/ejs.cc
```diff
@@ -0,0 +1,88 @@
...
+              }
+            }
+            if (is_prime) {
+              response.x.push_back(p);
+            }
+          }
+          return response;
+        }();
+
+        const bool html = [&request]() {
```
sompylasar: Why the calculating server knows about the renderer(s)?

examples/EJS/cpp/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```
+$ curl "localhost:3000?a=10&b=20"
+{"a":10,"b":20,"x":[11,13,17,19]}
+```
+
+Adding `npm run start` from `../nodejs` would start another, node.js, rendering server on port 3001:
```
mzhurovich: `npm install` first?

examples/EJS/cpp/README.md
dkorolev: Updated README in 83cd181.

examples/EJS/cpp/README.md
dkorolev: It's how it should be. Updated README in 83cd181.

examples/EJS/cpp/README.md
dkorolev: It's how it should be. Updated README in 83cd181.

examples/EJS/cpp/README.md
dkorolev: :+1: b8c5de8

examples/EJS/cpp/ejs.cc
dkorolev: It has the `--ejs_renderer_port` flag set, which defaults to `3001`. If `Accept: text/html` is in the headers, the rendered is invoked.

examples/EJS/nodejs/package.json
dkorolev: :+1: d1f3a14

examples/EJS/nodejs/template.ejs
dkorolev: :+1: 0136832

examples/EJS/nodejs/package.json
dkorolev: :+1:

dkorolev:
@sompylasar @grixa @mzhurovich PTAL && thanks!

examples/EJS/README.md
```diff
@@ -0,0 +1,38 @@
...
+## Details
+
+Just `make` in the `EJS/cpp` directory would build and start a simple server on port 3000:
+
+```
+$ curl "http://localhost:3000?a=10&b=20"
+{"a":10,"b":20,"x":[11,13,17,19]}
+```
+
+Indepenently from the C++ binary, running `npm i && npm run start` from `EJS/nodejs` would start another, node.js, rendering server on port 3001:
```
sompylasar: Indepen*d*ently

examples/EJS/nodejs/package.json
```diff
@@ -0,0 +1,40 @@
...
+  ],
+  "keywords": [
+    "current",
+    "ejs",
+    "cpp",
+    "backend"
+  ],
+  "main": "server.js",
+  "dependencies": {
+    "body-parser": ">=1.15.2",
```
sompylasar: Why `>=`? Either nothing (exact version), or `^` (latest with the same major version).

dkorolev:
@sompylasar :+1: :+1: :+1: 70825ef

mzhurovich:
The other possible design is to make Node.JS app a "rendering layer" that forwards requests to the Current binary, which in its turn serves only data JSONs and knows nothing about any renders etc. (what @sompylasar said). This example stays as is for now.
CC @grixa

PR #701, opened 2016-12-14
Claire tweaks.

mzhurovich:


Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+TEST(Karl, ChangeDependenciesInClaire) {
+  current::time::ResetToZero();
+
+  const auto stream_file_remover = current::FileSystem::ScopedRmFile(FLAGS_karl_test_stream_persistence_file);
+  const auto storage_file_remover = current::FileSystem::ScopedRmFile(FLAGS_karl_test_storage_persistence_file);
+  const unittest_karl_t karl(UnittestKarlParameters());
+  const current::karl::Locator karl_locator(Printf("http://localhost:%d/", FLAGS_karl_test_keepalives_port));
+  const uint16_t claire_port = PickPortForUnitTest();
+  current::karl::Claire claire(karl_locator, "unittest", claire_port, {"http://127.0.0.1:12345"});
+  claire.Register(nullptr, true);
```
dkorolev: This leaves one wondering what does `nullptr` signify. A comment, perhaps?

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+
+  // Karl sees single `claire`'s dependency on `127.0.0.1:12345` as unresolved.
+  {
+    unittest_karl_status_t status;
+    const auto body = HTTP(GET(Printf("http://localhost:%d?full&from=0", FLAGS_karl_test_fleet_view_port))).body;
+    ASSERT_NO_THROW(status = ParseJSON<unittest_karl_status_t>(body)) << body;
+#if 0
```
dkorolev: This `#if 0` section is safe to remove, right?

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+            HTTP(GET(Printf("http://localhost:%d?from=0", FLAGS_karl_test_fleet_view_port))).body));
+#endif
+    EXPECT_EQ(1u, status.machines.size()) << JSON(status);
+    ASSERT_TRUE(status.machines.count("127.0.0.1")) << JSON(status);
+    auto& server = status.machines["127.0.0.1"];
+    auto& per_ip_services = server.services;
+    ASSERT_EQ(1u, per_ip_services.size());
+    auto& claire_service_info = per_ip_services[claire.Codename()];
+    ASSERT_EQ("unittest", claire_service_info.service);
+    EXPECT_EQ(0u, claire_service_info.dependencies.size());
```
dkorolev: `EXPECT_TRUE(claire_service_info.dependencies.empty());` ?

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+    auto& unresolved_dependencies = claire_service_info.unresolved_dependencies;
+    ASSERT_EQ(1u, unresolved_dependencies.size());
+    ASSERT_EQ("http://127.0.0.1:12345/.current", unresolved_dependencies[0]);
+  }
+
+  // Launch another Claire to create resolvable dependency.
+  const int claire_to_depend_on_update_ts = 1000;
+  current::time::SetNow(std::chrono::microseconds(claire_to_depend_on_update_ts),
+                        std::chrono::microseconds(claire_to_depend_on_update_ts + 100));
+  const uint16_t claire_to_depend_on_port = PickPortForUnitTest();
```
dkorolev: Several `PickPortForUnitTest()`-s executed during test runtime could collide. Perhaps a flag with a default parameter equals `PickPortForUnitTest()` (or just a global static variable to force the call to complete at binary startup)?

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+
+  // Wait for Karl to receive `claire`'s latest keepalive.
+  {
+    const std::string karl_status_url =
+        Printf("http://localhost:%d?full&from=%d", FLAGS_karl_test_fleet_view_port, last_update_ts);
+    bool karl_received_last_keepalive = false;
+    while (!karl_received_last_keepalive) {
+      const auto status = TryParseJSON<unittest_karl_status_t>(HTTP(GET(karl_status_url)).body);
+      ASSERT_TRUE(Exists(status));
+      karl_received_last_keepalive = Value(status).machines.size() > 0u;
```
dkorolev: `!empty()`.

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+    ASSERT_EQ(1u, per_ip_services.size());
+    auto& claire_service_info = per_ip_services[claire.Codename()];
+    ASSERT_EQ("unittest", claire_service_info.service);
+    EXPECT_EQ(0u, claire_service_info.dependencies.size());
+    auto& unresolved_dependencies = claire_service_info.unresolved_dependencies;
+    ASSERT_EQ(1u, unresolved_dependencies.size());
+    ASSERT_EQ("http://127.0.0.1:12345/.current", unresolved_dependencies[0]);
+  }
+
+  // Launch another Claire to create resolvable dependency.
```
dkorolev: To create **a** resolvable dependency.

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+
+  const auto stream_file_remover = current::FileSystem::ScopedRmFile(FLAGS_karl_test_stream_persistence_file);
+  const auto storage_file_remover = current::FileSystem::ScopedRmFile(FLAGS_karl_test_storage_persistence_file);
+  const unittest_karl_t karl(UnittestKarlParameters());
+  const current::karl::Locator karl_locator(Printf("http://localhost:%d/", FLAGS_karl_test_keepalives_port));
+  const uint16_t claire_port = PickPortForUnitTest();
+  current::karl::Claire claire(karl_locator, "unittest", claire_port, {"http://127.0.0.1:12345"});
+  claire.Register(nullptr, true);
+
+  // Karl sees single `claire`'s dependency on `127.0.0.1:12345` as unresolved.
```
dkorolev: `s/single claire's dependency/claire's single dependency/`

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+        Printf("http://localhost:%d?full&from=%d", FLAGS_karl_test_fleet_view_port, last_update_ts);
+    bool karl_received_last_keepalive = false;
+    while (!karl_received_last_keepalive) {
+      const auto status = TryParseJSON<unittest_karl_status_t>(HTTP(GET(karl_status_url)).body);
+      ASSERT_TRUE(Exists(status));
+      karl_received_last_keepalive = Value(status).machines.size() > 0u;
+    }
+  }
+
+  // Karl lists resolved dependency for `claire`.
```
dkorolev: lists **a** resolved dependency.

Karl/test.cc
```diff
@@ -744,6 +744,88 @@ TEST(Karl, ChangeKarlWhichClaireReportsTo) {
...
+  {
+    // We should pass `claire_to_depend_on_update_ts` as `from` argument for proper resolving the dependencies.
```
dkorolev: ... for proper dependency resolution.

Karl/claire.h
```diff
@@ -439,7 +456,10 @@ class GenericClaire final : private DummyClaireNotifiable {
         }
       }
     } else if (r.method == "POST") {
-      if (r.url.query.has("report_to") && !r.url.query["report_to"].empty()) {
+      if (r.url.query.has("send_keepalive")) {
```
dkorolev: Perhaps `force_send_keepalive`.

Karl/claire.h
```diff
@@ -439,7 +456,10 @@ class GenericClaire final : private DummyClaireNotifiable {
...
+        ForceSendKeepalive();
+        r("OK\n");
```
dkorolev: Nit: If this is about to be used from outside the unit test, a chunked response with two chunks, `"Sending a forced keepalive to %s.\n"` and `"Done.\n"` could prove itself useful.

mzhurovich: What? `ForceSendKeepalive()` is non-blocking cv notify.

Karl/test.cc
mzhurovich: Just here or in all the tests?

Karl/test.cc
mzhurovich: Oops

Karl/test.cc
mzhurovich: `EXPECT_EQ` will print the actual count of elements, which I consider more informative. What do you think?

Karl/test.cc
mzhurovich: Do you suggest to rewrite all the tests?

Karl/test.cc
mzhurovich: How is it better?

Karl/test.cc
mzhurovich: And how could they collide if it's a static variable accessed from a single thread?

Karl/test.cc
dkorolev: How many more are there? If it's a minute of work ...

Karl/test.cc
dkorolev: Oh, right. Let's keep `EXPECT_EQ` then. It's the old Meyers' advice that `.size()` may be O(N) and is thus recommended to be avoided, khm-khm.

Karl/test.cc
dkorolev: Hmm. Maybe I wasn't fully awake when suggesting that. Within a single compilation unit, we should be good to go.

Karl/test.cc
dkorolev: Mayers' old cliche to avoid `size() > 0u`. Fine to ignore.

mzhurovich: I understand the idea of potential non-constant complexity of `.size()`, but this syntax looks less error-prone to me, as it's easier to forget `!` than write incorrect comparison.

mzhurovich:
@dkorolev Thanks and PTAL https://github.com/C5T/Current/pull/701/commits/8fff14d485a9bfd2dc4a68a4c921b91a7e5a3349

Karl/test.cc
dkorolev: :+1:

PR #702, opened 2016-12-15
FnCAS incremental improvements.

dkorolev:
Hi @mzhurovich,

* Returned fast-to-run non-JIT tests.
* Compiling all JIT implementation into the binary by default now.
* The JIT implementation to use can be selected by the template parameter.
* JIT-related template parameter can also be used to refer to non-JIT types.
* Minor refactoring to support the above.

Thanks,
Dima


dkorolev:
Travis is acting crazy today -- the failures are:

```
2.53s$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install g++-5 nasm gnuplot
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package g++-5
E: Couldn't find any package by regex 'g++-5'
```

Plus one where:

```

[ RUN      ] EventStore.SmokeWithInMemoryEventStore
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

FnCAS/fncas/differentiate.h
```diff
@@ -241,11 +248,21 @@ struct node_differentiate_impl<X> {
   }
 };
 
+// Helper code to enable JIT-based `f_impl`-s to be exposed as `fncas::function_t`.
```
mzhurovich: `g_impl` and `gradient_t`?

FnCAS/test.cc
```diff
@@ -67,63 +68,44 @@ static_assert(std::is_same<fncas::term_vector_t, std::vector<fncas::term_t>>::va
...
   {
     // grep for `42.000`.
     const fncas::variables_vector_t x(1);
-    const fncas::function_blueprint_t intermediate_function([](const fncas::term_vector_t& x) { return x[0] + 42; }(x));
+    const fncas::function_t<fncas::JIT::Blueprint> intermediate_function(
+        [](const fncas::term_vector_t& x) { return x[0] + 42; }(x));
     EXPECT_EQ(1042, intermediate_function({1000}));
 
-    const fncas::function_compiled_t compiled_function(intermediate_function);
+    const fncas::function_t<fncas::JIT::AS> compiled_function(intermediate_function);
```
mzhurovich: Shouldn't we use `JIT::Default` everywhere, except the tests specific for the particular JIT type?

FnCAS/fncas/base.h
```diff
@@ -64,7 +64,23 @@ enum class type_t : uint8_t { variable, value, operation, function };
 enum class operation_t : uint8_t { add, subtract, multiply, divide, end };
 enum class function_t : uint8_t { sqr, sqrt, exp, log, sin, cos, tan, asin, acos, atan, unit_step, ramp, end };
 
+enum class JIT {
+  Super,          // The superclass of all `function_t<>`-s and `gradient_t<>`-s respectively.
+  Blueprint,      // The blueprint of the function of gradient, evaluated by traversing an internal tree.
+  NativeWrapper,  // A wrapper over an `std::function<double(std::vector<double>&)>`, or approximate gradient computer.
```
mzhurovich: Technically speaking, these three are not "JIT types"...

FnCAS/fncas/differentiate.h
dkorolev: :+1: 7bfc8fb

FnCAS/test.cc
dkorolev: :+1: 7bfc8fb

FnCAS/fncas/base.h
dkorolev: Indeed they are not. I was hoping you'd suggest a better name for this `enum class`. :-)

dkorolev:
@mzhurovich Thx, PTAL @ 7bfc8fb, and the renaming of the `JIT` enum class is on my list, suggestions welcome.

FnCAS/fncas/base.h
```diff
@@ -64,7 +64,23 @@ enum class type_t : uint8_t { variable, value, operation, function };
 enum class operation_t : uint8_t { add, subtract, multiply, divide, end };
 enum class function_t : uint8_t { sqr, sqrt, exp, log, sin, cos, tan, asin, acos, atan, unit_step, ramp, end };
```
mzhurovich: Just realized that there are two different `function_t`-s: this one, and the one defined in `node.h` :-/
I also think that it's not correct to use `_t` notation for `enum class`-es, since they are not type aliases. What do you think about using the following names for the three declarations above: `Operand`, `MathOperation`, `MathFunction`?

mzhurovich:
@dkorolev `JIT-related template parameter can also be used to refer to non-JIT types` - what is the use-case for it? For now it looks for me like a "strong typing violation", when theoretically you can pass a syntactically correct, but meaningless template parameter.

FnCAS/fncas/node.h
```diff
@@ -405,7 +405,7 @@ struct X : std::vector<V>, noncopyable {
 };
 
 // Class "f_super" is the placeholder for function evaluators.
-// One implementation -- f_intermediate -- is provided by default.
+// Two implementation -- f_impl<JIT::NativeWrapper> and f_impl<JIT::Blueprint> -- are provided in this header.
```
dkorolev: :+1: 187ab5e

FnCAS/fncas/base.h
dkorolev: Auch :-)

Thx for the catch. `NodeType`, `MathOperation`, `MathFunction` in 187ab5e.

dkorolev:
@mzhurovich Good catch, PTAL.

Re. the "JIT-related" template parameter -- I don't have a clear picture yet, just had a hunch one exposes type, with a template parameter, would appear user-friendlier to an external FnCAS user than the type hierarchy. I may well be wrong.

mzhurovich:
@dkorolev Ok, I'm merging this and let's discuss `JIT` parameter later.

PR #703, opened 2016-12-16
A CSV parser.

dkorolev:


Utils/CSV/csv.h
```diff
@@ -0,0 +1,89 @@
...
+      CURRENT_THROW(CSVFileNotFoundException("The CSV file `" + filename + "` could not be opened."));
+    }
+    std::string line;
+    if (!std::getline(fi, line)) {
+      CURRENT_THROW(CSVFileFormatException("The CSV file `" + filename + "` is empty."));
+    }
+    if (!line.empty() && line.back() == '\r') {
+      line.resize(line.size() - 1u);  // NOTE(dkorolev): This may look like a hack, but I'd rather stay safe.
+    }
+    csv.header = current::strings::Split(line, ',');
```
sompylasar: Don't forget quoted values.

"Some,comma,separated,title","Second"

dkorolev:
I've intentionally forgotten them. Is there a CSV standard that includes
quoting rules though?

sompylasar:
@dkorolev https://tools.ietf.org/html/rfc4180#section-2

> While there are various specifications and implementations for the
   CSV format (for ex. [4], [5], [6] and [7]), there is no formal
   specification in existence, which allows for a wide variety of
   interpretations of CSV files.  This section documents the format that
   seems to be followed by most implementations:

dkorolev:
I suggest we deal with the issue as it shows up.

sompylasar:
@dkorolev And the full spec requires linear parser, split-based won't do.

   6.  Fields containing line breaks (CRLF), double quotes, and commas
       should be enclosed in double-quotes.  For example:

       "aaa","b CRLF
       bb","ccc" CRLF
       zzz,yyy,xxx

sompylasar:
> I suggest we deal with the issue as it shows up.

Sure, as you wish.

dkorolev:
Makes my suggestion not the most plausible, but the only plausible, one. :-)

mzhurovich:
Ok, it's super-naive implementation to satisfy our short-term needs :) Further we may consider another design with per-line parsing, checking the consistency of value count, supporting optional header, quoted values etc.

PR #704, opened 2016-12-24
File exceptions to include file/dir names.

dkorolev:
Hi @grixa -- I figured this would be useful while running your code.
CC @mzhurovich.


grixa:
Thanks!
LGTM

PR #709, opened 2016-12-29
ML-aware Irises demo.

dkorolev:
Yo!

After https://github.com/C5T/Current/pull/708.

Check out this visualization from Wikipedia: https://upload.wikimedia.org/wikipedia/commons/5/56/Iris_dataset_scatterplot.svg

Here's the output of `step2`, `http://localhost:3000`:

image

And here's the output of `step3`, `http://localhost:3001`:

image

All 24 plots are clickable.

Would appreciate it if you could test this code on Mac.

Thanks!
Dima

PR #710, opened 2017-01-03
Iris ML probability histograms visualization.

dkorolev:
Hi Max,

After https://github.com/C5T/Current/pull/709, here's what I came up with for precision visualization.

Descriptive model only:

descriprive-only

descriptive-only-training

descriptive-only-accuracy

Descriptive and discriminant:

descriptive-and-discriminant

descriptive-and-discriminant-training

descriptive-and-discriminant-accuracy

For the lower right image, here's the result of the full model:

image

And here's only the descriptive model's part:

image

### DIFF
--- a/FnCAS/fncas/optimize.h
+++ b/FnCAS/fncas/optimize.h
@@ -58,32 +58,27 @@ CURRENT_STRUCT_T(ObjectiveFunctionValue) {
     return *this;
   }
 };
+// clang-format on
 
-template<typename T>
+template <typename T>
 struct ObjectiveFunctionValueExtractorImpl {
   using value_t = typename T::value_t;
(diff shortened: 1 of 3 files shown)


FnCAS/fncas/optimize.h
```diff
@@ -49,18 +49,45 @@ namespace optimize {
...
-  operator T() const {
-    return value;
-  }
   ObjectiveFunctionValue<T>& AddPoint(const std::string& name, T value) {
     extra_values[name] = value;
     return *this;
   }
 };
 
+template<typename T>
```
mzhurovich: `make indent`?

dkorolev: :+1: 243442c

PR #712, opened 2017-01-31
Nested `Variant`-s.

dkorolev:


port.h
```diff
@@ -231,10 +231,15 @@ inline void CURRENT_ASSERTION_FAILED(const char* text, const char* file, int lin
 #define CURRENT_ASSERT(expr) ((expr) ? static_cast<void>(0) : CURRENT_ASSERTION_FAILED(#expr, __FILE__, __LINE__))
 #endif
 
-// The `FEWER_COMPILE_TIME_CHECKS` macro, when set, moves more checks from compile time to run time.
-// Pro: Faster and less RAM-demanding compilation. Con: Some type errors may go unnoticed.
-// TODO(dkorolev)+TODO(mzhurovich): We'd like to bind this to `!NDEBUG`, but let's keep it on by default for now.
-#define FEWER_COMPILE_TIME_CHECKS
+// The `VARIANT_CHECKS_AT_RUNTIME_INSTEAD_OF_COMPILE_TIME` macro, when set, moves `Variant` type checks
+// // from compile time to run time.
```
mzhurovich: Extra `//`

TypeSystem/variant.h
```diff
@@ -109,20 +123,20 @@ struct RuntimeTypeListHelpers<TypeListImpl<TS...>> {
...
+  virtual void UncheckedMoveFromUniquePtr(std::unique_ptr<current::variant::InnerType>) = 0;
 };
 
 // Note: `Variant<...>` never uses `TypeList<...>`, only `TypeListImpl<...>`.
 // Thus, it emphasizes performance over correctness.
 // The user hold the risk of having duplicate types, and it's their responsibility to pass in a `TypeList<...>`
 // instead of a `TypeListImpl<...>` in such a case, to ensure type de-duplication takes place.
 
-// Initializes an `std::unique_ptr<CurrentStruct>` given the object of the right type.
+// Initializes an `std::unique_ptr<urrent::variant::InnerType>` given the object of the right type.
```
mzhurovich: s/<urrent/<current/

port.h
dkorolev: :+1: 2acd1c4

TypeSystem/variant.h
dkorolev: :+1: 2acd1c4

dkorolev:
@mzhurovich PTAL

TypeSystem/variant.h
```diff
@@ -50,6 +54,16 @@ SOFTWARE.
 
 namespace current {
 
+namespace variant {
+#ifdef VARIANT_CHECKS_AT_RUNTIME_INSTEAD_OF_COMPILE_TIME
+// When variant inner type checks are run-time, no nested variants are allowed.
+using InnerType = CurrentStruct;
```
mzhurovich: What do you think about `InnerType` -> `object_super_t` or `object_base_t`?

dkorolev: I'd call it this way had it been an in-class type, but it's a namespace-only one.

Think it's OK to just have `current::variant::object_base_t`?

mzhurovich: If you prefer camel-case for this reason, what about `InnerBaseType` or `InnerSuperType`? I'd like to specify explicitly that it's not the inner type of the contained object, but the base type.

dkorolev: Let me do `current::variant::object_base_t` and see if it looks good.

dkorolev:
@mzhurovich PTAL @ f26535f.

TypeSystem/variant.h
```diff
@@ -239,7 +265,7 @@ struct VariantImpl<NAME, TypeListImpl<TYPES...>> : CurrentVariantNameHelper<NAME
   }
 
   template <typename X>
-  std::enable_if_t<!std::is_same<X, CurrentStruct>::value, X&> VariantValueImpl() {
+  std::enable_if_t<!std::is_same<X, current::variant::object_base_t>::value, X&> VariantValueImpl() {
```
mzhurovich: Hmm. How are these `enable_if`-s for `VariantValueImpl` & `VariantExistsImpl` are supposed to work?

dkorolev: Not sure I understand the question.

If it's about `enable_if` being part of the return type, not template declaration, I assume as there's no overloading involved, the current implementation would do.

If it's about the very presence of `enable_if`, it's to prohibit using the `object_base_t` as the type for `Exists<>` / `Value<>`.

dkorolev:
@mzhurovich : The code did the job it did as the top-level `current::{Exists/Value}` already has logic for `{Exists/Value}<T>([const] T&)`.

I've added a few test statements, and removed the apparently unused logic from `variant.h`. PTAL @ e7950f883256bc320d021de7f51efb141e60e708.

PR #719, opened 2017-02-12
A prototype of Storage::CQRS.

dkorolev:
Hi @mzhurovich,

This is a hack which should not be merged yet. Just wanted to get your feedback.

* * *
TL;DR:
* Added `/command/*` and `/query/*` to Storage.
* String-based dispatching: the `*`-s above are to be replaced by the name of the command or query, explicitly specified by the user.
* These strings are journaled as transaction meta fields (which is a plain map-string-string for now).
* As discussed, did **not** do:
  * structured objects in transaction meta fields, because a) compilation time would skyrocket, and b) evolution would be a pain.
  * compile-time CQRS "endpoints" registration (we could, but that would only save one line of user code at the expense of, again, slowing down compilation).

* * *

Remains:
* Test exceptions in user code, in both queries and commands.
* Implement and test rollbacks from within `/command/*`-s (will most likely fail now due to the wrong exception).
* Need to add URL parsing (for GET `/query/*`-ies with no or just a few simple parameters).
* Need to add the test for transaction meta fields being correctly journaled.
* Need to make sure Commands require a mutating HTTP verb: "GET" or "HEAD" are "Method Not Allowed" for `/command/*` (however, a question: should `/command/*` always require JSON body, or can we live with URL parameters for simple commands?)
* Refactoring, specifically when it comes to the place of parsing the body (or the URL), as now it's a plain string copy. Also, there's a copy-pasted section when initializing queries and then command. Finally, the "empty fields struct" check is runtime now, while I think it should be compile time.

* * *

Note: Will need to not forget to nginx-route `/query/` to followers, as, even though they are not necessarily GET/HEAD requests, they are, by design, eventually consistent.

(Perhaps expose some `/sync_query/*` as well, for strongly consistent queries?)

* * *

Edit, from my offline notes:

1. Enumerate CQRS endpoints? Ex. return a list via `/query` and `/command`?
2. Export CQRS endpoints' schema? Could begin with F# for now.
3. Don't forget to turn CQRS endpoints into 503s as the storage enters the graceful shutdown mode.


Storage/api_types.h
```diff
@@ -34,10 +34,14 @@ namespace current {
 namespace storage {
 namespace rest {
 
+// Top-level URL prefixes for Storage REST.
 const std::string kRESTfulDataURLComponent = "data";
 const std::string kRESTfulSchemaURLComponent = "schema";
+const std::string kRESTfulCQRSCommandURLComponent = "command";
+const std::string kRESTfulCQRSQueryURLComponent = "query";
```
mzhurovich: I vote for the second level here: "cqrs/command", "cqrs/query".
Top-level "query" may be used for the future generic query engine.

Blocks/URL/url.h
```diff
@@ -254,6 +257,55 @@ struct URLParametersExtractor {
...
+
+      // For the remaining field types, use `ParseJSON`. Overkill, but ensures any `CURRENT_STRUCT` can be URL-encoded.
+      template <typename T>
+      void operator()(const std::string& key, T& value) const {
+        const auto cit = q.find(key);
+        if (cit != q.end()) {
+          try {
+            ParseJSON(cit->second, value);
+          } catch (const current::TypeSystemParseJSONException&) {
+            // Silently discard the exception.
```
mzhurovich: Shouldn't this eventually result in `BadRequest`?

Blocks/URL/test.cc
```diff
@@ -265,3 +267,60 @@ TEST(URLTest, EmptyURLException) {
...
+  CURRENT_FIELD(m, (std::map<std::string, std::string>));
+};
+
+}  // namespace url_test
+
+TEST(URLTest, FillsCurrentStructsFromURLParameters) {
+  using namespace url_test;
+
+  {
+    const auto simple = URL("/simple?a=1&b=2&s=test with spaces").query.template FillObject<Simple>();
```
dkorolev: The constructor of Current's `URL` class will handle it. (I.e., if you `ComposeURL` this URL back, it'll escape those.)

I suggest we keep it this way.

Blocks/URL/test.cc
```diff
@@ -265,3 +267,60 @@ TEST(URLTest, EmptyURLException) {
...
+
+  {
+    Tricky tricky;
+
+    URL("/tricky?s=foo").query.FillObject(tricky);
+    ASSERT_TRUE(Exists(tricky.s));
+    EXPECT_EQ("foo", Value(tricky.s));
+
+    URL("/tricky?s=").query.FillObject(tricky);
+    ASSERT_TRUE(Exists(tricky.s));
```
dkorolev: :+1: eedb7b7

dkorolev: :+1: ref. eedb7b7

Storage/api.h
```diff
@@ -440,22 +456,79 @@ class RESTfulStorage {
...
+                  return dynamic_cast<COMMAND_IMPL&>(*type_erased_command.get())
+                      .template Command<MutableFields<STORAGE_IMPL>>(fields, url);
+                } else {
+                  return COMMAND_IMPL().template Command<MutableFields<STORAGE_IMPL>>(fields, url);
+                }
+              } catch (const TypeSystemParseJSONException& e) {
+                return Response(cqrs::CQRSParseJSONException(e.What()), HTTPResponseCode.BadRequest);
+              }
+              // TODO(dkorolev): Test transaction rollbacks.
+              return Response(cqrs::CQRSBadRequest(), HTTPResponseCode.BadRequest);
```
dkorolev: The code will get here if the `try` blocks throws an exception which is not `TypeSystemParseJSONException`, and not the one inherited from `TypeSystemParseJSONException`.

Storage/api.h
```diff
@@ -480,11 +553,119 @@ class RESTfulStorage {
...
 
+  template <typename T>
+  static std::shared_ptr<CurrentSuper> ParseCQRSRequest(const Request& request) {
+    // TODO(dkorolev): Proper exception types here!
+    auto object = std::make_shared<T>();
+    if (reflection::FieldCounter<T>::value > 0) {
+      if (!request.body.empty()) {
+        ParseJSON(request.body, *object);
+      } else {
+        request.url.query.FillObject(*object);
```
dkorolev: My plan for now (this PR is still work in progress) is:

1) If the command/query class is empty, don't even attempt to parse anything.
2) Otherwise, if the body is not empty (for both commands and queries, no matter the HTTP verb), parse the body as a JSON of the respective schema.
3) Otherwise, look into whether URL parameters are present.

HTTP verb should not be GET/HEAD for commands.

Storage/api.h
```diff
@@ -480,11 +553,119 @@ class RESTfulStorage {
...
+                                        [&storage, &f_run_query, handler, generic_input, type_erased_query](
+                                            immutable_fields_t fields) -> Response {
+                                          return f_run_query(
+                                              fields, std::move(type_erased_query), generic_input.restful_url_prefix);
+                                        },
+                                        std::move(request)).Detach();
+                          });
+          } catch (const Exception& e) {
+            // An exception has occurred while parsing user request.
+            request(Response("TODO(dkorolev): Error message here. Or reuse the `BadJSON` handler.",
```
dkorolev: This code is still work in progress. :)

Storage/rest/plain.h
```diff
@@ -293,6 +293,14 @@ struct Plain {
     }
   };
 
+  template <typename STORAGE>
+  struct RESTfulCQRSHandler {
+    template <typename F>
+    void Enter(Request request, F&& next) {
```
dkorolev: Yes, a convention for Current Storage. All RESTful handlers have `Enter`. @nyospe's, closed-source, secured implementation of REST overrides these `Enter`-s to check auth, and I concluded the least error-prone way would be to make CQRS endpoints securable in the same fashion,

Storage/api.h
```diff
@@ -440,22 +456,79 @@ class RESTfulStorage {
...
+
+  template <class QUERY_IMPL>
+  void AddCQRSQuery(const std::string& query) {
+    std::lock_guard<std::mutex> lock(data_->cqrs_handlers_mutex_);
+    // TODO(dkorolev): Exception if the handler is already registered.
+    data_->cqrs_query_map_[query] = std::make_pair(
+        [](const Request& request) -> std::shared_ptr<CurrentSuper> { return ParseCQRSRequest<QUERY_IMPL>(request); },
+        [query](immutable_fields_t fields, std::shared_ptr<CurrentSuper> type_erased_query, const std::string& url)
+            -> Response {
+              return dynamic_cast<QUERY_IMPL&>(*type_erased_query.get())
```
dkorolev: :+1: 4043309

Storage/api_types.h
dkorolev: Perhaps `/index/` for the generic query engine? You guys convinced me what we're doing is not really CQRS :)

Blocks/URL/url.h
dkorolev: I'm afraid it would make the logic trickier -- there's no simple way for URL parsing to trigger an error.

In my head, passing complex data in URL parameters is a bad style that's discouraged, and, under this assumption, the user who's doing that knows what are they doing. Keep in mind, this method extends the object, doesn't fully parse it.

Storage/api.h
mzhurovich: It will not get here as the exception won't be caught.

Storage/api_types.h
mzhurovich: No, definitely not `index`, it's very ambiguous.

mzhurovich:
@dkorolev btw, why `shared_ptr<CurrentSuper>`, not `CurrentStruct`?

Storage/api_types.h
dkorolev: Let me add `/cqrs/` in a separate commit, as I'm done with present refactoring.

Storage/api.h
dkorolev: Oh, /me stupid. Looking into.

Blocks/URL/test.cc
sompylasar: > The constructor of Current's `URL` class will handle it. (I.e., if you `ComposeURL` this URL back, it'll escape those.)
> 
> I suggest we keep it this way.

👌

Storage/api.h
sompylasar: So, your plan allows to pass the same command payload as a query string or as body, to the discretion of the caller. This may result in sensitive data (that is I assume mostly passed via body) to get into query string.

Storage/api.h
sompylasar: Sure, I understand that, though I think keeping the "public" part of it clean of "internals" such as TODOs is a good habit even in WIP code.

Storage/rest/plain.h
sompylasar: Cool.

Storage/api_types.h
sompylasar: > CQS prescribes separate methods for querying from or writing to a model: the query doesn't mutate state, while the command mutates state but does not have a return value.
> 
> CQRS prescribes a similar approach, except it's more of a path through your system. A query request takes a separate path from a command. The query returns data without altering the underlying system; the command alters the system but does not return data.

http://stackoverflow.com/questions/34255490/difference-between-cqrs-vs-cqs

mzhurovich: @sompylasar thnx, CQS looks like the best fit for now.

Storage/api.h
mzhurovich: @sompylasar AFAIK passing the body in GET requests, though possible, is not recommended by HTTP/1.1 and may cause problems with query caching. GraphQL HTTP interface, for example, uses the `query` URL query parameter for the query + `variables` for variables.

sompylasar: @mzhurovich Yes, but passing a query string in a POST is perfectly valid. What I am thinking about here is removing ambiguity about where to pass the payload: limit query string to GET and body to POST (no query in POST).

Storage/api.h
sompylasar: fb_img_1487220489641

Blocks/URL/url.h
mzhurovich: This is not a generic URL parsing process, this is a special method to be used for specific purposes. If  `TypeSystemParseJSONException` is not caught here, it should easily propagate to the CQRS handler that will respond with `Bad JSON`, which, I think, is good. Otherwise we risk to accidentally get crappy results for no obvious reason, because we silently ignore parameters parsing errors.

dkorolev:
Hi @mzhurovich @sompylasar!

Please take another look:
 
* Refactored and cleaned up the logic of URL+JSON parsing as per @sompylasar's request. 
* Cleaned up TODOs and exceptions.
* Made sure rollbacks are handled fine in commands.
* Made sure exceptions are handled fine in queries and commands.

Remains:

* Top-level `/cqs/`. @mzhurovich -- let's chat about this on the next call.
* Method not allowed for GET/HEAD on commands.
* Turn CQRS endpoints into 503s.
* CQRS endpoints enumeration.
* CQRS endpoints schema.

dkorolev:
Naming-wise, basically, I'm concerned CQ[R]S seem to almost explicitly require commands to not return any values. That looks odd to me, as the only true way to ensure atomicity is to have a [custom, user-defined] command return the result as it has been specifically as per the command has been run.

This alone makes me wonder whether `/cq[r]s/` is the right URL component.

sompylasar:
Not returning value from a command makes sense if the client is reactive, so that it gets a notification of the backend state changes through some persistent channel when they actually happen (this can be deferred from the command). The queries are still needed though to initialize the client state from the ground up.

Moreover, if a command affects multiple entities (and this set may differ depending on the command parameters and the current state of the backend) which one or many and in which structure will you return?

The backend state though can be sliced into logical pieces (closely related entities, maybe transformed to suit the client needs), and it's convenient to have the backend return that piece of the backend state from both "get state" query and all the commands that are dedicated to this piece of state.

sompylasar:
@dkorolev Reviewed, comments in commits.

Blocks/URL/url.h
dkorolev: The current version of the code does it the way you guys asked for :)

Storage/api_types.h
dkorolev: +1

Even though we return data from commands. Because not doing so is, well, not smart. :)

dkorolev:
@sompylasar, thx.

Re. `FromString<bool>` and URL parsing -- yes, it doesn't throw, and treats anything but explicit false (i.e., `0`, `false`, `False`, `FALSE`) as true.

Re. returning the exception in the boilerplate CQS response: I think returning exception type, file, and line can't really compromise any data, while simplifying debugging greatly. Besides, any production-ready code would not rely on this behavior. Also, I'd assume eventually this logic will also be wrap-pable into custom logic, so that, much like authentication, generic CQS responses could be tweaked. What do you think? 

@mzhurovich -- working on taking care of the remaining big picture pieces. As discussed offline, will leave the refactoring of `Storage/exceptions.h` part to later.

dkorolev:
@mzhurovich -- while making the changes @sompylasar asked for in `current::Exception` logic, I f*cked up exception error message generation. Will take a look tomorrow.

The CQS part should be good to go unit-test-wise as per our conversations. Refactoring of its own `exceptions.h` will come later.

PTAL && thanks!

dkorolev:
Yo! @mzhurovich @sompylasar PTAL.

sompylasar:
@dkorolev What is the idea of `What` and `OriginalWhat`? I couldn't understand.

dkorolev:
`What()` is full description, similar to C++ standard exceptions. We augment it with other details: exception details, file, line, and the very statement that did throw it.

`OriginalWhat()` keeps exception details only.

Prior to this PR we used to initialize `what_` and append to it. With this PR the data is stored in different member variables.

(Thus, one of `what_,original_what_` members is now unnecessary. Good point.)

dkorolev:
OK, looks like I've made it most of the way w/ the tests and exceptions.

The last remaining bit is the compilation error on Apple with Storage transaction response (?) triggering an attempt to JSON-serialize an unserializable type. I have a guess that type is `size_t`, and am trying to cast it into `uint64_t` in the very last commit.

Will resume my research tomorrow.

Thanks,
Dima

dkorolev:
This one is production-ready. PTAL.

@mzhurovich (have a safe flight), @sompylasar.

Thanks!
Dima

sompylasar:
:+1:

Blocks/URL/test.cc
```diff
@@ -265,3 +267,183 @@ TEST(URLTest, EmptyURLException) {
...
+    SimpleWithOptionals object;
+    // These values won't be touched.
+    object.b = 10000;
+    object.t = "bar";
+    URL("/test?a=42&s=foo")
+        .query.template FillObject<SimpleWithOptionals, current::url::FillObjectMode::Strict>(object);
+    EXPECT_EQ(42, object.a);
+    EXPECT_EQ("foo", object.s);
+    EXPECT_TRUE(Exists(object.b));
+    EXPECT_TRUE(Exists(object.t));
```
mzhurovich: `ASSERT_TRUE` before testing the `Value`s?

Bricks/exception.h
```diff
@@ -38,20 +39,45 @@ class Exception : public std::exception {
...
+    static char data[MAX_LENGTH + 1];
+    const std::string message = DetailedDescription();
+    if (message.length() < MAX_LENGTH) {
+      std::strcpy(data, message.data());
+    } else {
+      std::copy(message.begin(), message.begin() + MAX_LENGTH, data);
+      data[MAX_LENGTH] = '\0';
+    }
+    return data;
+  }
```
mzhurovich: So, `what()` and `What()` now return different things :-/
IMHO it becomes really not obvious, how to use this combination without consulting with the source code every time one needs to log an exception.

dkorolev:
Hi @mzhurovich,

I've (manually) merged your code with the current `master`of `Kernel`, build&run the tests, and they passed. The post-merge code is pushed to the `nosql_event_store_2_fix_by_dkorolev` branch of `RealtyShares/kernel`.

It also depends on https://github.com/C5T/Current/pull/722.

RealtyShares/kernel branch: https://github.com/RealtyShares/kernel/compare/nosql_event_store_2_fix_by_dkorolev?expand=1

RealtyShares/kernel commit: https://github.com/RealtyShares/kernel/commit/6d3e8d29c9d0287627c2ee0249ed1811484e1d6e

dkorolev:
Moving to https://github.com/C5T/Current/pull/723.

PR #720, opened 2017-02-13
Extending Current iterators to be more STL-compliant.

mzhurovich:
* STL-compliant type aliases exposed to support standard iterator operations like `std::advance`
* Added prefix and postfix `++`/`--` operators, return type changed to the iterator type instead of `void`
* Added reverse iterator class
* Added `rbegin()`/`rend()` + `lower_bound`/`upper_bound` wrappers for ordered maps in `GenericMapAccessor`

@dkorolev What did you want to change in naming? `// DIMA_FIXME: Naming.`
Although our Accessor and Iterators are de-facto `const`, I've decided not to rename them, as `GenericMapConstReverseIterator` looks a bit long. However, suggestions for the further improvements of these interfaces are welcome.
CC @grixa

### DIFF
--- a/Bricks/util/iterator.h
+++ b/Bricks/util/iterator.h
@@ -29,6 +29,7 @@
 #include "../template/is_unique_ptr.h"
 
 namespace current {
+namespace stl_wrappers {
 
 template <typename MAP>
 struct GenericMapIterator final {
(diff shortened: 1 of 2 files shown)


dkorolev:
LGTM++

Can't really recall what my concern with naming has been, but the `CamelCase` vs. STL naming bit begins to bug me. Perhaps we could make a move to same-as-STL naming, ex. `empty()`/`size()`/`lower_bound(...)`/etc.?

CC @nyospe

PR #721, opened 2017-02-24
Fix for Windows' `.tellp()`.

dkorolev:
Thanks @amarmer @nyospe, CC @mzhurovich @grixa.

TL;DR: On Windows the `std::ofstream` open with `std::ofstream::app` and w/o `std::ofstream::ate` doesn't automatically seek to the end of file for `.tellp()` purposes, resulting in an inconsistent index exception when publishing to the stream.

Thanks,
Dima


mzhurovich:
Hmm, Travis still sometime fails on Karl's test :-/

dkorolev:
Should this stop us from merging a pull request unrelated to Travis? :)

PR #722, opened 2017-03-08
Not shooting ourselves into the foot.

dkorolev:
Hi @mzhurovich -- making sure our diverged branches don't diverge too far.

No need to merge yet, as the CQS exceptions header file still contains "REST"-ful HTTP responses logic.

This unblocks your code in another repo.

Thanks,
Dima


TypeSystem/Reflection/test.cc
```diff
@@ -327,6 +327,18 @@ struct CollectFieldValues {
...
+  CURRENT_FIELD(b, size_t);
+};
+
+static_assert(std::is_same<decltype(std::declval<StructWithSizeTWhichShouldBeUInt64T>().a), uint64_t>::value, "");
+static_assert(std::is_same<decltype(std::declval<StructWithSizeTWhichShouldBeUInt64T>().b), uint64_t>::value, "");
+
+static_assert(std::is_same<decltype(std::declval<StructWithSizeTWhichShouldBeUInt64T>().a),
+                           decltype(std::declval<StructWithSizeTWhichShouldBeUInt64T>().b)>::value,
+              "");
+
```
mzhurovich: What's the purpose of this test? `size_t` is platform dependent by design and should never be used in `CURRENT_FIELD`.

TypeSystem/struct.h
```diff
@@ -96,6 +96,13 @@ struct FieldImpl<DF, T> {
   typedef T type;
 };
 
+// Allow `size_t` in JSON-serializable types, but make it `uint64_t` right way.
```
mzhurovich: Why do wee need it?

TypeSystem/Reflection/test.cc
mzhurovich: Oh, I see, I've re-mapped `size_t` inside the `CURRENT_STRUCT` implementation. Not really obvious decision imo :-/ Let's chat about it offline.

dkorolev:
Let's move to https://github.com/C5T/Current/pull/723.

PR #723, opened 2017-03-10
Converging (a follow-up to "A prototype of Storage::CQRS").

dkorolev:
Hi @mzhurovich,

This is as close as I got to make the CQS, URL-parameters, and `current::Exception`-related changes ready to me merged.

As discussed offline, the generation of the CQS rollback exception still is a challenge. I've worked around it with a poor man's attempt to see whether certain type can or can not be JSON-serialized.

After thinking over it, I don't really like the idea of introducing another, CQS-friendly, rollback semantics. Suggest we merge it as is for now.

Thanks,
Dima


Bricks/exception.h
```diff
@@ -35,32 +36,52 @@ namespace current {
...
+    line_ = line;
+  }
 
  private:
-  std::string what_;
+  // The user-defined description of the exception.
+  std::string original_description_;
+
+  // The description augmented with the full body of the invokation of `CURRENT_THROW`,
+  // as well as `__FILE__` and `__LINE__`. This is what is returned by `what()`, `overriding std::exception::what()`.
```
mzhurovich: But it's not augmented anywhere :(

mzhurovich:
@dkorolev LGTM! Thanks!

PR #729, opened 2017-05-18
Removed unobvious "offset + 2" conditions from the GenericHTTPRequestData

grixa:
They useless now and just make the code less clear.


dkorolev:
Merged as part of https://github.com/C5T/Current/pull/730. Thank you Grixa!

PR #737, opened 2017-05-30
ScanDirRecursive

sompylasar:
@mzhurovich @grixa @dkorolev 

Please take a look.

The gist: 
- `ScanDirItemInfo` replaces `std::string filename` to pass additional fields to the `ScanDirUntil` and `ScanDir` callbacks.
- `ScanDirUntil` and `ScanDir` takes `ScanDirRecursive::Yes` to scan directories recursively.

Please also suggest how to minimize copying around `ScanDirItemInfo` (the struct itself and its contents: `name` and `path` strings). I believe this may need improvement. _My naive usage of references led to segmentation fault and reads of uninitialized memory within the tests._

Thanks!

### DIFF
--- a/Bricks/file/file.h
+++ b/Bricks/file/file.h
@@ -253,9 +253,8 @@ struct FileSystem {
           const char* const name = find_data.cFileName;
           if (ScanDirCanHandleName(name)) {
             const bool is_directory = (find_data.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY);
-            const ScanDirParameters mask = is_directory
-                                             ? ScanDirParameters::ListDirsOnly
-                                             : ScanDirParameters::ListFilesOnly;
+            const ScanDirParameters mask =
+                is_directory ? ScanDirParameters::ListDirsOnly : ScanDirParameters::ListFilesOnly;
             if (static_cast<int>(parameters) & static_cast<int>(mask)) {
               if (!item_handler(ScanDirItemInfo(name, JoinPath(directory, name), is_directory))) {
                 return;
@@ -300,24 +299,22 @@ struct FileSystem {
       }
 #endif
     } else {
-      const std::function<bool(const ScanDirItemInfo&)> item_handler_recursive = [&item_handler, parameters](const ScanDirItemInfo& item_info) {
-        const ScanDirParameters mask = item_info.is_directory
-                                         ? ScanDirParameters::ListDirsOnly
-                                         : ScanDirParameters::ListFilesOnly;
-        if (static_cast<int>(parameters) & static_cast<int>(mask)) {
-          if (!item_handler(item_info)) {
-            return false;
-          }
-        }
-        if (item_info.is_directory) {
-          ScanDirUntil<ITEM_HANDLER>(item_info.pathname, std::forward<ITEM_HANDLER>(item_handler), parameters, ScanDirRecursive::Yes);
-        }
-        return true;
-      };
-      ScanDirUntil(directory,
-                   item_handler_recursive,
-                   ScanDirParameters::ListFilesAndDirs,
-                   ScanDirRecursive::No);
+      const std::function<bool(const ScanDirItemInfo&)> item_handler_recursive =
+          [&item_handler, parameters](const ScanDirItemInfo& item_info) {
+            const ScanDirParameters mask =
+                item_info.is_directory ? ScanDirParameters::ListDirsOnly : ScanDirParameters::ListFilesOnly;
+            if (static_cast<int>(parameters) & static_cast<int>(mask)) {
+              if (!item_handler(item_info)) {
+                return false;
+              }
+            }
+            if (item_info.is_directory) {
+              ScanDirUntil<ITEM_HANDLER>(
+                  item_info.pathname, std::forward<ITEM_HANDLER>(item_handler), parameters, ScanDirRecursive::Yes);
+            }
+            return true;
+          };
+      ScanDirUntil(directory, item_handler_recursive, ScanDirParameters::ListFilesAndDirs, ScanDirRecursive::No);
     }
   }
 
--- a/Bricks/file/file.h
+++ b/Bricks/file/file.h
@@ -225,89 +225,113 @@ struct FileSystem {
   };
 
   enum class ScanDirParameters : int { ListFilesOnly = 1, ListDirsOnly = 2, ListFilesAndDirs = 3 };
+  enum class ScanDirRecursive { No, Yes };
 
   static inline bool ScanDirCanHandleName(const char* const name) {
     return (*name && ::strcmp(name, ".") && ::strcmp(name, ".."));
   }
 
-  template <typename F>
+  template <typename ITEM_HANDLER>
   static inline void ScanDirUntil(const std::string& directory,
-                                  F&& f,
-                                  ScanDirParameters parameters = ScanDirParameters::ListFilesOnly) {
+                                  ITEM_HANDLER&& item_handler,
+                                  ScanDirParameters parameters = ScanDirParameters::ListFilesOnly,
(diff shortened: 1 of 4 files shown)


sompylasar:
Oh, and I would greatly appreciate a test run on Windows (at least of `Bricks/file`).

Bricks/file/file.h
```diff
@@ -214,79 +214,125 @@ struct FileSystem {
   // TODO(dkorolev): Make OutputFile not as tightly coupled with std::ofstream as it is now.
   typedef std::ofstream OutputFile;
 
+  struct ScanDirItemInfo {
+    const std::string name;
+    const std::string path;
+    const bool is_directory;
```
mzhurovich: Usually, the members of such structures don't have `const` qualifiers and the const-ness is controlled for the particular instance of a structure.

Bricks/file/file.h
```diff
@@ -214,79 +214,125 @@ struct FileSystem {
   // TODO(dkorolev): Make OutputFile not as tightly coupled with std::ofstream as it is now.
   typedef std::ofstream OutputFile;
 
+  struct ScanDirItemInfo {
+    const std::string name;
+    const std::string path;
```
mzhurovich: I'd propose `basename`/`pathname` naming convention here.
https://en.wikipedia.org/wiki/Basename
@dkorolev what do you think?

Bricks/file/file.h
```diff
@@ -214,79 +214,125 @@ struct FileSystem {
...
+    const bool is_directory;
+
+    ScanDirItemInfo() = delete;
+    ScanDirItemInfo(const std::string& name, const std::string& path, const bool& is_directory)
```
mzhurovich: Passing bool by const ref makes no sense, just `bool` is ok here.

Bricks/file/file.h
```diff
@@ -214,79 +214,125 @@ struct FileSystem {
...
-          if (static_cast<int>(parameters) & static_cast<int>(mask)) {
-            if (!f(filename)) {
-              return;
+        while (struct dirent* entry = ::readdir(dir)) {
+          const char* const name = entry->d_name;
+          if (ScanDirCanHandleName(name)) {
+            // `IsDir` is proved to be required on Ubuntu running in Parallels on a Mac,
+            // with Bricks' directory mounted from Mac's filesystem.
+            // `entry->d_type` is always zero there, see http://comments.gmane.org/gmane.comp.lib.libcg.devel/4236
+            const std::string& path = JoinPath(directory, name);
```
mzhurovich: Dangling reference. It should be `const std::string`.

Bricks/file/file.h
```diff
@@ -214,79 +214,125 @@ struct FileSystem {
...
+        if (item_info.is_directory) {
+          ScanDirUntil<ITEM_HANDLER>(item_info.path, std::forward<ITEM_HANDLER>(item_handler), parameters, ScanDirRecursive::Yes);
+        }
+        return true;
+      };
+      ScanDirUntil(directory,
+                   item_handler_recursive,
+                   ScanDirParameters::ListFilesAndDirs,
+                   ScanDirRecursive::No);
+    }
```
mzhurovich: This part looks super complex. Why not just remove top-level `if (recursive == ScanDirRecursive::No)` and augment processing loop contents with:
```
if (is_directory && recursive == ScanDirRecursive::Yes) {
  ScanDirUntil(path, item_handler, parameters, recursive);
}
```

Bricks/file/test.cc
```diff
@@ -293,30 +292,105 @@ TEST(File, ScanDirParameters) {
   FileSystem::MkDir(d, FileSystem::MkDirParameters::Silent);
 
   {
-    std::set<std::string> xs;
-    FileSystem::ScanDir(base_dir, [&xs](const std::string& x) { xs.insert(x); });
-    EXPECT_EQ("f", current::strings::Join(xs, ','));
+    std::set<std::string> names;
+    std::map<std::string, const FileSystem::ScanDirItemInfo> item_infos;
```
mzhurovich: There's no profit in using `const` value type in containers here.
Removing it will make the code below more concise and readable.

Bricks/file/test.cc
```diff
@@ -293,30 +292,105 @@ TEST(File, ScanDirParameters) {
...
+    FileSystem::ScanDir(
+        base_dir,
+        [&names, &item_infos](const FileSystem::ScanDirItemInfo& x) { names.insert(x.name); item_infos.insert(std::make_pair(x.name, x)); });
```
mzhurovich: Lines are too long. Run `make indent` pls.

Bricks/file/file.h
dkorolev: +1

Bricks/file/file.h
dkorolev: +1

Bricks/file/file.h
dkorolev: +1

Bricks/file/file.h
dkorolev: +1

Bricks/file/test.cc
dkorolev: +1

dkorolev:
Let me / my insomnia take care of these.

dkorolev:
I assume Ivan wanted to write his Mac-developed code to be Windows-safe, hence the outside-`#ifdef` logic.

Also, what a nasty template recursive type issue.

@sompylasar @mzhurovich PTAL && thanks!

Bricks/file/file.h
dkorolev: On a second thought, `name` and `path` are shorter. I'm fine with either (as well as with `file` + `dir`, `filename` + `dirname`, etc.), and suggest @sompylasar jumps in :)

Bricks/file/file.h
dkorolev: See my comment below re. @sompylasar wanting to keep Windows safe. Btw, @mzhurovich, could you run the tests on Windows pls?

dkorolev:
(@sompylasar, please `git pull` as well, I've been pushing straight into your fork&branch.)

mzhurovich:
@dkorolev @sompylasar I meant smth like [this](https://github.com/mzhurovich/Current/commit/6f45623be00736eab64ba9e170298064b902de9d)

dkorolev:
Right, this changes Windows code too, which Ivan doesn't have the means to test.

I'd be fine either way.

Bricks/file/file.h
mzhurovich: Is `path` supposed to include basename or not? :)

Bricks/file/file.h
mzhurovich: What's wrong with Windows-safety in the change proposed?

sompylasar:
Re: names, `basename` and `pathname` (not `dirname`, or add one more field for it which in recursive case will be useful, so maybe add it).

sompylasar:
I copied the recursive part from RmDirRecursive which is implemented this was: non-recursive part is OS-dependent, recursive is not.

sompylasar:
Now I'm fine with either variant that works and you see fit for the core codebase.

sompylasar:
Added `dirname`: https://github.com/C5T/Current/pull/737/commits/42b095f8e15c95426fb16616981377fc983de749

dkorolev:
:+1:

dkorolev:
@sompylasar merging to unblock you. Don't forget to `git submodule update ; cd Current ; git fetch ; git checkout $commit_from_c5t_current ; cd .. ; git add Current`.

sompylasar:
@dkorolev Thanks! I'll post the next PR with recursive `ServeStaticFilesFrom` as soon as I implement it, and this will truly unblock me. Until that, I won't update submodule.

dkorolev:
окау

PR #739, opened 2017-06-01
Fix compilation and tests on Windows

grixa:
- Fix the `Bricks/net/http/impl/server.h` compilation on msvc2015
- Fix the HTTPAPI.PayloadTooLarge test.
- Fix a couple of warnings on msvc2015

### DIFF
--- a/Blocks/HTTP/impl/posix_client.h
+++ b/Blocks/HTTP/impl/posix_client.h
@@ -105,9 +105,16 @@ class GenericHTTPClientPOSIX final {
         // It swallows the `SocketWriteException` exception for huge payloads, as Current's HTTP server logic
         // does intentionally close the HTTP connection prematurely if `Content-Length` exceeds a reasonable limit.
         try {
+#ifndef CURRENT_WINDOWS
           connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n", true);
           connection.BlockingWrite("\r\n", true);
           connection.BlockingWrite(request_body_contents_, false);
+#else
+          // TODO(grigoriyn): this fix for the PayloadTooLarge test on Windows is temporary, need to revisit it.
+          connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n\r\n" +
+                                       request_body_contents_,
+                                   false);
+#endif
         } catch (const net::SocketWriteException&) {
           if (request_body_contents_.length() <= net::constants::kMaxHTTPPayloadSizeInBytes) {
             throw;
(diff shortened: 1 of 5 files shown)


Blocks/HTTP/impl/posix_client.h
```diff
@@ -105,9 +105,16 @@ class GenericHTTPClientPOSIX final {
...
         try {
+#ifndef CURRENT_WINDOWS
           connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n", true);
           connection.BlockingWrite("\r\n", true);
           connection.BlockingWrite(request_body_contents_, false);
+#else
+          // TODO(grigoriyn): this fix for the PayloadTooLarge test on Windows is temporary, need to revisit it.
+          connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n\r\n" +
+                                       request_body_contents_,
+                                   false);
```
mzhurovich: Hmm. What's the difference?

grixa: In 3xBlockingWrite version of the code on Windows the "Content-Length" is received, analyzed on server, predefined response is sent back (in case of too large length value) and the connection is closed **before** the client finishes the last `BlockingWrite` (with the body itself). And when it tries to get the respone, the `::recv` fails with `WSAECONNABORTED`. There should be more universal solution of that, of course.

Blocks/HTTP/impl/posix_client.h
```diff
@@ -105,9 +105,16 @@ class GenericHTTPClientPOSIX final {
         // It swallows the `SocketWriteException` exception for huge payloads, as Current's HTTP server logic
         // does intentionally close the HTTP connection prematurely if `Content-Length` exceeds a reasonable limit.
         try {
+#ifndef CURRENT_WINDOWS
           connection.BlockingWrite("Content-Length: " + std::to_string(request_body_contents_.length()) + "\r\n", true);
           connection.BlockingWrite("\r\n", true);
           connection.BlockingWrite(request_body_contents_, false);
+#else
+          // TODO(grigoriyn): this fix for the PayloadTooLarge test on Windows is temporary, need to revisit it.
```
dkorolev: `s/grigoriyn/grixa/`? :)

TypeSystem/Serialization/json/json.h
```diff
@@ -248,7 +248,7 @@ class JSONParser final {
     current_ = &document_;
   }
 
-  operator bool() const { return current_; }
+  operator bool() const { return current_ != nullptr; }
```
dkorolev: ;-)

Blocks/HTTP/impl/posix_client.h
grixa: :)

dkorolev:
@grixa, this one is LGTM, feel free to merge it in!

PR #741, opened 2017-06-03
Support compile time warnings for the `current::strings::Printf`

grixa:
Use `__attribute__(__format__,...)` to enable printf-like warnings.
Unfortunately, there is no an alternative for Windows: msvc2015 has `SAL Annotations` and `_Printf_format_string_` option, but it works only with `/analyze` option (or through `Run Code Analysis on Solution`).

### DIFF
--- a/Bricks/net/http/test.cc
+++ b/Bricks/net/http/test.cc
@@ -431,20 +431,20 @@ TEST(PosixHTTPServerTest, ChunkedSmoke) {
   };
   std::string body;
   std::string chunk(10, '.');
-  for (size_t i = 0; i < 10000; ++i) {
-    for (size_t j = 0; j < 10; ++j) {
+  for (uint64_t i = 0; i < 10000; ++i) {
+    for (uint64_t j = 0; j < 10; ++j) {
       chunk[j] = 'A' + ((i + j) % 26);
     }
     body += chunk;
   }
-  for (size_t length = body.length(); length > 1500; length -= (length / 7)) {
+  for (uint64_t length = body.length(); length > 1500; length -= (length / 7)) {
     const auto offset = body.length() - length;
-    for (size_t chunks = length; chunks; chunks = chunks * 2 / 3) {
+    for (uint64_t chunks = length; chunks; chunks = chunks * 2 / 3) {
       std::string chunked_body;
       for (uint64_t i = 0; i < chunks; ++i) {
-        const size_t start = offset + static_cast<size_t>(length * i / chunks);
-        const size_t end = offset + static_cast<size_t>(length * (i + 1) / chunks);
-        chunked_body += current::strings::Printf("%lX\r\n", end - start) + body.substr(start, end - start);
+        const uint64_t start = offset + length * i / chunks;
+        const uint64_t end = offset + length * (i + 1) / chunks;
+        chunked_body += current::strings::Printf("%llX\r\n", end - start) + body.substr(start, end - start);
       }
 
       std::thread t(EchoServerThreadEntry, Socket(FLAGS_net_http_test_port));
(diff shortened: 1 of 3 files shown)


Bricks/net/http/test.cc
```diff
@@ -441,10 +441,10 @@ TEST(PosixHTTPServerTest, ChunkedSmoke) {
     const auto offset = body.length() - length;
     for (size_t chunks = length; chunks; chunks = chunks * 2 / 3) {
       std::string chunked_body;
-      for (size_t i = 0; i < chunks; ++i) {
-        const size_t start = offset + length * i / chunks;
-        const size_t end = offset + length * (i + 1) / chunks;
-        chunked_body += current::strings::Printf("%X\r\n", end - start) + body.substr(start, end - start);
+      for (uint64_t i = 0; i < chunks; ++i) {
```
dkorolev: But `chunks` is `size_t`, why this change?

Bricks/net/http/test.cc
```diff
@@ -441,10 +441,10 @@ TEST(PosixHTTPServerTest, ChunkedSmoke) {
...
+        const size_t start = offset + static_cast<size_t>(length * i / chunks);
+        const size_t end = offset + static_cast<size_t>(length * (i + 1) / chunks);
+        chunked_body += current::strings::Printf("%lX\r\n", end - start) + body.substr(start, end - start);
```
dkorolev: Perhaps just cast to a `uint64_t` here, as the 2nd parameter to `Printf`?

Bricks/strings/printf.h
```diff
@@ -33,6 +33,9 @@ SOFTWARE.
 namespace current {
 namespace strings {
 
+#ifndef CURRENT_WINDOWS
```
dkorolev: I'm cool with this staying for `CURRENT_WINDOWS`. @mzhurovich, you too? :-)

dkorolev: Oh, I'm an idiot. This is *not* for Windows.

Perhaps guard it by compiler type/version then? An `#ifdef` for whether we're under `g++` or `clang++`?

CC @mzhurovich

mzhurovich: I suppose it should be supported by any version of GCC/Clang that knows about c++11, so adding some specific version guards may be an overkill.

Bricks/net/http/test.cc
mzhurovich: As the commit message says - "integer overflow" :)

Bricks/strings/printf.h
dkorolev: Hate to sound like a prick, but there are other compilers. Perhaps introduce a top-level `CURRENT_GCC_OR_CLANG` into `port.h` and guard by it here?

Bricks/net/http/test.cc
dkorolev: OH SHI~!

Then perhaps make the outer look `uint64_t` too?

Sorry, still hungry and still waking up after a lava hike yesterday.

dkorolev: *the outer loop

Bricks/strings/printf.h
mzhurovich: Ok, `__attributes__` is a GNU extension and should be supported by a GNU compiler, so the right ifdef here is
```
#ifdef __GNUC__
```

grixa: Thanks! Fixed in the a33e245

Bricks/net/http/test.cc
grixa: You are right, it's better to use `uint64_t` everywhere around to make the code look less confusing. 59ea5d0

dkorolev:
Thank you Grisha and Max! Merging.

PR #742, opened 2017-06-04
README: Contribution, Quick Start, Development Dependencies

sompylasar:
Please feel free to add OS-specific hints and dependencies you know about.


README.md
```diff
@@ -29,3 +29,34 @@ Storage layer with super easy to use in-memory data views and Sherlock-based per
...
+
+## Contribution
+
+Welcome, contributors! Please [start here by signing the CLA](https://github.com/C5T/Current/blob/master/contributors/README.md).
+
+## Quick Start
+
+Install the development dependencies (see below).
+
+Clean, build and run all tests, verify code:
```
dkorolev: "and run tests" or "and run all the tests". Not sure what does "verify code" mean here.

README.md
```diff
@@ -29,3 +29,34 @@ Storage layer with super easy to use in-memory data views and Sherlock-based per
...
+```
+make clean
+make test
```
dkorolev: A comment would help, ex.
```
# Builds and runs all tests as a single binary. Slow, eats up tons of CPU, but measures coverage.
# Consider `make individual_tests` or `make test` within individual directories to run the subset of tests.
```

README.md
```diff
@@ -29,3 +29,34 @@ Storage layer with super easy to use in-memory data views and Sherlock-based per
...
+make check
```
dkorolev: Another comment maybe:
```
# "Builds" all header files individually, twice each header file, and "links" these pairs together.
# Done to ensure no symbols are leaked to the global namespace, as Current is a header-only library.
```

README.md
```diff
@@ -29,3 +29,34 @@ Storage layer with super easy to use in-memory data views and Sherlock-based per
...
+```
+
+Build and run tests for each module separately:
+```
+make individual_tests
+```
+
+Build and run tests for one of the modules (e.g. `Blocks/HTTP`):
+```
+( cd Blocks/HTTP && make -j test )
```
dkorolev: The `-j` is useless in most or all per-dir tests, as it's just one dependency-free target to build (unlike what we have in RS thx to Nathan).

README.md
```diff
@@ -29,3 +29,34 @@ Storage layer with super easy to use in-memory data views and Sherlock-based per
...
+```
+
+## Development Dependencies
+
+- `nasm` for `FnCAS`. **macOS**: `brew install nasm`
```
dkorolev: Perhaps add some "or disable FnCAS tests" comment? Perhaps a macro and an environmental var for it, as we're on this page?

README.md
```diff
@@ -29,3 +29,34 @@ Storage layer with super easy to use in-memory data views and Sherlock-based per
...
+- `geninfo` from `lcov` for coverage report. **macOS**: `brew install lcov`
```
dkorolev: I'd note that this part is unnecessary. The green `"ALL TESTS PASS"` does mean that all tests pass, and often times that's all we need.

README.md
mzhurovich: `-j` makes no sense since there's always a single `test.cc`

README.md
```diff
@@ -29,3 +29,34 @@ Storage layer with super easy to use in-memory data views and Sherlock-based per
...
+- `clang-format` for code formatting. **macOS**: `brew install clang-format`
```
dkorolev: This part is a bit tricky as `clang-format` versions change and they affect formatting. I believe we're on 3.6 now, and it's an open question where and how do we move forward (the newer one messes up the lambdas badly). I'd love to have more clarity here too, but so far we don't quite.

README.md
dkorolev: Although "global namespace" is misleading here. "No symbols are exported, and the ODR will not be violated when linking together two objects, each of which is independently using Current."

CC @sompylasar @grixa.

README.md
mzhurovich: Current uses `clang-format-3.6` and on OS X it was available under `brew install llvm36`, but looks like it's not available anymore ((

README.md
sompylasar: "verify code" stands for `make check`

README.md
sompylasar: It's easier to install an extra dependency than to disable the tests. This is a quick start guide.

README.md
sompylasar: The top-level `make test` fails after `ALL TESTS PASS` if the test coverage tool is missing, so it's easier to install an extra dependency.

README.md
sompylasar: Slightly extended.

README.md
sompylasar: Removed -j

sompylasar: Removed -j

README.md
sompylasar: clang-format-3.8 is available, though it makes more changes than yours 3.6: it sorts includes, and one more thing I don't remember.

README.md
dkorolev: "Clean, build and run all tests, verify code" just does not sound English.

dkorolev:
:+1:

PR #743, opened 2017-06-06
Unblockage

dkorolev:


Blocks/Persistence/file.h
```diff
@@ -409,7 +409,7 @@ class FilePersister {
       CURRENT_THROW(ss::InconsistentTimestampException(iterator.head + std::chrono::microseconds(1), timestamp));
     }
     iterator.head = timestamp;
-    const auto head_str = Printf(constants::kHeadFromatString, timestamp.count());
+    const auto head_str = Printf(constants::kHeadFromatString, static_cast<long long>(timestamp.count()));
```
mzhurovich: Ouch, there's a typo: s/kHeadFromatString/kHeadFormatString/

mzhurovich:
Thanks @dkorolev!
Are the warnings result of the latest @grixa's change that enabled format specifier checks?
Also TIL that there are special `PRIxx/SCNxx` macros defined in standard for int**x**_t printf/scanf platform-independent specifiers: http://en.cppreference.com/w/cpp/types/integer
Do we want try use them or stick with casting to `long long`?

Blocks/Persistence/file.h
dkorolev: 6064730

dkorolev:
> Are the warnings result of the latest @grixa's change that enabled format specifier checks?
I believe so.

> Also TIL that there are special PRIxx/SCNxx macros defined in standard for intx_t printf/scanf platform-> independent specifiers: http://en.cppreference.com/w/cpp/types/integer
> Do we want try use them or stick with casting to long long?
Nah :-)

dkorolev:
I'll take the liberty to merge it :)

PR #744, opened 2017-06-13
ServeStaticFilesFrom with separate route and URL prefixes. URL redirects refactor.

sompylasar:

Unfortunately, this required the changes to the `URL` class behavior: its "previous URL" handling and defaults handling. There were too many variants of use of this class's options, some of which were conflicting with each other (e.g. default scheme with an unknown port or no scheme with an unknown port).

The `SocketResolveAddressException` was filled with the resolve arguments to have more details about the error in runtime.

CC @dkorolev @grixa @mzhurovich

### DIFF
--- a/Blocks/HTTP/impl/posix_server.h
+++ b/Blocks/HTTP/impl/posix_server.h
@@ -120,7 +120,12 @@ struct StaticFileServer {
       : content(std::move(content)),
         content_type(std::move(content_type)),
         serves_directory(serves_directory),
-        trailing_slash_redirect_url(trailing_slash_redirect_url) {}
+        trailing_slash_redirect_url(trailing_slash_redirect_url) {
+    // TODO(sompylasar): Test the empty `trailing_slash_redirect_url` case when this class goes public.
+    // For now, make `trailing_slash_redirect_url` mandatory for serving directories.
+    CURRENT_ASSERT((serves_directory ? !trailing_slash_redirect_url.empty() : true));
+  }
+
   void operator()(Request r) {
     if (r.method == "GET") {
       if (serves_directory == r.url_path_had_trailing_slash) {
@@ -135,13 +140,21 @@ struct StaticFileServer {
         r.connection.SendHTTPResponse(current::net::DefaultNotFoundMessage(),
                                       HTTPResponseCode.NotFound,
                                       current::net::constants::kDefaultHTMLContentType);
-      } else {
+      } else if (!trailing_slash_redirect_url.empty()) {
         // Redirect to add trailing slash to the directory URL. Example: `/static` -> `/static/`.
         // The trailing slash is required to make the browser relative URL resolution algorithm use directory as base
         // URL for the index file served at that directory URL (without filename).
         // See RFC1808, `Resolving Relative URLs`, `Step 6`: https://www.ietf.org/rfc/rfc1808.txt
         r.connection.SendHTTPResponse(
             "", HTTPResponseCode.Found, content_type, current::net::http::Headers({{"Location", trailing_slash_redirect_url}}));
+      } else {
+        // TODO(sompylasar): Test the empty `trailing_slash_redirect_url` case this class goes public: this piece.
+        // Respond with HTTP 404 Not Found if we're serving a directory, don't have a trailing slash,
+        // and don't know where to redirect that. Example: `/static`.
+        // Note: This should never happen with `ServeStaticFilesFrom` because for directories it always provides the URL.
+        r.connection.SendHTTPResponse(current::net::DefaultNotFoundMessage(),
+                                      HTTPResponseCode.NotFound,
+                                      current::net::constants::kDefaultHTMLContentType);
       }
     } else {
       r.connection.SendHTTPResponse(current::net::DefaultMethodNotAllowedMessage(),
--- a/Blocks/HTTP/impl/posix_server.h
+++ b/Blocks/HTTP/impl/posix_server.h
@@ -291,18 +291,24 @@ class HTTPServerPOSIX final {
             const bool path_components_empty = item_info.path_components_cref.empty();
             const std::string path_components_joined = current::strings::Join(item_info.path_components_cref, '/');
 
-            // `route_for_directory` has no trailing slash.
+            // `route_for_directory` must have leading slash and must not have trailing slash except if it's a root.
             const std::string route_for_directory =
                 options.public_route_prefix +
                 ((options.public_route_prefix == "/" || path_components_empty) ? "" : "/") +
                 path_components_joined;
+            CURRENT_ASSERT(route_for_directory == "/" || (route_for_directory.front() == '/' && route_for_directory.back() != '/'));
 
             // Ignore files nested in directories named with a leading dot (means hidden in POSIX).
             if (route_for_directory.find("/.") != std::string::npos) {
               return;
             }
 
-            const std::string route_for_file = route_for_directory + (route_for_directory == "/" ? "" : "/") + item_info.basename;
+            // `route_for_file` must have leading slash and must not have trailing slash.
+            const std::string route_for_file =
+                route_for_directory +
+                (route_for_directory == "/" ? "" : "/") +
+                item_info.basename;
+            CURRENT_ASSERT(route_for_file.front() == '/' && route_for_file.back() != '/');
 
             const bool is_index_file =
                 (std::find(options.index_filenames.begin(), options.index_filenames.end(), item_info.basename) !=
@@ -319,10 +325,12 @@ class HTTPServerPOSIX final {
                     ServeStaticFilesFromCannotServeMoreThanOneIndexFile(route_for_directory + ' ' + item_info.basename));
               }
 
+              // `trailing_slash_redirect_url` must have trailing slash.
               std::string trailing_slash_redirect_url =
                   options.public_url_prefix +
-                  ((options.public_url_prefix.back() == '/' || path_components_empty) ? "" : "/") +
+                  (options.public_url_prefix.back() == '/' ? "" : "/") +
                   (path_components_empty ? "" : path_components_joined + "/");
+              CURRENT_ASSERT(trailing_slash_redirect_url.length() > 0 && trailing_slash_redirect_url.back() == '/');
 
               auto static_file_server = std::make_unique<StaticFileServer>(content, content_type, true, trailing_slash_redirect_url);
               scope += Register(route_for_directory, *static_file_server);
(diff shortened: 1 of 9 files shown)


Blocks/HTTP/test.cc
```diff
@@ -357,6 +357,28 @@ TEST(HTTPAPI, Redirect) {
   EXPECT_EQ(Printf("http://localhost:%d/to", FLAGS_net_api_test_port), response.url);
 }
 
+TEST(HTTPAPI, RedirectToFullURL) {
+  // TODO(sompylasar): Cannot test redirect to a different port because the HTTP client
+  // is following the redirect and tries to connect to the redirect target,
+  // and we do not have a second port to occupy for the tests.
```
dkorolev: You can totally use a second port if you want to.

dkorolev:
Looks like the

```
[ RUN      ] HTTPAPI.ServeStaticFilesFromOptionsCustomPublicRoutePrefixAndPublicUrlPrefix
unknown file: Failure
C++ exception with description "docu/client/../../../../Bricks/net/http/impl/../../tcp/impl/posix.h:557	SocketConnectException()	" thrown in the test body.
[  FAILED  ] HTTPAPI.ServeStaticFilesFromOptionsCustomPublicRoutePrefixAndPublicUrlPrefix (3 ms)
```

test is consistently failing on Travis.

Blocks/HTTP/test.cc
sompylasar: Should I add a "flag" for it then, or just this port plus 1 will suffice?

sompylasar:
RE: SocketConnectException, weird, will look into it.

Blocks/HTTP/test.cc
dkorolev: A flag would be better. It'll allocate a new port, yes.

Blocks/URL/url.h
```diff
@@ -356,7 +359,7 @@ struct URL : URLParametersExtractor, URLWithoutParametersParser {
 
   // Extra parameters for previous host and port are provided in the constructor to handle redirects.
   URL(const std::string& url,
-      const std::string& previous_scheme = kDefaultScheme,
+      const std::string& previous_scheme = "",
       const std::string& previous_host = "",
       const int previous_port = 0)
```
mzhurovich: We've already changed `int port` to `uint16_t` port in some places - let's fix it here as well?

sompylasar: Sure.

Blocks/HTTP/impl/posix_server.h
```diff
@@ -126,7 +141,7 @@ struct StaticFileServer {
         // URL for the index file served at that directory URL (without filename).
         // See RFC1808, `Resolving Relative URLs`, `Step 6`: https://www.ietf.org/rfc/rfc1808.txt
         r.connection.SendHTTPResponse(
-            "", HTTPResponseCode.Found, content_type, current::net::http::Headers({{"Location", r.url.path + '/'}}));
+            "", HTTPResponseCode.Found, content_type, current::net::http::Headers({{"Location", trailing_slash_redirect_url}}));
```
mzhurovich: Should the Location header be set if `trailing_slash_redirect_url` is empty?

sompylasar: Good question, I assumed this class to be of internal use, and in this case it won't be empty, but yes, I can add a check to be safe.

Blocks/HTTP/test.cc
sompylasar: Done. https://github.com/C5T/Current/pull/744/commits/c795aa3915da1446ce3d574a13bc35efa55420fb

Blocks/HTTP/test.cc
```diff
@@ -1068,13 +1095,60 @@ TEST(HTTPAPI, ServeStaticFilesFromOptionsCustomURLBase) {
...
+
+  // Subsubdirectory index file.
+  EXPECT_EQ("<h1>HTML index</h1>", HTTP(GET(Printf("http://localhost:%d/static/something/sub_dir/sub_sub_dir/", FLAGS_net_api_test_port))).body);
+
+  // Redirect from subsubdirectory without trailing slash to subsubdirectory with trailing slash.
+  ASSERT_THROW(HTTP(GET(Printf("http://localhost:%d/static/something/sub_dir/sub_sub_dir", FLAGS_net_api_test_port))),
+               HTTPRedirectNotAllowedException);
+  const auto sub_sub_dir_response = HTTP(GET(Printf("http://localhost:%d/static/something/sub_dir/sub_sub_dir", FLAGS_net_api_test_port)).AllowRedirects());
+  EXPECT_EQ("http://localhost/anything/sub_dir/sub_sub_dir/", sub_sub_dir_response.url);
+}
```
mzhurovich: Running `make indent` won't hurt

Bricks/net/tcp/impl/posix.h
```diff
@@ -517,10 +517,10 @@ inline addrinfo_t GetAddrInfo(const std::string& host, const std::string& serv =
   hints.ai_protocol = IPPROTO_TCP;
   const int retval = ::getaddrinfo(host.c_str(), serv.c_str(), &hints, &result);
   if (!result) {
-    CURRENT_THROW(SocketResolveAddressException());
+    CURRENT_THROW(SocketResolveAddressException(host + ' ' + serv));
```
mzhurovich: 👍

Blocks/HTTP/impl/posix_server.h
```diff
@@ -88,33 +88,48 @@ struct ServeStaticFilesFromCannotServeMoreThanOneIndexFile : ServeStaticFilesExc
 };
 
 struct ServeStaticFilesFromOptions {
-  std::string url_base;
+  // HTTP server route prefix.
+  std::string public_route_prefix;
+
+  // User-facing URL prefix. Used to prefix the route in the trailing slash redirects.
+  std::string public_url_prefix;
```
mzhurovich: Isn't the route above a "user-facing" as well?

Blocks/URL/test.cc
```diff
@@ -49,7 +49,7 @@ TEST(URLTest, SmokeTest) {
   u = URL("www.google.com:8080");
   EXPECT_EQ("www.google.com", u.host);
   EXPECT_EQ("/", u.path);
-  EXPECT_EQ("http", u.scheme);
+  EXPECT_EQ("", u.scheme);  // Since there is no rule for port "8080", no scheme is returned.
```
mzhurovich: 👍

Blocks/URL/test.cc
```diff
@@ -95,42 +101,58 @@ TEST(URLTest, CompositionTest) {
...
-  EXPECT_EQ("https://www.google.com:442/", URL("www.google.com", "https", "", 442).ComposeURL());
-  EXPECT_EQ("https://www.google.com/", URL("www.google.com", "https", "", 443).ComposeURL());
-  EXPECT_EQ("https://www.google.com:444/", URL("www.google.com", "https", "", 444).ComposeURL());
-  // Since there is no rule from "23" to "telnet", no scheme is specified.
-  EXPECT_EQ("www.google.com:23/", URL("www.google.com", "", "", 23).ComposeURL());
+  EXPECT_EQ("https://www.google.com:442/", URL("https://www.google.com:442").ComposeURL());
+  EXPECT_EQ("https://www.google.com/", URL("https://www.google.com:443").ComposeURL());
+  EXPECT_EQ("https://www.google.com:444/", URL("https://www.google.com:444").ComposeURL());
+  // Since there is no rule for port "23", no scheme is returned.
+  EXPECT_EQ("www.google.com:23/", URL("www.google.com:23").ComposeURL());
```
mzhurovich: In its previous version, this test was explicitly testing a behavior with URL ctor parameters dealing with redirects. I'd suggest to keep the separate tests for port-scheme mapping and tests for derived scheme.

Blocks/HTTP/impl/posix_server.h
sompylasar: Not in every case, that's why this PR. Imagine: user-facing URL is `http://api.example.com/files/*` but the server serves at `/static/*`. If the server has to issue a trailing slash redirect for `http://api.example.com/files/somedir`, it should return either `/files/somedir/` or `http://api.example.com/files/somedir/`, neither of these contain `/static`

Blocks/URL/test.cc
sompylasar: Agree. This was about morning, I could barely wrap my head around the obscure logic of the "previous scheme", "previous port", "previous host", their default values and implicit behavior, etc. I'd rather pull the redirect logic out of `URL` constructor if that sounds better.

sompylasar:
RE: `SocketConnectException` in `HTTPAPI.ServeStaticFilesFromOptionsCustomPublicRoutePrefixAndPublicUrlPrefix`, this is because it likely goes to port `80` somewhere; it started failing when I turned off the local `nginx` which was serving port `80` on my machine; so please be aware of that implicit dependency.

Sometimes, especially in low-level APIs, default values cause troubles.

sompylasar:
And obviously, the public URL where the redirects go is configured without a port, so `80` is assumed. `"http://localhost/anything"});`
This test needs the secondary port server.

Blocks/HTTP/impl/posix_server.h
sompylasar: Done. https://github.com/sompylasar/Current/commit/1c658d0f6b265ef1f79728c59ffa4a223e98b1bd

Blocks/URL/url.h
sompylasar: Done. https://github.com/C5T/Current/pull/744/commits/c795aa3915da1446ce3d574a13bc35efa55420fb

sompylasar:
This fix https://github.com/C5T/Current/pull/744/commits/4f397cb9f34c0589d3f0b2fc7304b0232f19922d caused `HTTPRedirectLoopException` 😞  investigating, hold on...

dkorolev:
@sompylasar, in lieu of best TDD practices, could you show the shortest possible test that fails w/ the current code to illustrate the issue you're working on fixing?

I still struggle to get to the root of it. Thanks!

sompylasar:
@dkorolev I'm sorry, I got used to push immediately to save the work in the cloud/central server, and run the tests in parallel to that. But pushing to a PR incurs extra work: both human (review) and CI (build), and I keep forgetting that. I've fixed that, added assertions and extended exception info, see latest commits.

sompylasar:
I am now reworking the `URL` module to have an explicit `RedirectToURL` method which deals with all the redirect-related defaults and substitutions. The logic is too complex and spaghetti to fit into the single constructor.

dkorolev:
Oh, so it's the same as Hypermedia base for our REST/CRUD storage. Got it, thanks!

Sure thing, OK to make small commits. My TDD commit was not re. those, but re. it being unclear to me before your last commit what the root issue is.

sompylasar:
`URL` class reworked, see: https://github.com/C5T/Current/pull/745 (merged here as well).

sompylasar:
Closing in favor of https://github.com/C5T/Current/pull/746

PR #745, opened 2017-06-14
Blocks/URL: Simplified the `URL` class public API and impl; added `username` and `password`.

sompylasar:
#### ~~ALSO MERGED INTO https://github.com/C5T/Current/pull/744~~ (https://github.com/C5T/Current/pull/744 closed in favor of https://github.com/C5T/Current/pull/746 which has not yet got the `URL` changes that fix the with-port-to-without-port redirect)

* Removed extra parameters to reduce the API surface that requires support and testing.
* Simplified internal logic (it's still complex and hard to grasp, unfortunately, because we chose to support many partial URL formats and make it smart).
* Made explicit methods for filling with default values and for redirect handling.
* Added support for username and password (there was a TODO for that).
* Simplified `URL` class inheritance, using methods, not constructors/initializers, to easier propagate data from one to another without a need for a class field.
* Replaced public `url_without_parameters` field with a `ComposeURLWithoutParameters` method as the field value is hard to keep in sync with the other properties' values.
* Fixed `EncodeURIComponent` encoding extra characters which should not be encoded according to RFC.

PR #746, opened 2017-06-14
ServeStaticFilesFrom with separate route and URL prefixes.

sompylasar:

1. If a directory is requested via a URL without a trailing slash (ex. `http://localhost:1234/directory`), `ServeStaticFilesFrom` knows this is a directory, and instead of serving the index file from that directory right away (keeping the URL at `http://localhost:1234/directory`), it issues a redirect to the URL with a trailing slash (ex. `http://localhost:1234/directory/`) because the URLs to directories with a trailing slash and without a trailing slash are treated differently by the browsers with respect to the relative URL resolution from the index file (ex. `./some.js` for this URL `http://localhost:1234/directory` is this URL `http://localhost:1234/some.js`; but `./some.js` for this URL `http://localhost:1234/directory/` is `http://localhost:1234/directory/some.js`).

2. Between the server with `ServeStaticFilesFrom` and the user of the browser, there may be a proxy server which transforms the URLs (ex. `http://public.example.com/files/file.html` is transformed to `http://localhost:1234/public/api/static/file.html`). This proxy server only does a forward transformation, but not the reverse one, because the response body can also contain user-facing URLs, and the proxy server does not know how to transform the body. This means the server has to know how to build the user-facing URLs i.e. to perform the reverse URL transformation.

3. Before this change, `ServeStaticFilesFrom` was not aware of the shape of the user-facing URLs. In the simplest case, when there is no proxy, this worked because the route at which the server responds matches the user-facing URL (ex. this URL `http://localhost:1234/public/api/directory` when directly accessed from a browser located at `localhost` matches the route `/public/api/directory` at which the file is served, and can be redirected to `/public/api/directory/` with the trailing slash to ensure the correct relative URL resolution on the browser side). On the other hand, when the proxy is involved serving at `http://public.example.com/directory` and is prefixing `/public/api` under the hood, the browser expects `http://public.example.com/directory/` when redirected from `http://public.example.com/directory` but instead gets `http://public.example.com/public/api/directory/` resolved by the browser from the root-relative URL `/public/api/directory/` by prepending the current scheme and host. This results in `HTTP 404 Not Found` because nothing is served at route `/public/api/public/api/directory`.

4. After this change, `ServeStaticFilesFrom` takes an optional public URL prefix which substitutes the route prefix when building the user-facing URLs for the redirects.

Additionally:
* `SocketResolveAddressException` message now contains the resolve arguments to have more details about the error in runtime.
* `HTTPRedirectLoopException` message now contains the URLs of the loop.

Changes are squashed from branch `serve_static_files_public_url_prefix_different_from_internal_route_prefix` without `url_refactor`, see https://github.com/C5T/Current/pull/744 and https://github.com/C5T/Current/pull/745

Tests not passing in this commit are the reason for the `url_refactor` changes: when a redirect occurs from a URL with a port ("previous URL") to a URL without a port, the port gets propagated to the resulting URL and to the "Location" header.

```
[ RUN      ] URLTest.RedirectPreservesSchemeHostAndPortTest
test.cc:144: Failure
Value of: URL("/empty_all_with_previous_empty_scheme", URL("localhost:8080")).ComposeURL()
  Actual: "http://localhost:8080/empty_all_with_previous_empty_scheme"
Expected: "localhost:8080/empty_all_with_previous_empty_scheme"
test.cc:156: Failure
Value of: URL("ftp://host_no_port_path", URL("meh://localhost:8080")).ComposeURL()
  Actual: "ftp://host_no_port_path:8080/"
Expected: "ftp://host_no_port_path/"
test.cc:157: Failure
Value of: URL("blah://new_host/foo", URL("meh://localhost:5000")).ComposeURL()
  Actual: "blah://new_host:5000/foo"
Expected: "blah://new_host/foo"
[  FAILED  ] URLTest.RedirectPreservesSchemeHostAndPortTest (0 ms)
```

The full stack of the redirect with HTTP server and client cannot be tested with the Current tests setup because they always use ports and cannot redirect to a URL without a port because `HTTPClient` tries to connect to it.

CC @dkorolev @mzhurovich @grixa

### DIFF
--- a/Blocks/HTTP/impl/posix_client.h
+++ b/Blocks/HTTP/impl/posix_client.h
@@ -80,7 +80,7 @@ class GenericHTTPClientPOSIX final {
       redirected = false;
       const std::string composed_url = parsed_url.ComposeURL();
       if (all_urls.count(composed_url)) {
-        CURRENT_THROW(current::net::HTTPRedirectLoopException());
+        CURRENT_THROW(current::net::HTTPRedirectLoopException(current::strings::Join(all_urls, ' ') + ' ' + composed_url));
       }
       all_urls.insert(composed_url);
       current::net::Connection connection(
@@ -132,8 +132,12 @@ class GenericHTTPClientPOSIX final {
       // the numerical response code "200" can be accessed with the same method as the "/path".
       const int response_code_as_int = atoi(http_request_->RawPath().c_str());
       response_code_ = HTTPResponseCode(response_code_as_int);
+      // Follow the redirects automatically.
+      // Note: This is by no means a complete redirect implementation.
       if (response_code_as_int >= 300 && response_code_as_int <= 399 && !http_request_->location.empty()) {
-        // Note: This is by no means a complete redirect implementation.
+        if (!allow_redirects_) {
+          CURRENT_THROW(current::net::HTTPRedirectNotAllowedException());
+        }
         redirected = true;
         parsed_url = URL(http_request_->location, parsed_url);
         response_url_after_redirects_ = parsed_url.ComposeURL();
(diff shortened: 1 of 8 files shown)


sompylasar:
Closing, will be resubmitted without the failing `URL` tests.

sompylasar:
Resubmitted: https://github.com/C5T/Current/pull/752

PR #747, opened 2017-06-15
Blocks/URL: Fixed `EncodeURIComponent` encoding extra characters which should not be encoded according to RFC.

sompylasar:
Before:
```
[ RUN      ] URLTest.ExtractsURLParameters
test.cc:192: Failure
Value of: u.ComposeURL()
  Actual: "http://www.google.com/a?encoded%23query%2Dwith%5Funreserved%7Echars%21and%25reserved%3A%5Bchars%5Dplus%2Band%20space=#foo"
Expected: "http://www.google.com/a?encoded%23query-with_unreserved~chars!and%25reserved%3A%5Bchars%5Dplus%2Band%20space=#foo"
[  FAILED  ] URLTest.ExtractsURLParameters (2 ms)
```

### DIFF
--- a/Blocks/URL/url.h
+++ b/Blocks/URL/url.h
@@ -208,31 +208,44 @@ struct URLParametersExtractor {
 
   static bool IsHexDigit(char c) { return (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F'); }
 
+  static bool IsURIComponentSpecialCharacter(char c) {
+    // https://github.com/lyokato/cpp-urilite/blob/723083f98ddf42f2b610b62444c598236b5ce3e8/include/urilite.h#L52-L65
+    // RFC2396 unreserved?
+    return ((c >= 'a' && c <= 'z')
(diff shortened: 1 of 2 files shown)


Blocks/URL/url.h
```diff
@@ -208,31 +208,44 @@ struct URLParametersExtractor {
...
 
   static std::string EncodeURIComponent(const std::string& decoded) {
-    std::string encoded;
+    std::ostringstream encoded;
     for (const char c : decoded) {
-      if (::isalpha(c) || ::isdigit(c)) {
-        encoded += c;
+      if (IsURIComponentSpecialCharacter(c)) {
+        encoded << c;
+      } else if (c == ' ') {
```
dkorolev: Looks redundant.

sompylasar: Inspired by this: https://github.com/lyokato/cpp-urilite/blob/723083f98ddf42f2b610b62444c598236b5ce3e8/include/urilite.h#L90-L92 – where the encoded space varies based on plus encoding mode (which Current's `URL` does not have).

PR #748, opened 2017-06-15
Blocks/HTTP/impl/posix_client: Throw `HTTPRedirectNotAllowedException` as soon as a redirecting status code is received.

sompylasar:
Previously, the exception was thrown after all the requests caused by the redirects have been performed.
If the client does not allow redirects, it should not follow them.

### DIFF
--- a/Blocks/HTTP/impl/posix_client.h
+++ b/Blocks/HTTP/impl/posix_client.h
@@ -132,8 +132,12 @@ class GenericHTTPClientPOSIX final {
       // the numerical response code "200" can be accessed with the same method as the "/path".
       const int response_code_as_int = atoi(http_request_->RawPath().c_str());
       response_code_ = HTTPResponseCode(response_code_as_int);
+      // Follow the redirects automatically.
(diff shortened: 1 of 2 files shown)


Blocks/HTTP/impl/posix_client.h
```diff
@@ -231,13 +243,10 @@ struct ImplWrapper<HTTPClientPOSIX> {
   }
 
   template <typename REQUEST_PARAMS, typename RESPONSE_PARAMS>
-  inline static void ParseOutput(const REQUEST_PARAMS& request_params,
+  inline static void ParseOutput(const REQUEST_PARAMS& /*request_params*/,
```
dkorolev: Nit: Omit parameter name, or `static_cast<void>(it)` is better.

sompylasar: 👍 That's how it was here for the next parameter.

sompylasar: https://github.com/C5T/Current/pull/753

PR #752, opened 2017-06-15
Blocks/HTTP/impl/posix_server: ServeStaticFilesFrom public URL prefix.

sompylasar:

1. If a directory is requested via a URL without a trailing slash (ex. `http://localhost:1234/directory`), `ServeStaticFilesFrom` knows this is a directory, and instead of serving the index file from that directory right away (keeping the URL at `http://localhost:1234/directory`), it issues a redirect to the URL with a trailing slash (ex. `http://localhost:1234/directory/`) because the URLs to directories with a trailing slash and without a trailing slash are treated differently by the browsers with respect to the relative URL resolution from the index file (ex. `./some.js` for this URL `http://localhost:1234/directory` is this URL `http://localhost:1234/some.js`; but `./some.js` for this URL `http://localhost:1234/directory/` is `http://localhost:1234/directory/some.js`).

2. Between the server with `ServeStaticFilesFrom` and the user of the browser, there may be a proxy server which transforms the URLs (ex. `http://public.example.com/files/file.html` is transformed to `http://localhost:1234/public/api/static/file.html`). This proxy server only does a forward transformation, but not the reverse one, because the response body can also contain user-facing URLs, and the proxy server does not know how to transform the body. This means the server has to know how to build the user-facing URLs i.e. to perform the reverse URL transformation.

3. Before this change, `ServeStaticFilesFrom` was not aware of the shape of the user-facing URLs. In the simplest case, when there is no proxy, this worked because the route at which the server responds matches the user-facing URL (ex. this URL `http://localhost:1234/public/api/directory` when directly accessed from a browser located at `localhost` matches the route `/public/api/directory` at which the file is served, and can be redirected to `/public/api/directory/` with the trailing slash to ensure the correct relative URL resolution on the browser side). On the other hand, when the proxy is involved serving at `http://public.example.com/directory` and is prefixing `/public/api` under the hood, the browser expects `http://public.example.com/directory/` when redirected from `http://public.example.com/directory` but instead gets `http://public.example.com/public/api/directory/` resolved by the browser from the root-relative URL `/public/api/directory/` by prepending the current scheme and host. This results in `HTTP 404 Not Found` because nothing is served at route `/public/api/public/api/directory`.

4. After this change, `ServeStaticFilesFrom` takes an optional public URL prefix which substitutes the route prefix when building the user-facing URLs for the redirects.

### DIFF
--- a/Blocks/HTTP/impl/posix_server.h
+++ b/Blocks/HTTP/impl/posix_server.h
@@ -88,12 +88,21 @@ struct ServeStaticFilesFromCannotServeMoreThanOneIndexFile : ServeStaticFilesExc
 };
 
 struct ServeStaticFilesFromOptions {
-  std::string url_base;
+  // HTTP server route prefix.
+  std::string route_prefix;
+
+  // User-facing URL prefix. Used to prefix the route in the trailing slash redirects.
+  std::string public_url_prefix;
+
+  // Names of files to serve if a directory URL is requested, in the priority order (first found will be served).
   std::vector<std::string> index_filenames;
 
-  explicit ServeStaticFilesFromOptions(std::string url_base = "/",
-                                       std::vector<std::string> index_filenames = {"index.html", "index.htm"})
-      : url_base(std::move(url_base)), index_filenames(std::move(index_filenames)) {}
+  explicit ServeStaticFilesFromOptions(std::string route_prefix_in = "/",
+                                       std::string public_url_prefix_in = "",
+                                       std::vector<std::string> index_filenames_in = {"index.html", "index.htm"})
+    : route_prefix(std::move(route_prefix_in)),
+      public_url_prefix(public_url_prefix_in.empty() ? route_prefix : std::move(public_url_prefix_in)),
+      index_filenames(std::move(index_filenames_in)) {}
 };
 
 // Helper to serve a static file.
@@ -101,20 +110,27 @@ struct ServeStaticFilesFromOptions {
 struct StaticFileServer {
   std::string content;
   std::string content_type;
-  bool url_path_points_to_directory;
-  StaticFileServer(std::string content, std::string content_type, bool url_path_points_to_directory)
+  bool serves_directory;
+  std::string trailing_slash_redirect_url;
+
+  StaticFileServer(std::string content,
+                   std::string content_type,
+                   bool serves_directory,
+                   std::string trailing_slash_redirect_url = "")
       : content(std::move(content)),
-        content_type(std::move(content_type)),
-        url_path_points_to_directory(url_path_points_to_directory) {}
+        content_type(content_type),
+        serves_directory(serves_directory),
+        trailing_slash_redirect_url(trailing_slash_redirect_url) {}
+
   void operator()(Request r) {
     if (r.method == "GET") {
-      if (url_path_points_to_directory == r.url_path_had_trailing_slash) {
+      if (serves_directory == r.url_path_had_trailing_slash) {
         // 1) Respond with the content if we're serving a directory and have a trailing slash. Example: `/static/`
         // (`static` is a directory, not a file).
         // 2) Respond with the content if we're serving a file and don't have a trailing slash. Example:
         // `/static/index.html`, `/static/file.png`.
         r.connection.SendHTTPResponse(content, HTTPResponseCode.OK, content_type);
-      } else if (!url_path_points_to_directory && r.url_path_had_trailing_slash) {
+      } else if (!serves_directory && r.url_path_had_trailing_slash) {
         // Respond with HTTP 404 Not Found if we're serving a file and have a trailing slash. Example:
         // `/static/index.html/`.
         r.connection.SendHTTPResponse(current::net::DefaultNotFoundMessage(),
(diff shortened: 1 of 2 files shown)


Blocks/HTTP/impl/posix_server.h
```diff
@@ -88,33 +88,49 @@ struct ServeStaticFilesFromCannotServeMoreThanOneIndexFile : ServeStaticFilesExc
...
+  // User-facing URL prefix. Used to prefix the route in the trailing slash redirects.
+  std::string public_url_prefix;
+
+  // Names of files to serve if a directory URL is requested, in the priority order (first found will be served).
   std::vector<std::string> index_filenames;
 
-  explicit ServeStaticFilesFromOptions(std::string url_base = "/",
-                                       std::vector<std::string> index_filenames = {"index.html", "index.htm"})
-      : url_base(std::move(url_base)), index_filenames(std::move(index_filenames)) {}
+  explicit ServeStaticFilesFromOptions(std::string route_prefix_in = "/",
```
dkorolev: `explicit` is redundant here.

Blocks/HTTP/impl/posix_server.h
```diff
@@ -126,7 +142,7 @@ struct StaticFileServer {
         // URL for the index file served at that directory URL (without filename).
         // See RFC1808, `Resolving Relative URLs`, `Step 6`: https://www.ietf.org/rfc/rfc1808.txt
         r.connection.SendHTTPResponse(
-            "", HTTPResponseCode.Found, content_type, current::net::http::Headers({{"Location", r.url.path + '/'}}));
+            "", HTTPResponseCode.Found, content_type, current::net::http::Headers({{"Location", trailing_slash_redirect_url}}));
```
dkorolev: Line too long.

Blocks/HTTP/test.cc
```diff
@@ -76,6 +76,11 @@ DEFINE_int32(net_api_test_port,
              "Local port to use for the test API-based HTTP server. NOTE: This port should be different from "
              "ports in other network-based tests, since API-driven HTTP server will hold it open for the whole "
              "lifetime of the binary.");
+DEFINE_int32(net_api_test_port_secondary,
+             PickPortForUnitTest(),
+             "Local port to use for the test API-based HTTP server for multi-port tests. NOTE: This port should be different from "
```
dkorolev: Line too long.

Blocks/HTTP/test.cc
```diff
@@ -985,96 +1057,195 @@ TEST(HTTPAPI, ServeStaticFilesFrom) {
...
+
+  // Redirect from subsubdirectory without trailing slash to subsubdirectory with trailing slash.
+  {
+    ASSERT_THROW(HTTP(GET(Printf("http://localhost:%d/static/something/sub_dir/sub_sub_dir", FLAGS_net_api_test_port))),
+                 HTTPRedirectNotAllowedException);
+    const auto sub_sub_dir_response = HTTP(GET(Printf("http://localhost:%d/static/something/sub_dir/sub_sub_dir", FLAGS_net_api_test_port)).AllowRedirects());
+    EXPECT_EQ(200, static_cast<int>(sub_sub_dir_response.code));
+    EXPECT_EQ(Printf("http://localhost:%d/anything/sub_dir/sub_sub_dir/", FLAGS_net_api_test_port_secondary), sub_sub_dir_response.url);
+    EXPECT_EQ("Done.", sub_sub_dir_response.body);
+  }
```
dkorolev: There are too many too long lines here. I'll merge it regardless.

sompylasar: What's the issue with the long lines? Why should I terminate a line at a point where it hasn't yet said what it had to say? Broken lines lead to harder merge conflict resolution because they get mixed up by the source control diffing algorithms that are line-based. This specific C++ coding style where the lines break at different logical points does not help to minimize the diffs and avoid/simplify merge conflict resolution.

dkorolev: You're free to break the line before the limit.

sompylasar: Yes I am. The lines broken like that will be harder to scan vertically, there will be a zigzag two times longer that will have to be consumed even if you don't need to read it in full. And even if I format the code for readability, the next clang-format run will squash it.

sompylasar: By the way, I haven't heard any objective argument that shows the benefits of this coding style and alignment.

mzhurovich: Can't we all just stop arguing and agree on keep using `clang-format` with its current settings?

@sompylasar The shortest answer is that the line size is adjusted to fit GH window without horizontal scrolling.

sompylasar: 1. There is no ancient version of clang-format for macOS.
2. OK, line width for GitHub, that's nice. What about the alignment format? 

> I have formatted the code that fitted the line length, still these short lines got messed up, too.
>
> The way the function arguments are aligned in this coding style by shifting them to the end of the function name 1) don't contribute to line shortness; 2) allow multiple ways of aligning the same code shape based on the name width; 3) cause a lot more lines to change when the name changes.

mzhurovich: @sompylasar Sorry, we'll try to solve the problem with upgrading clang-format version as soon as we have capacity for it :( And yes, we know that sometimes the formatted output is not the best way to represent the code. If this is the case, I usually add `clang-format off/on` comments for the particular part of the code.
Thanks for good questions and your contribution to the project!

sompylasar: @mzhurovich Thanks for understanding.

PR #754, opened 2017-06-15
Blocks/HTTP/impl/posix_client: Fixed following redirect from a URL with non-default port to a full URL without a port.

sompylasar:
When a redirect occurs from a URL with a port ("previous URL") to a URL without a port, the port was propagated to the resulting URL and to the "Location" header.
Example:
```
test.cc:156: Failure
Value of: URL("ftp://host_no_port_path", URL("meh://localhost:8080")).ComposeURL()
  Actual: "ftp://host_no_port_path:8080/"
Expected: "ftp://host_no_port_path/"
test.cc:157: Failure
Value of: URL("blah://new_host/foo", URL("meh://localhost:5000")).ComposeURL()
  Actual: "blah://new_host:5000/foo"
Expected: "blah://new_host/foo"
```

The redirect URL transformation logic was implemented in the `URL` class constructor.
It was so multi-purpose (both for redirects and for assigning default values) so that there was no easy way to reason about it when fixing this port propagation issue and keep the rest of test cases green.
The multi-inheritance structure of the `URL` class added more to that spaghetti.
The above led to the decision to simplify the `URL` class public API and internal implementation.
* Removed extra parameters to reduce the API surface that requires support and testing.
* Simplified internal logic (it's still complex and hard to grasp, unfortunately, because we chose to support many partial URL formats and make it smart).
* Made explicit static functions for redirect handling and for filling with default values.
* Simplified `URL` class inheritance, using static functions, not constructors/initializers, to easier propagate data from one to another without a need for a class field (removed `url_without_parameters` field).
* Instead of public `url_without_parameters` field, added `ComposeURLWithoutParameters` method because the field value is hard to keep in sync with the other properties' values (the scheme, host, port).
* Added `ParseQueryString` static function to label that intent where needed (see `Midichlorians/Server/server.h`).

The full stack of the redirect with HTTP server and client cannot be tested with the Current tests setup for now because the spawned servers always use ports and cannot redirect to a URL without a port because `HTTPClient` tries to connect to it. There is a way to test the actual use case: to bind to the port `80`, but this requires root access. I've implemented this test and checked with root access but disabled it with `#if 0` for now.

### DIFF
--- a/Blocks/URL/url.h
+++ b/Blocks/URL/url.h
@@ -70,6 +70,7 @@ namespace impl {
 
 namespace {
 const char* const kDefaultScheme = "http";
+const char* const kDefaultHost = "localhost";
 }
 
 struct URLWithoutParametersParser {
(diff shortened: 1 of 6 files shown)


Blocks/URL/url.h
```diff
@@ -361,29 +356,82 @@ struct URLParametersExtractor {
...
+      url.port = target_url_parsed.port;
+    }
+
+    url.path = target_url_parsed.path;
+
+    url.parameters_vector = target_url_parsed.parameters_vector;
+    url.query = target_url_parsed.query;
+    url.fragment = target_url_parsed.fragment;
+
+    return MakeURLWithDefaults(url);
```
dkorolev: :+1:

Blocks/URL/url.h
```diff
@@ -361,29 +356,82 @@ struct URLParametersExtractor {
...
+  }
+
+  static URL MakeRedirectedURL(const URL& source_url_parsed, const std::string& url) {
+    return URL::MakeRedirectedURL(source_url_parsed, URL(url));
+  }
+
+  // Parses query strings without the question mark and `application/x-www-form-urlencoded` HTTP bodies.
+  static const std::map<std::string, std::string> ParseQueryString(const std::string& body) {
```
dkorolev: I'm sure we could use it in a few other places, where the `URL("/?blahblahblah")` hack is employed now. Thanks @sompylasar.

dkorolev:
Resolved the conflict manually.

PR #757, opened 2017-06-21
Replication benchmarks - generic scenario and a separate one.

grixa:
.

### DIFF
--- a/examples/Benchmark/generic/scenario_replication.h
+++ b/examples/Benchmark/generic/scenario_replication.h
@@ -36,9 +36,9 @@
 #ifndef CURRENT_MAKE_CHECK_MODE
 DEFINE_string(remote_url, "", "Remote url to subscribe to.");
 DEFINE_uint16(local_port, 8383, "Local port to spawn the source stream on.");
-DEFINE_string(db, "db.json", "Path to load the source stream data from.");
+DEFINE_string(db, "data.json", "Path to load the source stream data from.");
 DEFINE_bool(regenerate_db, true, "Regenerate the stream data to replicate.");
-DEFINE_bool(use_file_persistence, true, "Use file persisters in the replicated streams.");
+DEFINE_string(persister, "disk", "The type of the replicator-side persister - one of 'disk' / 'memory'");
 DEFINE_uint32(entry_length, 100, "The length of the string member values in the generated stream entries.");
 DEFINE_uint32(entries_count, 1000, "The number of entries to replicate.");
 #else
@@ -46,7 +46,7 @@ DECLARE_string(remote_url);
 DECLARE_uint16(local_port);
 DECLARE_string(db);
 DECLARE_bool(regenerate_db);
-DECLARE_bool(use_file_persistence);
+DECLARE_string(persister);
 DECLARE_uint32(entry_length);
 DECLARE_uint32(entries_count);
 #endif
@@ -55,6 +55,8 @@ SCENARIO(stream_replication, "Replicate the Current stream of simple string entr
   std::unique_ptr<benchmark::replication::stream_t> stream;
   std::string stream_url;
   HTTPRoutesScope scope;
+  enum class PERSISTER_TYPE : int { DISK, MEMORY };
+  PERSISTER_TYPE persister_type;
 
   stream_replication() {
     if (FLAGS_remote_url.empty()) {
@@ -69,6 +71,13 @@ SCENARIO(stream_replication, "Replicate the Current stream of simple string entr
                    .Register("/raw_log", URLPathArgs::CountMask::None | URLPathArgs::CountMask::One, *stream);
     } else {
       stream_url = FLAGS_remote_url;
+    }
+    const std::map<std::string, PERSISTER_TYPE> supported_persisters = {{"disk", PERSISTER_TYPE::DISK},
+                                                                        {"memory", PERSISTER_TYPE::MEMORY}};
+    try {
+      persister_type = supported_persisters.at(FLAGS_persister);
+    } catch (const std::out_of_range&) {
+      throw std::logic_error(current::strings::Printf("Unsupported persister type: %s", FLAGS_persister.c_str()));
     }
   }
 
@@ -87,7 +96,7 @@ SCENARIO(stream_replication, "Replicate the Current stream of simple string entr
   }
 
   void RunOneQuery() override {
-    if (FLAGS_use_file_persistence) {
+    if (persister_type == PERSISTER_TYPE::DISK) {
       const auto filename = current::FileSystem::GenTmpFileName();
       const auto replicated_stream_file_remover = current::FileSystem::ScopedRmFile(filename);
       Replicate<benchmark::replication::stream_t>(filename);
--- a/examples/Benchmark/generic/scenario_replication.h
+++ b/examples/Benchmark/generic/scenario_replication.h
@@ -57,20 +57,24 @@ SCENARIO(stream_replication, "Replicate the Current stream of simple string entr
   HTTPRoutesScope scope;
   enum class PERSISTER_TYPE : int { DISK, MEMORY };
   PERSISTER_TYPE persister_type;
+  uint64_t entries_count;
 
   stream_replication() {
     if (FLAGS_remote_url.empty()) {
       if (FLAGS_regenerate_db) {
         current::FileSystem::RmFile(FLAGS_db, current::FileSystem::RmFileParameters::Silent);
         stream = benchmark::replication::GenerateStream(FLAGS_db, FLAGS_entry_length, FLAGS_entries_count);
+        entries_count = FLAGS_entries_count;
       } else {
         stream = std::make_unique<benchmark::replication::stream_t>(FLAGS_db);
+        entries_count = stream->Persister().Size();
       }
-      stream_url = current::strings::Printf("127.0.0.1:%u/raw_log", FLAGS_local_port);
-      scope += HTTP(FLAGS_local_port)
-                   .Register("/raw_log", URLPathArgs::CountMask::None | URLPathArgs::CountMask::One, *stream);
+      stream_url = current::strings::Printf("127.0.0.1:%u/", FLAGS_local_port);
+      scope +=
+          HTTP(FLAGS_local_port).Register("/", URLPathArgs::CountMask::None | URLPathArgs::CountMask::One, *stream);
     } else {
       stream_url = FLAGS_remote_url;
+      entries_count = current::FromString<uint64_t>(HTTP(GET(stream_url + "?sizeonly")).body);
     }
     const std::map<std::string, PERSISTER_TYPE> supported_persisters = {{"disk", PERSISTER_TYPE::DISK},
                                                                         {"memory", PERSISTER_TYPE::MEMORY}};
@@ -88,7 +92,7 @@ SCENARIO(stream_replication, "Replicate the Current stream of simple string entr
     auto replicator = std::make_unique<current::sherlock::StreamReplicator<STREAM>>(replicated_stream);
     {
       const auto subscriber_scope = remote_stream.Subscribe(*replicator);
-      while (replicated_stream.Persister().Size() != FLAGS_entries_count) {
+      while (replicated_stream.Persister().Size() != entries_count) {
         std::this_thread::yield();
       }
     }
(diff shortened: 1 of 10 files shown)


examples/Benchmark/generic/scenario_replication.h
```diff
@@ -0,0 +1,102 @@
...
+#include "../../../Sherlock/replicator.h"
+
+#include "../../../Bricks/dflags/dflags.h"
+
+#ifndef CURRENT_MAKE_CHECK_MODE
+DEFINE_string(remote_url, "", "Remote url to subscribe to.");
+DEFINE_uint16(local_port, 8383, "Local port to spawn the source stream on.");
+DEFINE_string(db, "db.json", "Path to load the source stream data from.");
+DEFINE_bool(regenerate_db, true, "Regenerate the stream data to replicate.");
+DEFINE_bool(use_file_persistence, true, "Use file persisters in the replicated streams.");
```
dkorolev: Perhaps `DEFINE_string(persister, "disk", "One of 'disk' / 'memory'")`?

examples/Benchmark/replication/replication_server.cc
```diff
@@ -0,0 +1,52 @@
...
+int main(int argc, char** argv) {
+  ParseDFlags(&argc, &argv);
+  std::unique_ptr<benchmark::replication::stream_t> stream;
+  if (FLAGS_regenerate_db) {
+    current::FileSystem::RmFile(FLAGS_db, current::FileSystem::RmFileParameters::Silent);
+    stream = benchmark::replication::GenerateStream(FLAGS_db, FLAGS_entry_length, FLAGS_entries_count);
+  } else {
+    stream = std::make_unique<benchmark::replication::stream_t>(FLAGS_db);
+  }
+  auto scope =
```
dkorolev: `const`

examples/Benchmark/replication/replication_client.cc
```diff
@@ -0,0 +1,63 @@
...
+void Replicate(ARGS&&... args) {
+  STREAM replicated_stream(std::forward<ARGS>(args)...);
+  using RemoteStreamReplicator = current::sherlock::StreamReplicator<STREAM>;
+  current::sherlock::SubscribableRemoteStream<benchmark::replication::Entry> remote_stream(FLAGS_url);
+  auto replicator = std::make_unique<RemoteStreamReplicator>(replicated_stream);
+
+  const auto start_time = std::chrono::system_clock::now();
+  {
+    const auto subscriber_scope = remote_stream.Subscribe(*replicator);
+    while (replicated_stream.Persister().Size() != FLAGS_total_entries) {
```
dkorolev: Perhaps add a check, before the measured section, that the side of the stream to be replicating from matches `--total_entries`?

examples/Benchmark/replication/replication_client.cc
```diff
@@ -0,0 +1,63 @@
...
+      std::this_thread::yield();
+    }
+  }
+  const auto duration =
+      std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now() - start_time).count();
+  std::cout << "Replication filished, total time: " << duration / 1000.0 << " seconds." << std::endl;
```
dkorolev: Why not output EPS (entries per second) and MBPS too? Also, perhaps in some TSV-friendly format, to make it easier to batch-run multiple scenarios via a shell script and `cut -f #` the results?

examples/Benchmark/replication/replication_client.cc
```diff
@@ -0,0 +1,63 @@
...
+
+DEFINE_string(url, "127.0.0.1:8383/raw_log", "Url to subscribe to.");
+DEFINE_string(db, "replicated_db.json", "Path to load the source stream data from.");
+DEFINE_uint32(total_entries, 10000, "Entries number to replicate.");
+
+template <typename STREAM, typename... ARGS>
+void Replicate(ARGS&&... args) {
+  STREAM replicated_stream(std::forward<ARGS>(args)...);
+  using RemoteStreamReplicator = current::sherlock::StreamReplicator<STREAM>;
+  current::sherlock::SubscribableRemoteStream<benchmark::replication::Entry> remote_stream(FLAGS_url);
```
dkorolev: For a standlone tool, perhaps more console output would help here? Ex. `"Connecting to the stream at $URL ..." << flush`, followed by some `"\b\b\bOK\nValidating schema ..."`, and then the number of records in the remote stream?

examples/Benchmark/replication/generate_stream.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace benchmark {
+namespace replication {
+
+using stream_t = current::sherlock::Stream<Entry, current::persistence::File>;
+
+inline std::unique_ptr<stream_t> GenerateStream(const std::string& output_file,
+                                                uint32_t entry_length,
+                                                uint32_t entries_count) {
+  const char symbols[] = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
dkorolev: Question: Why not subtract those few characters (effectively, `JSON(Entry()).length()`), to keep entries exactly the size we'd expect them to be? Perhaps even subtract `(JSON(Entry()).length() + 1)`, to include the `'\n'` in a `curl $URL/raw_log?entries_only&nowait >log.txt`?

examples/Benchmark/replication/entry.h
```diff
@@ -0,0 +1,43 @@
...
+#include "../../../TypeSystem/struct.h"
+
+namespace benchmark {
+namespace replication {
+
+CURRENT_STRUCT(Entry) {
+  CURRENT_FIELD(s, std::string, "");
+
+  CURRENT_DEFAULT_CONSTRUCTOR(Entry) {}
+  CURRENT_CONSTRUCTOR(Entry)(std::string s) : s(std::move(s)) {}
```
dkorolev: :+1:

examples/Benchmark/replication/generate_stream.h
dkorolev: With an assert that `entry_length` should be at least that, of course.

examples/Benchmark/generic/scenario_storage.h
```diff
@@ -54,8 +54,7 @@ CURRENT_STRUCT(UInt32KeyValuePair) {
 CURRENT_STRUCT(StringKeyValuePair) {
   CURRENT_FIELD(key, std::string);
   CURRENT_FIELD(value, uint32_t);
-  CURRENT_CONSTRUCTOR(StringKeyValuePair)(const std::string& key = "", uint32_t value = 0)
-      : key(key), value(value) {}
+  CURRENT_CONSTRUCTOR(StringKeyValuePair)(const std::string& key = "", uint32_t value = 0) : key(key), value(value) {}
```
dkorolev: Thanks for `make indent`! Perhaps change this to "pass by copy, initialize by move" too?

examples/Benchmark/generic/scenario_replication.h
```diff
@@ -0,0 +1,102 @@
...
+    if (FLAGS_remote_url.empty()) {
+      if (FLAGS_regenerate_db) {
+        current::FileSystem::RmFile(FLAGS_db, current::FileSystem::RmFileParameters::Silent);
+        stream = benchmark::replication::GenerateStream(FLAGS_db, FLAGS_entry_length, FLAGS_entries_count);
+      } else {
+        stream = std::make_unique<benchmark::replication::stream_t>(FLAGS_db);
+      }
+      stream_url = current::strings::Printf("127.0.0.1:%u/raw_log", FLAGS_local_port);
+      scope += HTTP(FLAGS_local_port)
+                   .Register("/raw_log", URLPathArgs::CountMask::None | URLPathArgs::CountMask::One, *stream);
```
dkorolev: With all the flags, might as well introduce one for `"/raw_log"`?

examples/Benchmark/generic/scenario_replication.h
```diff
@@ -0,0 +1,102 @@
...
+    } else {
+      stream_url = FLAGS_remote_url;
+    }
+  }
+
+  template <typename STREAM, typename... ARGS>
+  void Replicate(ARGS && ... args) {
+    STREAM replicated_stream(std::forward<ARGS>(args)...);
+    using RemoteStreamReplicator = current::sherlock::StreamReplicator<STREAM>;
```
dkorolev: Used only in one place -- is it worth it?

examples/Benchmark/generic/scenario_replication.h
```diff
@@ -0,0 +1,102 @@
...
+#include "../../../Blocks/HTTP/api.h"
+#include "../../../Sherlock/replicator.h"
+
+#include "../../../Bricks/dflags/dflags.h"
+
+#ifndef CURRENT_MAKE_CHECK_MODE
+DEFINE_string(remote_url, "", "Remote url to subscribe to.");
+DEFINE_uint16(local_port, 8383, "Local port to spawn the source stream on.");
+DEFINE_string(db, "db.json", "Path to load the source stream data from.");
+DEFINE_bool(regenerate_db, true, "Regenerate the stream data to replicate.");
```
dkorolev: Suggestion (CC @mzhurovich): Perhaps use two flags instead, some `--use_db` or `--tmp_db`. If `--use_db` is unset, the temporary one is generated from scratch is used, and that's the default behavior. If the user sets `--use_db` (or `--external_db` or `--existing_db` or `--use_existing_db`, I'm bad at naming flags), it's used untouched.

mzhurovich: @dkorolev Not sure I fully understand your idea. What two flags exactly, `--db` and `--use_db`? What's the desired behavior? Could you elaborate pls?

examples/Benchmark/generic/scenario_replication.h
mzhurovich: +1

examples/Benchmark/generic/scenario_replication.h
grixa: But for what purpose? It's used internally, and for the case with the remote stream instead of local one there is the `remote_url` argument. I suggest to remove "/raw_log" and leave just "127.0.0.1:<port>/".

examples/Benchmark/generic/scenario_replication.h
grixa: I also don't fully understand. What difference between the `--use_db` flag in your suggestion and current `--regenerate_db`?

dkorolev: The idea was to make sure there's no "read-write" flag.

in the present setup it's "--db", which _may or may not_ be overwritten if another flag is set. I may add the other flag forgetting of `--db` being set earlier in the command (stretch, I know, but still), and surprise myself later on realizing I've overwritten the DB file, possibly an important one. Practically speaking, I'd be reluctant to use `--db` of this tool to point to some prod database file, just because I'm too scared it can be overwritten.

And two flags eliminate such a risk altogether. That was my thinking :)

dkorolev:
LGTM again.

My only concern before merging this in is the flag reuse for read-only vs. read-write DB file. Do you guys think I'm overly paranoid there? :)

examples/Benchmark/generic/scenario_replication.h
grixa: Good point, didn't think about that.

dkorolev:
OK, not I'm lost even more :)

@grixa, `main()` in `examples/Benchmark/replication/replication_client.cc` will erase the `FLAGS_db` file no mater what :/

And, while you're generating a temporary file name for the db now, there are two flags -- `--db` and `--existing_db`, for the function that can be handled with just `--db`: if unset, generate a temporary one, and if set, use the existing one and do not overwrite it.

Am I not seeing anything important? :)

dkorolev:
s/not/now/

grixa:
Well, I thought it's natural for the replication client to remove the db before start writing it. Because if not, the second execution of the benchmark with the same args leads to the wrong result. An alternative approach is to use tmp files and remove them each time, but in this case we can't look inside and check the content.

dkorolev:
> An alternative approach is to use tmp files and remove them each time, but in this case we can't look inside and check the content.

Some `--keep_tmpdb_contents` would do the trick, right?

dkorolev:
... but I'm merging it regardless!

PR #766, opened 2017-07-13
scripts/individual-tests.sh: Exclude filter for the batch test run.

sompylasar:
- Useful when tracking the tests which may implicitly affect each other.

_Includes https://github.com/C5T/Current/pull/765 to avoid merge conflicts on adjacent lines._

### DIFF
--- a/scripts/individual-tests.sh
+++ b/scripts/individual-tests.sh
@@ -8,8 +8,7 @@ FAILURES=""
 
 for i in $(find . -name test.cc | sort -g) ; do
   DIR=$(dirname $i)
-  echo -e -n "\n\033[0m\033[1mDir\033[0m: \033[36m"
-  echo $DIR
+  echo -e "\n\033[0m\033[1mDir\033[0m: \033[36m${DIR}\033[0m"
   cd $DIR
   if ! make -s test ; then
     FAILURES="$FAILURES\n- $DIR"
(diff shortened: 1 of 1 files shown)


scripts/individual-tests.sh
```diff
@@ -6,15 +6,21 @@ set -u -e
...
+
 for i in $(find . -name test.cc | sort -g) ; do
   DIR=$(dirname $i)
-  echo -e -n "\n\033[0m\033[1mDir\033[0m: \033[36m"
-  echo $DIR
-  cd $DIR
-  if ! make -s test ; then
-    FAILURES="$FAILURES\n- $DIR"
+  echo -e "\n\033[0m\033[1mDir\033[0m: \033[36m${DIR}\033[0m"
+  if [[ " ${EXCLUDE[@]} " =~ " $DIR " ]]; then
```
dkorolev: I'm no bash expert, but: why the spaces?

sompylasar: @dkorolev That's not bash-specific. For the edge case of matching the first and the last item: always match an item which has space from both sides.

For example, it was done the same way in JavaScript (jQuery) to find an item in a space-separated className string without splitting and building dynamic regexps before the `classList.has()` API was made available.

dkorolev: ЩЛ

dkorolev: OK

PR #772, opened 2017-07-23
Make reflection great again

dkorolev:
Hi Max!

I managed to push it through. Proud on the one hand, quite a few fixes left to be done on the other one.

The good:
* It builds and passes the tests.
* `TypeID`-s of our prod remain unchanged.

Remains:
1. Schema export of self-referring structs. I've just disabled the respective test for now.
1. Pure C++ (native `struct`-s) exporter would result in the code that does not compile, as the short names (`_Z`, no digits) of template-instantiated structs are being used.
1. That extension to the serialization/deserialization of the `Variant` in the `Minimalistic` format (early runtime failure).
1. `Variant`-s magic (export the schema of `Variant<Templated<One>, Templated<Two>>` into `.h`, confirm it restores into itself).
1. Test the evolutors (this may be best to do on live data).
1. The top-level `make test` doesn't build (looks like an extra/missing bracket).
1. After this PR is done with, merging it into `alt-master` could be a minor pain too.

To get the schema to play with I've used `curl localhost:8283/raw_log/schema.h >full.h`.

Thanks,
Dima


dkorolev:
Status update: While the reflector itself (the `CurrentTypeID<>` part, and the recursive semi-compile-time magic associated with it) seems to do its job well, the thread-local `Reflector()`, which is used to generate our type-describing structures (and, ultimately, the C++ et. al. exported schemas) needs some TLC.

I've localized the issue in the few hours I could dedicate to this issue today (https://github.com/dkorolev/Current/commit/8cd46bd553d38b59ae47d7dc309a89ffa5854585). Will resume tomorrow.

dkorolev:
I should also note that `CurrentTypeName<>` is used as part of `CurrentTypeID<>`, as, when the cycle has to be terminated, it's `CRC(CurrentTypeName<T>())` now. I don't think it's a bad solution by itself, just to keep in mind that `CurrentTypeName` is also something we shouldn't change moving forward (assuming we're cool with this design).

dkorolev:
And this:

> `Error in `pthread_key_create()`. Terminating.libc++abi.dylib: cannot create pthread key for __cxa_get_globals()`

from [here](https://travis-ci.org/C5T/Current/jobs/257583579) does bug me. Investigating, although it's likely I won't come to any solution today.

Max, do you know anything about pthread-based thread locals? Perhaps we could just use C++'s native `thread_local` in Mac? Or perhaps in case we do need a custom one we could simply re-implement it using one `pthread` call per thread, not one per { thread * singleton type } ?

dkorolev:
Also, @mzhurovich, could you please test this on Windows?

mzhurovich:
@dkorolev This error on Travis looks like system resources exhaustion.
Our TLS implementation via pthread was written because at that time clang on OS X didn't support `thread_local`. I've checked the recent version of XCode and seems it works fine now :)

dkorolev:
> Our TLS implementation via pthread was written because at that time clang on OS X didn't support thread_local. I've checked the recent version of XCode and seems it works fine now :)

Yay! :+1:

dkorolev:
Deprecated via https://github.com/C5T/Current/pull/780.

PR #774, opened 2017-07-25
Clang warning

agserm:
Found this warning cropping up on clang 3.9. The new code is a bit more verbose but does not invoke undefined behavior in the preprecessor.


dkorolev:
:+1:

PR #775, opened 2017-07-25
Fixed the replication benchmark scenario.

grixa:


dkorolev:
Phew. That was easy :)

PR #778, opened 2017-07-29
Gradient boosted trees WIP.

dkorolev:


dkorolev:
Closing to re-create the `gbt` pull request (which is the simple one) after resolving the reflection one (which is much harder to implement now as I've locallized the compilation time / RAM footprint issue).

PR #779, opened 2017-07-30
Unsafe stream subscription

grixa:
- Unsafe subscription to a stream, remote and local. Used by default via HTTP (switches to the "safe" one by adding "checked" flag to the subscription URL).
- Unsafe stream replication (and as a part of it - unsafe publishing to the file persister).

The replication benchmark shows the following on my machine:
100000 entries, 1000 bytes length each:
```
// old logic, parse JSONs on both sides
.current/replication_client --use_safe_replication --use_checked_subscription 
Seconds	EPS	MBps
3.59469	27818.8	26.53
// "unsafe" subscription, but parse received values before writing them on the client side 
.current/replication_client --use_safe_replication 
Seconds	EPS	MBps
3.08612	32403.1	30.902
// "unsafe" logic on both sides (parse only idxts_t in the FilePersister)
.current/replication_client
Seconds	EPS	MBps
2.72819	36654.3	34.9563
```
1000000 entries, 100 bytes length each:
```
// old logic, parse JSONs on both sides
.current/replication_client --use_safe_replication --use_checked_subscription 
Seconds	EPS	MBps
18.6243	53693.4	5.1206
// "unsafe" subscription, but parse received values before writing them on the client side 
.current/replication_client --use_safe_replication 
Seconds	EPS	MBps
15.0389	66494.3	6.34139
// "unsafe" logic on both sides (parse only idxts_t in the FilePersister)
.current/replication_client
Seconds	EPS	MBps
13.9142	71868.8	6.85394
```
I have several concerns:
  1. Naming :) May be it would be better in case of subscription to call it "unchecked" rather than "unsafe".
  2. UpdateHead produce some troubles: it is transferred the same way for both subscription types and the receiver should check that there is no '\t' and then parse it. Haven't any ideas of better solution for now.
  3. Some subscription options, such as `period`, require to parse idxts_t part of an each line to extract the current entry timestamp value. As I understand, we can't use the fact that each timestamp is stored in the FilePersister, because it's an implementation detail, that could be changed any moment. So, these options make the "unchecked" subscription less effective. Should we prohibit them or produce some kind of warning?
  4. When we subscribe (in "checked" mode) to the stream with "spoiled" data, it doesn't throw an exception. It crashes instead (in fact it trows an exception, but in a separate thread), so the `EXPECT_THROW(spoiled_stream.Subscribe(...))` doesn't work. Because of that I can't demonstrate an actual comparison in the `UnsafeVsSafeSubscription` test. Is it OK?

### DIFF
--- a/examples/benchmark/replication/replication_client.cc
+++ b/examples/benchmark/replication/replication_client.cc
@@ -67,9 +67,11 @@ void Replicate(ARGS&&... args) {
   {
     std::cerr << "Subscribing to the stream ..." << std::flush;
     const std::chrono::milliseconds print_delay(500);
-	  const auto subscriber_scope = FLAGS_use_safe_replication
-		? (current::stream::SubscriberScope)remote_stream.Subscribe(*replicator, 0, FLAGS_use_checked_subscription)
-		: (current::stream::SubscriberScope)remote_stream.SubscribeUnsafe(*replicator, 0, FLAGS_use_checked_subscription);
+    const auto subscriber_scope =
+        FLAGS_use_safe_replication
+            ? (current::stream::SubscriberScope)remote_stream.Subscribe(*replicator, 0, FLAGS_use_checked_subscription)
+            : (current::stream::SubscriberScope)remote_stream.SubscribeUnsafe(
+                  *replicator, 0, FLAGS_use_checked_subscription);
     std::cerr << "\b\b\bOK" << std::endl;
     auto next_print_time = start_time + print_delay;
     while (replicated_stream->Data()->Size() < records_to_replicate) {
--- a/examples/benchmark/replication/replication_client.cc
+++ b/examples/benchmark/replication/replication_client.cc
@@ -67,11 +67,11 @@ void Replicate(ARGS&&... args) {
   {
     std::cerr << "Subscribing to the stream ..." << std::flush;
     const std::chrono::milliseconds print_delay(500);
-    const auto subscriber_scope =
-        FLAGS_use_safe_replication
-            ? (current::stream::SubscriberScope)remote_stream.Subscribe(*replicator, 0, FLAGS_use_checked_subscription)
-            : (current::stream::SubscriberScope)remote_stream.SubscribeUnsafe(
-                  *replicator, 0, FLAGS_use_checked_subscription);
+    const auto subscriber_scope = FLAGS_use_safe_replication
+                                      ? static_cast<current::stream::SubscriberScope>(
+                                            remote_stream.Subscribe(*replicator, 0, FLAGS_use_checked_subscription))
+                                      : static_cast<current::stream::SubscriberScope>(remote_stream.SubscribeUnsafe(
+                                            *replicator, 0, FLAGS_use_checked_subscription));
     std::cerr << "\b\b\bOK" << std::endl;
     auto next_print_time = start_time + print_delay;
     while (replicated_stream->Data()->Size() < records_to_replicate) {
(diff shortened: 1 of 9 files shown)


examples/benchmark/replication/replication_client.cc
```diff
@@ -65,7 +67,11 @@ void Replicate(ARGS&&... args) {
   {
     std::cerr << "Subscribing to the stream ..." << std::flush;
     const std::chrono::milliseconds print_delay(500);
-    const auto subscriber_scope = remote_stream.Subscribe(*replicator);
+    const auto subscriber_scope =
+        FLAGS_use_safe_replication
+            ? (current::stream::SubscriberScope)remote_stream.Subscribe(*replicator, 0, FLAGS_use_checked_subscription)
```
dkorolev: С++-style cast please.

stream/pubsub.h
```diff
@@ -335,6 +340,89 @@ class PubSubHTTPEndpointImpl : public AbstractSubscriberObject {
...
+        return current_us;
+      };
+      if (!serving_) {
+        if (current_index >= params_.i &&                                            // Respect `i`.
+            (params_.tail == 0u || (last.index - current_index) < params_.tail) &&   // Respect `tail`.
+            (from_timestamp_.count() == 0u || GetCurrentUs() >= from_timestamp_)) {  // Respect `since` and `recent`.
+          serving_ = true;
+        }
+        // Reached the end, didn't started serving and should not wait.
+        if (!serving_ && current_index == last.index && params_.no_wait) {
```
dkorolev: This block is already in `if (!serving_)`.

dkorolev: And `>=` for index looks scary to me. Shouldn't it always increase by just one?

At the same time, `current_index == last.index ` looks too strict, what if we started from past the end for some reason?

stream/replicator.h
```diff
@@ -152,6 +156,48 @@ class SubscribableRemoteStream final {
       }
     }
 
+    template <SubscriptionMode MODE>
+    ENABLE_IF<MODE == SubscriptionMode::Safe, void> PassEntriesToSubscriber(const std::vector<std::string>& lines,
```
dkorolev: `, void>` is unnecessary IIRC.

stream/stream.h
```diff
@@ -119,6 +119,8 @@ CURRENT_STRUCT(StreamSchemaFormatNotFoundError) {
 template <typename ENTRY>
 using DEFAULT_PERSISTENCE_LAYER = current::persistence::Memory<ENTRY>;
 
+enum class SubscriptionMode : int { Safe, Unsafe };
```
dkorolev: `{ Unsafe = 0, Safe = 1 }`? (Yes, naming is another topic to talk about.)

stream/stream.h
```diff
@@ -537,14 +576,14 @@ class Stream final {
...
+      };
       current::stream::SubscriberScope http_chunked_subscriber_scope =
-          Subscribe(*http_chunked_subscriber,
-                    begin_idx,
-                    [this, borrowed_impl, subscription_id]() {
-                      // Note: Called from a locked section of `borrowed_impl->http_subscriptions_mutex`.
-                      borrowed_impl->http_subscriptions[subscription_id].second = nullptr;
-                    });
+          request_params.checked
+              ? (current::stream::SubscriberScope)Subscribe(*http_chunked_subscriber, begin_idx, done_callback)
```
dkorolev: C-style cast again.

stream/pubsub.h
grixa: This whole logic is almost identical to the `operator()` for the "safe" subscription case above.
1. `if (!serving_)` makes sense if the condition right above this line just became true and switched the `serving_` to `true`.
2. `current_index >= params_.i`, I suppose, covers the case when for some reason the first entry which is passed here has greater index then required to start serving.
3. `current_index` and `last.index` both passed from outside. So how can the first one possibly be greater then the second one?

dkorolev:
This PR is deprecated by https://github.com/C5T/Current/pull/798, right?

grixa:
Sure

PR #780, opened 2017-08-06
Make reflection great again, and fix compilation times now.

dkorolev:
This PR replaces https://github.com/C5T/Current/pull/772.

TL;DR: Everything we're reviewed in https://github.com/C5T/Current/pull/772, reverted name shortening (last two commits there), and moved the compiler-heavy logic from compile time to runtime.

Confirmed Travis does not die (it used to, on a Mac), and confirmed the compilation time of a heavy project is within 5% of what it was before this refactoring.

@mzhurovich, @sompylasar, @grixa -- could you please review this? The next steps are to merge @sompylasar's updated TypeScript schema changes (into `master`), merge @grixa's unsafe subscriptions (into `alt-master`), and merge `master` and `alt-master`. I could do most of the painful work myself, but this reflection change is too large and too centerpiece that it requires a deeper review.

This reflection change is also worth testing on Windows. @mzhurovich, would you be so kind to find time for it?

Thanks,
Dima

PS: There's an ugly `TODO_DKOROLEV_` `#define`-d symbol. I plan to rename it into the `*_PARANOID_*` one upon merging `master` and `alt-master`.


TypeSystem/Reflection/types.h
```diff
@@ -180,23 +181,35 @@ CURRENT_STRUCT(ReflectedType_Struct_Field) {
...
+  // The `TemplateInnerTypeExpandedName()` method serves two purposes:
+  // 1) It generates the unique name for the structure for the schema-exported `CURRENT_STRUCT_T`-s.
+  //    (As `CURRENT_STRUCT_T`-s are exported as regular `CURRENT_STRUCT`-s, one per instantiation.)
+  // 2) It preserves the TypeID-s the way they used to be before the refactoring of the reflector.
+  std::string TemplateInnerTypeExpandedName() const {
     if (Exists(template_id)) {
-      // TODO(dkorolev): This code relies on `<>` at the end of the names of `CURRENT_STRUCT_T`-s. Fix it.
       CURRENT_ASSERT(native_name.length() > 2);
       CURRENT_ASSERT(native_name.substr(native_name.length() - 2) == "_Z");
       return native_name.substr(0, native_name.length() - 2) + "_T" + current::ToString(Value(template_id));
```
sompylasar: OK, I see now. Can we make a function that returns the exported type name, that can be used on any `ReflectedType_XXX` or `TypeID` from the schema exporters, if that is not `CurrentTypeName`?

TypeSystem/Reflection/types.h
```diff
@@ -180,23 +181,35 @@ CURRENT_STRUCT(ReflectedType_Struct_Field) {
...
-  // 2) it makes sure the the 'autogen -> #include -> autogen' loop does not change TypeIDs of templated types.
-  std::string CanonicalName() const {
+  // The `TemplateInnerTypeExpandedName()` method serves two purposes:
+  // 1) It generates the unique name for the structure for the schema-exported `CURRENT_STRUCT_T`-s.
+  //    (As `CURRENT_STRUCT_T`-s are exported as regular `CURRENT_STRUCT`-s, one per instantiation.)
+  // 2) It preserves the TypeID-s the way they used to be before the refactoring of the reflector.
+  std::string TemplateInnerTypeExpandedName() const {
     if (Exists(template_id)) {
-      // TODO(dkorolev): This code relies on `<>` at the end of the names of `CURRENT_STRUCT_T`-s. Fix it.
       CURRENT_ASSERT(native_name.length() > 2);
```
sompylasar: `> 2` versus `>= 3` below, looks the same.

TypeSystem/Reflection/types.h
```diff
@@ -180,23 +181,35 @@ CURRENT_STRUCT(ReflectedType_Struct_Field) {
...
+  // TemplateFullNameDecorator() returns a function that converts names ending with "_Z"
+  // into the names ending with "_T...", where `...` is the TypeID of the `template_id` of this struct.
+  // Prerequisite: The struct is an instantiation of the template (or is declared so).
+  std::function<std::string(const std::string&)> TemplateFullNameDecorator() const {
+    CURRENT_ASSERT(Exists(template_id));
+    const TypeID t = Value(template_id);
+    return [t](const std::string& input) -> std::string {
+      CURRENT_ASSERT(input.length() >= 3);
+      CURRENT_ASSERT(input.substr(input.length() - 2) == "_Z");
+      return input.substr(0, input.length() - 1) + 'T' + current::ToString(static_cast<uint64_t>(t));
```
sompylasar: Can this be a pure function that takes two arguments and is used both here and above?

TypeSystem/Reflection/types.h
```diff
@@ -216,7 +229,7 @@ inline uint64_t ROL64(const TypeID type_id, size_t nbits) {
 }
 
 inline TypeID CalculateTypeID(const ReflectedType_Struct& s) {
-  uint64_t hash = current::CRC32(s.CanonicalName());
+  uint64_t hash = current::CRC32(s.TemplateInnerTypeExpandedName());
```
sompylasar: This doing CRC32 over a name to get a hash looks repeated in some other place. Consider pulling out this piece of logic into a reusable function if these uses are the same.

TypeSystem/Schema/golden/smoke_test_struct.cc
```diff
@@ -69,7 +69,7 @@ struct TemplatedInheriting_T9200000002835747520 : A {
 };
 struct Templated_T9209626390174323094 {
   int32_t foo;
-  TemplatedInheriting_T9200000002835747520 bar;
+  TemplatedInheriting_Z bar;
```
sompylasar: What if there are two or more templated with different template typenames? Which one of them will this reference?

TypeSystem/Schema/golden/smoke_test_struct.cc
```diff
@@ -106,12 +106,12 @@ struct FullTest {
...
   Variant_B_A_B_B2_C_Empty_E q;
-  Templated_T9209980946934124423 w1;
-  Templated_T9227782344077896555 w2;
-  Templated_T9209626390174323094 w3;
-  TemplatedInheriting_T9209980946934124423 w4;
-  TemplatedInheriting_T9227782344077896555 w5;
-  TemplatedInheriting_T9201673071807149456 w6;
+  Templated_Z w1;
+  Templated_Z w2;
+  Templated_Z w3;
```
sompylasar: These should have different typenames, no?

TypeSystem/Schema/schema.h
```diff
@@ -179,12 +179,18 @@ struct CurrentStructPrinter<CPPLanguageSelector::CurrentStructs> {
...
+    const std::string struct_name = s.native_name;
     os << "CURRENT_STRUCT(" << struct_name;
     if (s.super_id != TypeID::CurrentStruct) {
       os << ", " << type_name(s.super_id, "");
     }
     os << ") {\n";
+    if (Exists(s.template_id)) {
+      CURRENT_ASSERT(s.native_name.length() >= 3u);
+      CURRENT_ASSERT(s.native_name.substr(s.native_name.length() - 2u) == "_Z");
+      os << "  CURRENT_EXPORTED_TEMPLATED_STRUCT(" << s.native_name.substr(0, s.native_name.length() - 2) << ", "
```
sompylasar: This piece should be reused from a single place. This is the third copy of the logic so far.

TypeSystem/Schema/schema.h
```diff
@@ -390,16 +394,26 @@ struct LanguageSyntaxCPP : CurrentStructPrinter<CPP_LANGUAGE_SELECTOR> {
...
+            os_ << "  CURRENT_NAMESPACE_TYPE("
+                << Value<ReflectedType_Struct>(type_substance).TemplateFullNameDecorator()(type_name)
+                << ", current_userspace::" << TypeName(input_type.first) << ");\n";
           }
         }
         if (Exists(namespace_to_expose_)) {
           const NamespaceToExpose& expose = Value(namespace_to_expose_);
           os_ << "\n  // Privileged types.\n";
           for (const auto& t : expose.types) {
+            // NOTE(dkorolev): If the uses requests to expose a templated type, it's their responsibility ]
```
sompylasar: Stray `]` at the end.

TypeSystem/Schema/schema.h
```diff
@@ -390,16 +394,26 @@ struct LanguageSyntaxCPP : CurrentStructPrinter<CPP_LANGUAGE_SELECTOR> {
...
+            // to only do it for at most one instantiation per type.
```
sompylasar: So, templated can be with only one template typename? Then maybe it's easier to drop support for the templated types and force to use explicit compile-time Variants?

TypeSystem/Schema/schema.h
```diff
@@ -1187,9 +1220,7 @@ struct LanguageSyntaxImpl<Language::TypeScript> final {
               oss_ << "UNKNOWN_BASIC_TYPE_" + current::ToString(p.type_id);  // LCOV_EXCL_LINE
             }
           }
-          void operator()(const ReflectedType_Enum& e) const {
-            oss_ << "C5TCurrent.Enum_IO('" << e.name << "')";
-          }
+          void operator()(const ReflectedType_Enum& e) const { oss_ << "C5TCurrent.Enum_IO('" << e.name << "')"; }
```
sompylasar: For the record, this part was heavily reworked in my PR.

TypeSystem/Reflection/test.cc
```diff
@@ -509,4 +725,48 @@ TEST(Reflection, BaseTypeMatters) {
...
+  }).join();
+  std::thread([one, two]() {
+    EXPECT_EQ(static_cast<uint64_t>(two), static_cast<uint64_t>(current::reflection::CurrentTypeID<Two>()));
+    EXPECT_EQ(static_cast<uint64_t>(one), static_cast<uint64_t>(current::reflection::CurrentTypeID<One>()));
+  }).join();
+}
+
+namespace reflection_test {
+
+CURRENT_STRUCT_T(RSRange) {
```
dkorolev: It's essential to confirm the numerical value or the type ID for this `RandomShitMinMaxRange` remains what it is.

sompylasar: :+1:

dkorolev:
@mzhurovich @sompylasar -- before I get to the finish line, here's another bit of refactoring.

* `CurrentTypeName<T>` is now `CurrentTypeName<T, NameFormat = NameFormat::DebugDump>`.
* The `DebugDump` format is a fully-specified C++ identifier, such as `Optional<std::vector<std::pair<std::string, Variant<Foo, Bar>>>>`. It's not designed to be exportable as part of the schema, but is kept for internal references and, well, debug dumps.
* Another format is `NameFormat::AsIdentifier`. It generates the identifier name that can be used in TypeScript, F#, etc. It still is fully specified, and would look as `Optional_Vector_Pair_String_Variant_Foo_Bar`. Yes, it's not fully reconstructible due to no separators, but I'm still thinking we may be able to get away with this one, not having to use the `_T921...` digits style.
* And the third format is the `NameFormat::Z` one. It's what was there before, with `_Z` at the end of `CURRENT_STRUCT_T`'s names. We may decide to retire it altogether, but that would change many TypeIDs in prod (they start from a `CRC32` of such a symbol), hence I'm keeping it as is for now.

@sompylasar, @mzhurovich, the changes you proposed, to a) unify / get rid of the `_Z` magic, and b) add some `template_inner_type` to our exported JSON schema, are relatively simple from this point on (part of the change is still in this commit, commented out). Let's sync up before I push those through.

Thanks,
Dima

dkorolev:
Another great thing about https://github.com/C5T/Current/pull/780/commits/1cc012b11dc9235b79df9cd041630a0ed94da1ae is that it highlights where is the `_Z` notation used. These places are:
1. JSON variant cases serialization
  * field names for `Current` and `Minimalistic`,
  * the value of `$` for `JavaScript`.
2. Sherlock schema export.

For serialization, I can see different solutions. On the other hand, the value in '$' and the key name of the object should match. On the other hand, for `Minimalistic` to be able to parse `JavaScript` (assuming no collisions), variant case names should match. Using the "C++" name there is one option, as long as we don't have to keep those strings as valid identifiers. To other options are the present `AsIdentifier`, or the "good old" `*_T921...`.

@sompylasar, is that OK for TypeScript schema if the key for the variant case is not an identifier? In other words, is it OK for an object such as `Variant<Fraction, RSRange<double>>` to serialize into:
```
{
  "RSRange<double>": { ... }
}
```
With, possibly, a `"$":"RSRange<double>"` key in case of `JSONFormat::JavaScript`.

Thanks,
Dima

dkorolev:
> Searchability broken on purpose? {Java,Type}Script -> JavaScript, TypeScript

Yes.

dkorolev:
> They look clean in TypeScript type names; for F# I'd add a type aliases for the numeric timestamp types similarly to how I made a standard library of types for TypeScript.
> ...and in "current_type" type names if the "C5TCurrent." is a concern.

@sompylasar , would you be so kind to formulate your proposal that immediately and unambiguously translates into an actionable one in my head? I've re-read this three times and am 100% positive I do not have a clear idea of what change are you looking for, or of what metrics would you use to gauge whether a change I may propose does or does not fit the bill.

Thanks.

sompylasar:
@dkorolev For all types in "AsIdentifier" use capitalized type names, particularly the ones listed as "current_type" in the primitive types DSL file: https://github.com/C5T/Current/blob/cabcd3d3/TypeSystem/primitive_types.dsl.h

I'm looking for consistency in naming convention which is now absent. Capitalizing containers and removing `_t` from some of the type names, but not the other, means to me that the convention chosen here is new: neither from C++, nor from TypeScript. Given the primitives file, it's not from Current either.

dkorolev:
@sompylasar as in 83a6193 ?

dkorolev:
@mzhurovich, @sompylasar  -- I hope the last commit makes the change complete.

@mzhurovich -- I took the liberty to change the schema output format. Upon adding and changing fields (`super_id` is `Optional<>` now), I also changed the JSON format to `Minimalistic`. Hope it's OK from your end.

@sompylasar -- your comments re. the `_Z` -> `_T_...` refactoring are valid, but I suggest we put them aside for now. They only affect C++ schema export; thus, why don't we first confirm it does the job in its present form (one of the purposes for this refactoring was to fix it), and then, as the time comes, make that part of the code clean, knowing we have a live example usage of it to regression-test against.

Gentlemen?

sompylasar:
LGTM. 👍 

1. https://github.com/C5T/Current/pull/780/commits/2f7df8a7cb77dd525fe7d32a7b806f6be437a4a4#r131551749
> Is it inevitable that these pieces of code that create `ReflectedType_XXX` are repeated twice, first in TypeID calculation, second here? Why can't the TypeID calculation return the instances of `ReflectedType_XXX`? This is not obvious, maybe a verbose comment block next to `ReflectorImpl` could help.

2. Could you please make these pieces of code consistent, if not reused from a single place?
```
      CURRENT_ASSERT(input.length() >= 3);
      CURRENT_ASSERT(input.substr(input.length() - 2) == "_Z");
      return input.substr(0, input.length() - 1) + 'T' + current::ToString(static_cast<uint64_t>(t));
```
```
      CURRENT_ASSERT(native_name.length() > 2);
      CURRENT_ASSERT(native_name.substr(native_name.length() - 2) == "_Z");
      return native_name.substr(0, native_name.length() - 2) + "_T" + current::ToString(Value(template_inner_id));
```
```
       CURRENT_ASSERT(s.native_name.length() >= 3u);
       CURRENT_ASSERT(s.native_name.substr(s.native_name.length() - 2u) == "_Z");
       os << "  CURRENT_EXPORTED_TEMPLATED_STRUCT(" << s.native_name.substr(0, s.native_name.length() - 2)
         << type_name(Value(s.template_inner_id), "") << ");\n";
```

dkorolev:
1., re:
> Is it inevitable that these pieces of code that create ReflectedType_XXX are repeated twice, first in TypeID calculation, second here? Why can't the TypeID calculation return the instances of ReflectedType_XXX? This is not obvious, maybe a verbose comment block next to ReflectorImpl could help.
What's the exact piece of code you're talking about?

2. re., the inconsistency -- as said above, a) I'd like the whole `_Z` thing to eventually go away, with that code, and b) if this, admittedly, suboptimal piece of code does the job, refactoring/changing/eliminating it later would only get easier as there are real-life usecases of them.

sompylasar:
@dkorolev 

> 1., re:
> 
> > Is it inevitable that these pieces of code that create `ReflectedType_XXX` are repeated twice, first in TypeID calculation, second here? Why can't the TypeID calculation return the instances of `ReflectedType_XXX`? This is not obvious, maybe a verbose comment block next to `ReflectorImpl` could help.
> 
> What's the exact piece of code you're talking about?

These: https://github.com/C5T/Current/pull/780/files#diff-ea5e1a6efd2103878cbaf95ea4e0b8adR79
```
// Populated by `InternalCurrentTypeID`, used as a `ThreadLocalSingleton`.
struct RecursiveTypeTraverser {
  // ...

  template <typename TK, typename TV>
  TypeID operator()(TypeSelector<std::map<TK, TV>>) {
    return CalculateTypeID(ReflectedType_Map(CurrentTypeID_<TK>(), CurrentTypeID_<TV>()));
  }

  template <typename TK, typename TV>
  TypeID operator()(TypeSelector<std::unordered_map<TK, TV>>) {
    return CalculateTypeID(ReflectedType_UnorderedMap(CurrentTypeID_<TK>(), CurrentTypeID_<TV>()));
  }

  // ...

  template <typename T>
  std::enable_if_t<IS_CURRENT_STRUCT(T), TypeID> operator()(TypeSelector<T>) {
    ReflectedType_Struct result;
    result.native_name = CurrentTypeName<T, NameFormat::Z>();

    result.super_id = ReflectSuper<T>();
    if (Exists(result.super_id)) {
      result.super_name = CurrentTypeName<typename T::super_t, NameFormat::Z>();
    }
  
    result.template_inner_id = ReflectTemplateInnerType<T>();
    if (Exists(result.template_inner_id)) {
      result.template_inner_name = CurrentTypeName<TemplateInnerType<T>, NameFormat::Z>();
    }
  
    VisitAllFields<T, FieldTypeAndNameAndIndex>::WithoutObject(
        StructFieldsTraverser<T>(top_level_type_, top_level_type_name_, result.fields));

    return CalculateTypeID(result);
  }
```
and https://github.com/C5T/Current/pull/780/files#diff-ea5e1a6efd2103878cbaf95ea4e0b8adR310
```
// `ReflectorImpl` is a thread-local singleton to generate reflected types metadata at runtime.
struct ReflectorImpl {
  // ...

  template <typename TK, typename TV>
  ReflectedType operator()(TypeSelector<std::map<TK, TV>>) {
    ReflectType<TK>();
    ReflectType<TV>();
    ReflectedType_Map result(CurrentTypeID<TK>(), CurrentTypeID<TV>());
    result.type_id = CurrentTypeID<std::map<TK, TV>>();
    return ReflectedType(std::move(result));
  }

  template <typename TK, typename TV>
  ReflectedType operator()(TypeSelector<std::unordered_map<TK, TV>>) {
    ReflectType<TK>();
    ReflectType<TV>();
    ReflectedType_UnorderedMap result(CurrentTypeID<TK>(), CurrentTypeID<TV>());
    result.type_id = CurrentTypeID<std::unordered_map<TK, TV>>();
    return ReflectedType(std::move(result));
  }

  // ...

  template <typename T>
  std::enable_if_t<IS_CURRENT_STRUCT(T), ReflectedType> operator()(TypeSelector<T>) {
    ReflectedType_Struct s;
    s.native_name = CurrentTypeName<T, NameFormat::Z>();
  
    s.super_id = ReflectSuper<T>();
    if (Exists(s.super_id)) {
      s.super_name = CurrentTypeName<typename T::super_t, NameFormat::Z>();
    }

    s.template_inner_id = ReflectTemplateInnerType<T>();
    if (Exists(s.template_inner_id)) {
      s.template_inner_name = CurrentTypeName<TemplateInnerType<T>, NameFormat::Z>();
    }

    VisitAllFields<T, FieldTypeAndNameAndIndex>::WithoutObject(InnerStructFieldsTraverser<T>(s.fields));
    s.type_id = CurrentTypeID<T>();

    return s;
  }
```

dkorolev:
@sompylasar, I don't understand. What exactly is your concern, and how would you like the change (or at least its outcome) look like?

dkorolev:
Ah, you just mean code duplication in those two pieces? Yes, let's keep it the way it is until the next round of refactoring. Need to at least confirm RAM usage and compilation times are as they should be.

TL;DR: These are two independent stages, one internal one "user-facing". They follow different routes altogether, and do different things, with the 2nd stage depending on the 1st one being completed.

sompylasar:
@dkorolev Yes, code duplication. That's why I said about a verbose comment next to the class/struct where these are defined: 

> Is it inevitable that these pieces of code [...] are repeated twice? [...] This is not obvious, maybe a verbose comment block next to `ReflectorImpl` could help.

Basically this might work:

> These are two separate stages, one internal and one "user-facing". They follow different routes altogether, and do different things, with the 2nd stage depending on the 1st one being completed.

dkorolev:
> Basically this might work.

That's how it is now.

sompylasar:
This might work as a verbose comment next to the class. Now it is one line here:
```
// Populated by `InternalCurrentTypeID`, used as a `ThreadLocalSingleton`.
struct RecursiveTypeTraverser {
```
and one line here:
```
// `ReflectorImpl` is a thread-local singleton to generate reflected types metadata at runtime.
struct ReflectorImpl {
```
and no details about stages and interdependencies that you outlined.

dkorolev:
@sompylasar 1ae5265

TypeSystem/Reflection/test.cc
```diff
@@ -118,6 +117,101 @@ static_assert(std::is_same<current::reflection::FieldTypeWrapper<Foo>,
 
 }  // namespace reflection_test
 
+TEST(Reflection, CurrentTypeName) {
+  using current::reflection::CurrentTypeName;
+  using current::reflection::NameFormat;
+  using namespace reflection_test;
+
+  // NameFormat::Z: C++ types, template names as `_Z`, variant names expanded as `_B_..._E`.
```
mzhurovich: `template names` -> `template class names`

TypeSystem/Reflection/test.cc
```diff
@@ -118,6 +117,101 @@ static_assert(std::is_same<current::reflection::FieldTypeWrapper<Foo>,
...
+  // Unused for the very naming, kept for legacy reasons of inner TypeID computation when the dependency is cyclic (?).
+  // TODO(dkorolev) + TODO(mzhurovich): Why don't we retire this? It would change some TypeIDs, so what?
```
mzhurovich: At least we have to think about remote stream schema ID check that will start failing.

TypeSystem/Reflection/test.cc
```diff
@@ -141,26 +235,100 @@ TEST(Reflection, StructAndVariant) {
...
-  EXPECT_EQ(9345111461746810545ull, static_cast<uint64_t>(self_c.fields[1].type_id));
+  EXPECT_EQ(typeid_c, static_cast<uint64_t>(self_c.type_id));
+  EXPECT_EQ(typeid_vb, static_cast<uint64_t>(self_c.fields[0].type_id));
+  EXPECT_EQ(typeid_msc, static_cast<uint64_t>(self_c.fields[1].type_id));
+}
 
-  EXPECT_NE(
-      static_cast<uint64_t>(Value<ReflectedType_Variant>(Reflector().ReflectType<FooBarBaz>()).type_id),
-      static_cast<uint64_t>(Value<ReflectedType_Variant>(Reflector().ReflectType<Variant<Foo, Bar, Baz>>()).type_id));
+#ifdef TODO_DKOROLEV_EXTRA_PARANOID_DEBUG_SYMBOL_NAME
```
mzhurovich: Do we plan too keep `TODO_DKOROLEV_` prefix?

TypeSystem/Reflection/test.cc
dkorolev: +1

TypeSystem/Reflection/test.cc
dkorolev: There's some `*_PARANOID_*` symbol in `alt-master` already IIRC, hence I suggest we merge first, and then rename this.

dkorolev:
@mzhurovich -- two minor code changes suggested by you: done.

TypeSystem/Schema/golden/smoke_test_struct.cc
mzhurovich: @dkorolev There's no declaration of `TemplatedInheriting_Z` in this file. How it supposed to work? :-/

dkorolev: Pure C++ schema exporter is unused now. It's missing another important piece: a base class with a virtual destructor for what were `CURRENT_STRUCT`-s.

My understanding was that we revisit this once we get into the high-performance `dlopen` territory.

TypeSystem/Schema/schema.h
dkorolev: :+1: 1be00b0

TypeSystem/typename.h
```diff
@@ -30,80 +30,307 @@ SOFTWARE.
...
+
 namespace reflection {
 
-template <typename T, bool TRUE_IF_CURRENT_STRUCT, bool TRUE_IF_VARIANT>
+struct CurrentVariantDefaultName;
+
+// `DebugDump`: ex. "Templated<std::vector<std::string>>", "std::vector<Templated<uint64_t>>". C++-level output.
+// `Z`: ex. "std::vector<std::string>", "Templated_Z". *ONLY* for legacy reasons of TypeID computation, TO BE FIXED.
+// `AsIdentifier`: ex. "Vector_Templated_T_Foo", for serialization and for schema dumps into {Java,Type}Script et. al.
+enum class NameFormat : int { DebugDump = 0, Z, AsIdentifier };
```
mzhurovich: @dkorolev s/DebugDump/FullCPP/?

dkorolev: :+1: 0ee8309

PR #784, opened 2017-08-29
Storage containers expose their field names.

mzhurovich:


sompylasar:
@mzhurovich Could you please add an example where this can be useful? I can think of what this is needed for, but it's not obvious now.

dkorolev:
@sompylasar, it's the least painful way to enable _that_ feature in _that_ proprietary code by _that_ guy.

PR #785, opened 2017-10-06
A few Mac-friendly tweaks while I have a chance.

dkorolev:


.gitignore
```diff
@@ -55,3 +55,4 @@ CMakeLists.txt
 # Let's not even talk about it.
 node_modules
 **/node_modules
+**/package-lock.json
```
sompylasar: I see you've upgraded npm >_<

sompylasar: By the way, the Node package should have its own gitignore, for modularity. Global gitignore is a "god object".

dkorolev: Why not .gitignore all those lockfiles regardless?

sompylasar: 1. Not all files with a lock name are lock files in the sense you're thinking.
2. You assume that you "own" all the subdirectories, and the rules equally apply to all of them (which may or may not be true).

dkorolev: Explain in one sentence what specific change would you like to see, or greenlight this PR please.

sompylasar: If you didn't notice, I have greenlighted.

PR #786, opened 2017-10-11
I f*cking had to re-create the `typesystem` dir to get rid of those case-insensitive warnings on Mac.

dkorolev:


dkorolev:
Interestingly, to fully fix the issue I had to re-clone the repo. Just fast-forwarding didn't do the job.

@sompylasar, here's me openly stating git does suck. Only on a Mac though. Because Mac's file system sucks. Oh well.

sompylasar:
I'd be surprised if this didn't kill the historical change tracking for these files (git blame). A two-commit rename (move) might have been a better idea.

dkorolev:
I tried that. It didn't work.

sompylasar:
I believe it did work, but you haven't re-cloned to see that.

PR #787, opened 2017-10-24
Fix `HTTP(GET("http://something/without/port"))` to use port 80, not 0.

sompylasar:
Regression in:
- https://github.com/C5T/Current/pull/754
- https://github.com/C5T/Current/commit/3d701107d6fd89c7da63048738da59d7280260ba

Cause:
`URL` constructor behavior was changed to not set port to 80 if the original URL string doesn't specify port, but `HTTP` relied on the old behavior when attempting to connect.
This reproduces only if the port in the URL is omitted, and wasn't tested because it's hard to bind to port 80 to accept connections (sudo is required to run the test).

Tests (sudo required to bind to port 80):
```
( cd Blocks/HTTP/ && make clean && make test )
( cd Blocks/HTTP/.current && sudo ./test --net_api_test_port 80 --current_runtime_arch Darwin )
( cd Blocks/HTTP/.current && sudo ./test --net_api_test_port_secondary 80 --current_runtime_arch Darwin )
```

Thanks @mzhurovich for finding this.

### DIFF
--- a/Blocks/HTTP/test.cc
+++ b/Blocks/HTTP/test.cc
@@ -368,7 +368,7 @@ TEST(HTTPAPI, RedirectToRelativeURL) {
 }
 
 TEST(HTTPAPI, RedirectToFullURL) {
-  ASSERT_FALSE(FLAGS_net_api_test_port_secondary == FLAGS_net_api_test_port);
+  ASSERT_NE(FLAGS_net_api_test_port_secondary, FLAGS_net_api_test_port);
   // Need a live port for the redirect target because the HTTP client is following the redirect
   // and tries to connect to the redirect target, otherwise throws a `SocketConnectException`.
   const auto scope_redirect_to = HTTP(FLAGS_net_api_test_port_secondary).Register("/to", [](Request r) { r("Done."); });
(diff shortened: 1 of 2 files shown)


Blocks/HTTP/test.cc
```diff
@@ -360,10 +360,15 @@ TEST(HTTPAPI, RedirectToRelativeURL) {
...
-  EXPECT_EQ(Printf("http://localhost:%d/to", FLAGS_net_api_test_port), response.url);
+  EXPECT_EQ((
+    FLAGS_net_api_test_port == 80
+      ? "http://localhost/to"
+      : Printf("http://localhost:%d/to", FLAGS_net_api_test_port)
+  ), response.url);
 }
 
 TEST(HTTPAPI, RedirectToFullURL) {
+  ASSERT_FALSE(FLAGS_net_api_test_port_secondary == FLAGS_net_api_test_port);
```
dkorolev: `ASSERT_NE`?

sompylasar: 👍 https://github.com/C5T/Current/pull/787/commits/00f1acddf7431d12c8b3bb66087d68a911a2f7a1

PR #791, opened 2017-10-30
Hotfix for autogenerated evolution of Optional<TemplatedStruct>.

mzhurovich:
@dkorolev This is a hotfix. I'm 99% sure that we have to treat templated types more generically, to support evolution for all containers.


dkorolev:
LGTM

PR #793, opened 2017-11-01
Compose functions for URLPathArgs.

mzhurovich:


Blocks/URL/url.h
```diff
@@ -515,6 +515,23 @@ struct URLPathArgs {
 
   void add(const std::string& arg) { args_.push_back(arg); }
 
+  std::string ComposeURLPathFromArgs() const {
+    std::string result;
+    for (auto cit = args_.crbegin(); cit != args_.crend(); ++cit) {
+      result += "/" + *cit;
```
dkorolev: `EncodeURLComponent`?

Blocks/URL/url.h
```diff
@@ -515,6 +515,23 @@ struct URLPathArgs {
 
   void add(const std::string& arg) { args_.push_back(arg); }
 
+  std::string ComposeURLPathFromArgs() const {
+    std::string result;
+    for (auto cit = args_.crbegin(); cit != args_.crend(); ++cit) {
```
dkorolev: Won't `for (const auto& component : (*this)) {` do the trick here? It's already a reverse iterator.

Blocks/HTTP/test.cc
```diff
@@ -171,6 +171,55 @@ TEST(HTTPAPI, RegisterWithURLPathParams) {
...
+  EXPECT_EQ("/user/a (/0, /user/a/0)", run("/user/a/0/"));
+
+  EXPECT_EQ("/user/a/1 (/, /user/a/1)", run("/user/a/1"));
+  EXPECT_EQ("/user/a/1 (/, /user/a/1)", run("/user/a/1/"));
+
+  EXPECT_EQ("/user/a (/2, /user/a/2)", run("/user/a/2"));
+  EXPECT_EQ("/user/a (/2, /user/a/2)", run("/user/a/2/"));
+
+  EXPECT_EQ("/ (/user/a/1/blah, /user/a/1/blah)", run("/user/a/1/blah"));
+  EXPECT_EQ("/ (/user/a/1/blah, /user/a/1/blah)", run("/user/a/1/blah/"));
```
dkorolev: A nit from @sompylasar who happened to be looking over my shoulder: in the proper world of URLs the trailing slash, if provided originally, carries an important meaning ("is significant" as he phrased it), and is best to be kept; both in `ComposeURLPathFromArgs()` and in `ComposeURLPath()`.

I won't blame you if you decide to not give a damn as of now :-)

Blocks/URL/url.h
mzhurovich: But `URLWithoutParametersParser::ComposeURL()` doesn't do it. Consistency FTW?

Blocks/URL/url.h
mzhurovich: Yes, it will

Blocks/HTTP/test.cc
mzhurovich: `URLPathArgs` does not contain this information. It's stored separately in `Request`.

Blocks/URL/url.h
dkorolev: One could argue `EncodeURLComponent` must be there for both. What if there's a slash in one of the URL path args?

Blocks/HTTP/test.cc
dkorolev: Ah, right. I suggest we ignore it for now then.

dkorolev:
LGTM!

Issue #796, opened 2017-12-05
Blocks/HTTP/request: Segmentation fault after responding twice with Response


sompylasar:
```cpp
TEST(HTTPAPI, HandlesRespondTwiceWithString) {
  const auto scope = HTTP(FLAGS_net_api_test_port)
                         .Register("/respond_twice",
                                   [](Request r) {
                                     r("OK");
                                     r("FAIL");
                                   });
  const string url = Printf("http://localhost:%d/respond_twice", FLAGS_net_api_test_port);
  const auto response = HTTP(GET(url));
  EXPECT_EQ(200, static_cast<int>(response.code));
  EXPECT_EQ("OK", response.body);
  EXPECT_EQ(url, response.url);
}

TEST(HTTPAPI, HandlesRespondTwiceWithResponse) {
  const auto scope = HTTP(FLAGS_net_api_test_port)
                         .Register("/respond_twice",
                                   [](Request r) {
                                     r(Response("OK", HTTPResponseCode.OK));
                                     // FIXME: On this line, `Request r` is invalid after `std::move(*this)` in `Request#operator()`.
                                     r(Response("FAIL", HTTPResponseCode(762)));  // https://github.com/joho/7XX-rfc
                                   });
  const string url = Printf("http://localhost:%d/respond_twice", FLAGS_net_api_test_port);
  const auto response = HTTP(GET(url));
  EXPECT_EQ(200, static_cast<int>(response.code));
  EXPECT_EQ("OK", response.body);
  EXPECT_EQ(url, response.url);
}
```

```
[ RUN      ] HTTPAPI.HandlesRespondTwiceWithString
HTTP route failed in user code:
./../../Bricks/net/http/impl/server.h:572
AttemptedToSendHTTPResponseMoreThanOnce()
[       OK ] HTTPAPI.HandlesRespondTwiceWithString (1 ms)
[ RUN      ] HTTPAPI.HandlesRespondTwiceWithResponse
make: *** [test] Segmentation fault: 11
```

Demo: https://github.com/sompylasar/Current/commit/38a13271b487f68f9feb5f6354476ada77fa409a

dkorolev:
Ugh.

dkorolev:
Fixed.

PR #797, opened 2017-12-05
Blocks/HTTP/request: Segmentation fault after responding twice with Response

sompylasar:
### DO NOT MERGE, NOW FAILING TESTS WITH SEGFAULT

```
[ RUN      ] HTTPAPI.HandlesRespondTwiceWithString
HTTP route failed in user code:
./../../Bricks/net/http/impl/server.h:572
AttemptedToSendHTTPResponseMoreThanOnce()
[       OK ] HTTPAPI.HandlesRespondTwiceWithString (1 ms)
[ RUN      ] HTTPAPI.HandlesRespondTwiceWithResponse
make: *** [test] Segmentation fault: 11
```

Refs https://github.com/C5T/Current/issues/796

### DIFF
--- a/Blocks/HTTP/request.h
+++ b/Blocks/HTTP/request.h
@@ -103,12 +103,18 @@ struct Request final {
   // Support objects with user-defined HTTP response handlers.
   template <typename T>
   inline typename std::enable_if<HasRespondViaHTTP<current::decay<T>>(0)>::type operator()(T&& that_dude_over_there) {
+    if (!unique_connection) {
+      CURRENT_THROW(net::AttemptedToSendHTTPResponseMoreThanOnce());
+    }
     that_dude_over_there.RespondViaHTTP(std::move(*this));
   }
 
   // A shortcut to allow `[](Request r) { r("OK"); }` instead of `r.connection.SendHTTPResponse("OK")`.
   template <typename... TS>
   void operator()(TS&&... params) {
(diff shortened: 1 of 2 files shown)


dkorolev:
The `std::move` doesn't physically invalidate the object being moved away; it'll still be destructed later on as it leaves the scope. What `std::move` does is allows the user to treat the object being moved as the temporary one, which can be invalidated right away (such as, a `string` could just claim the buffer previously allocated in the string being moved, as opposed to allocating a new one and copying the data over).

Move of `Request` is handled by the move constructor and move assignment operator (`Request(Request&& what_is_being_moved)` and `Request& operator=(Request&& what_is_being_moved)`.

I gotta run for a birthday dinner, but the proper fix should be the one that marks `what_is_being_moved` as "you are moved away", and checks for "was I moved away" as the response is attempted to be sent (twice).

Perhaps just setting the "response was already sent" flag to `true` in `what_is_being_moved` would be good enough.

sompylasar:
Right. This is what is already being done, with a throw if already responded, in the `connection.SendHTTPResponse` that is called from the regular `operator()` but indeed has to be done in this special `operator()` as well.

dkorolev:
@sompylasar -- fixed and tested the fix, see the last two commits.
@mzhurovich -- would you pls merge this one?

PR #798, opened 2017-12-09
Fast unchecked subscription and unsafe replication

grixa:
1. Unsafe stream subscription (from #779 ) and its tests. Should we call it "unchecked" instead?
2. Unsafe replication on client side - write to the file directly without parsing the entries.
3. Straightforward 1Mb (should we make it an option?) cache in the `ChunckedResponseSender`.
4. Fake mode for the replication_client in the benchmark.
5. Fake stream mode for the replication_server in the benchmark.

PR #801, opened 2018-01-03
Fast unchecked subscription and unsafe replication

grixa:
A reincarnation of the #798


blocks/persistence/exceptions.h
```diff
@@ -64,6 +64,12 @@ struct PersistenceFileNotWritable : PersistenceException {
       : PersistenceException("Persistence file not writable: `" + filename + "`.") {}
 };
 
+struct UnsafePublishBadIndexTimestampException : PersistenceException {
+  explicit UnsafePublishBadIndexTimestampException(uint64_t expected, uint64_t found)
+      : PersistenceException(current::strings::Printf(
+          "Expecting index %lld, seeing %lld.", static_cast<long long>(expected), static_cast<long long>(found))) {}
```
mzhurovich: BTW, cross-platform macro for printing `uint64_t` is `PRIu64`

stream/stream_impl.h
```diff
@@ -142,6 +142,20 @@ class StreamPublisherImpl {
     return result;
   }
 
+  template <current::locks::MutexLockStatus MLS>
+  idxts_t PublisherPublishUncheckedImpl(std::string&& entry_json) {
+	  const auto result = data_->persister.template PersisterPublishUncheckedImpl<MLS>(std::move(entry_json));
```
mzhurovich: `PersisterPublishUncheckedImpl` has only one version accepting `const std::string&`. Why do you need `std::string&&` versions of Publish implementations?

blocks/persistence/file.h
```diff
@@ -468,6 +468,37 @@ class FilePersister {
     return idxts;
   }
 
+  template <current::locks::MutexLockStatus MLS>
+  idxts_t PersisterPublishUncheckedImpl(const std::string& entry_json) {
```
mzhurovich: Strictly speaking, parameter is not entry JSON, it's raw log line.

mzhurovich: `Entry` may be confused with the stream entry itself that is stored after `\t`.

bricks/net/http/impl/server.h
```diff
@@ -597,14 +605,32 @@ class GenericHTTPServerConnection final : public HTTPResponder {
...
-            // Force every chunk to be sent out by passing `false` as the second argument.
-            connection_.BlockingWrite(constants::kCRLF, false);
+            if (!data.empty()) {
+              const auto chunk_header = strings::Printf("%lX", data.size()) + constants::kCRLF;
+              const auto chunk_size = chunk_header.size() + data.size() + constants::kCRLFLength;
+              if (cache_size_ && (flush == ChunkFlush::Flush || chunk_size > CACHE_SIZE - cache_size_)) {
+                connection_.BlockingWrite(data_cache_, cache_size_, true);
+                cache_size_ = 0;
+              }
+              if (flush == ChunkFlush::Flush || chunk_size > cache_size_) {
```
mzhurovich: Shouldn't this be `... || chunk_size > CACHE_SIZE - cache_size_` as above?

blocks/persistence/exceptions.h
dkorolev: We have none of those in Current btw, and I kind of like the pure way with `%lld` and a `static_cast<long long>`. Btw, `3rdparty/rapidjson/msinttypes/inttypes.h` [re]defines this macro for some old Visual Studio builds.

bricks/net/http/impl/server.h
```diff
@@ -73,6 +73,8 @@ SOFTWARE.
 
 #endif  // CURRENT_BRICKS_DEBUG_HTTP
 
+#define CURRENT_BRICKS_HTTP_DEFAULT_CHUNK_CACHE_SIZE  1024*1024
```
dkorolev: Parenthesis? :-)

stream/replicator.h
```diff
@@ -162,6 +167,42 @@ class SubscribableRemoteStream final {
       }
     }
 
+    template <SubscriptionMode MODE = SM>
+    ENABLE_IF<MODE == SubscriptionMode::Checked> PassEntryToSubscriber(const std::string& entry_json) {
+      const auto split = current::strings::Split(entry_json, '\t');
+      const auto tsoptidx = ParseJSON<ts_optidx_t>(split[0]);
```
dkorolev: Missing a check that `split` has at least one element.

bricks/net/http/impl/server.h
grixa: Good point, it's actually a bad mistake, which appeared in the aca9f24. Sure, the condition should look like `chunk_size > CACHE_SIZE`. There is no need to substract the `cache_size_`, because if the `chunk_size` was bigger than `CACHE_SIZE -
 cache_size`, but smaller than the `CACHE_SIZE`, it would be checked above and as a result the `cache_size_` became zero.

PR #805, opened 2018-01-21
Several minor tweaks.

dkorolev:
Hi @mzhurovich, @grixa,

A couple minor tweaks on the way to:

1. Refactoring class `Response` to support JSON format selection (otherwise the user has to manually set the origin header for each `JSON<JSONFormat::Minimalistic>`, which sucks, and
2. The journaled file system and its applications.

On the bright side, `RespondViaHTTP` is now not SFINAE-powered.

Thanks,
Dima

### DIFF
--- a/bricks/net/http/headers/headers.h
+++ b/bricks/net/http/headers/headers.h
@@ -224,6 +224,12 @@ struct Headers final {
     return *this;
   }
 
+  Headers& Clear(const std::string& header) {
+    Header::ThrowIfHeaderIsCookie(header);
+    map.erase(header);
+    return *this;
+  }
+
   bool Has(const std::string& header) const {
     Header::ThrowIfHeaderIsCookie(header);
     return map.find(header) != map.end();
--- a/bricks/net/http/headers/headers.h
+++ b/bricks/net/http/headers/headers.h
@@ -224,7 +224,7 @@ struct Headers final {
     return *this;
   }
 
-  Headers& Clear(const std::string& header) {
+  Headers& Remove(const std::string& header) {
     Header::ThrowIfHeaderIsCookie(header);
     map.erase(header);
     return *this;
--- a/storage/container/dictionary.h
+++ b/storage/container/dictionary.h
@@ -52,6 +52,7 @@ class GenericDictionary {
 
   bool Empty() const { return map_.empty(); }
   size_t Size() const { return map_.size(); }
+  bool Has(sfinae::CF<key_t> x) const { return map_.find(x) != map_.end(); }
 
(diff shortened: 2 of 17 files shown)


bricks/net/http/headers/headers.h
```diff
@@ -224,6 +224,12 @@ struct Headers final {
     return *this;
   }
 
+  Headers& Clear(const std::string& header) {
```
sompylasar: Clear -> Remove? Clear usually stands for RemoveAll and goes without arguments

dkorolev: 3d8b2cc9

typesystem/variant.h
```diff
@@ -212,6 +212,15 @@ struct VariantImpl<NAME, TypeListImpl<TYPES...>> : IHasUncheckedMoveFromUniquePt
...
 
+#ifdef VARIANT_CHECKS_AT_RUNTIME_INSTEAD_OF_COMPILE_TIME
+  template <typename T, typename... ARGS>
+#else
+  template <typename T, typename... ARGS, class ENABLE = std::enable_if_t<TypeListContains<typelist_t, T>::value>>
+#endif  // VARIANT_CHECKS_AT_RUNTIME_INSTEAD_OF_COMPILE_TIME
+  T& Construct(ARGS&&... args) {
+    object_ = std::make_unique<T>(std::forward<ARGS>(args)...);
+    return *dynamic_cast<T*>(object_.get());
+  }
```
mzhurovich: @dkorolev Why do you need this ^^^?

storage/container/dictionary.h
```diff
@@ -52,6 +52,7 @@ class GenericDictionary {
 
   bool Empty() const { return map_.empty(); }
   size_t Size() const { return map_.size(); }
+  bool Has(sfinae::CF<key_t> x) const { return map_.find(x) != map_.end(); }
```
mzhurovich: Add the same method for matrix containers as well?

typesystem/variant.h
dkorolev: For one-liners, such as `r(Response(Variant<Foo, Bar>().template Construct<Foo>(42)));`. Proved to be a useful construct to return a JSON object what wraps as an variant without having to declare extra variables and/or to use excessive `Value<Foo>(...)`.

storage/container/dictionary.h
dkorolev: :+1: ed9a57fe59666802824c3e4fbb05b6831e28127e, 93803336f6c2ae57af28adf51b019cce25213a4b

dkorolev:
Merging per an offline chat w/ @mzhurovich.

PR #809, opened 2018-02-11
A yet another relatively useless C++ HTML-generating library. 

dkorolev:
Hi @mzhurovich,

I had to do it again ... and decided to make the code generic, and have it belong here.

CCs, and it's probably OK to only look at the test:

* @grixa because metaprogramming.
* @sompylasar because it's f*cking HTML, that, yes, no one generates manually these days.
* @idej because you're the reviewer of the proprietary part.
* @rep-movsd because you did implement something similar and would probably like to see this.

### DIFF
--- a/blocks/html/html.h
+++ b/blocks/html/html.h
@@ -32,6 +32,7 @@ SOFTWARE.
 
 #include <sstream>
 #include <string>
+#include <vector>
 
 #include "../../bricks/exception.h"
 #include "../../bricks/util/singleton.h"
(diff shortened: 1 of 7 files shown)


blocks/html/html.h
```diff
@@ -0,0 +1,175 @@
...
+#include <string>
+#include <vector>
+
+#include "../../bricks/exception.h"
+#include "../../bricks/util/singleton.h"
+
+namespace current {
+namespace html {
+
+// An exception that is thrown if multiple interesting HTML generation scopes per thread are attempted to be used.
```
sompylasar: interesting -> intersecting

dkorolev: Good 👁 , 0baf6d44 👍

dkorolev:
@sompylasar, it's really not about closing the tags correctly.

It is about generating HTML from C++ in a way that is at least somewhat maintainable, because `ostringstream << "</close_old_tag><open_new_tag>"` is just too bad.

How could the Virtual DOM idea be implemented in C++? Note that in the design I am proposing properties are strongly typed; e.g., the user can only insert the tags that were defined as legal, and the user can only set the properties that were defined as legal for those tags.

dkorolev:
(I would also go as far as saying that capitalized `HTML` being searchable, along with the shortest possible syntax I could think of, is among the main design concepts behind this implementation.)

sompylasar:
I wouldn't try making strongly typed attributes and limiting them to a predefined set (React stopped doing that a version ago). There are, for instance, data attributes (that starts with `data-`, also note the dash) that are all valid. Also `aria-` for accessibility, there are about 30 of them.

Instead, make one "tag" type and make sure you properly escape and quote attribute values and text content when you render the markup. You can choose to either only allow strings as attributes, or go the extra mile to allow numbers and, for special occasions such as `style` (CSS), objects.

Then you might want to implement components (functions or classes with a render function that return a VDOM subtree). These components, defined by the user, can have strongly typed props, when they render, they would pass them down to either other components with strongly typed props or to lower-level HTML with loosely typed attributes.

To define a VDOM component, you would use CURRENT_VDOM_COMPONENT analogous to CURRENT_STRUCT that will add boilerplate necessary to support lists of attributes and VDOM tree children.

If you only render to string, you don't need the complexity of VDOM reconciliation algorithms (partial application of newly rendered subtrees), you just ask the root VDOM component to render, it'll return you the VDOM tree.

For implementation of the VDOM Node type (can be either component or tag), have a look at:

* https://github.com/facebook/react/blob/master/packages/react/src/ReactElement.js
* https://github.com/developit/preact/blob/master/src/h.js

For implementation of VDOM tree to string rendering, have a look at:

* https://github.com/facebook/react/blob/master/packages/react-dom/src/server/ReactPartialRenderer.js
* https://github.com/developit/preact-render-to-string/blob/master/src/index.js

dkorolev:
>  If you get a chance, step away and look what you're doing in a bigger picture, not through the tight scope of diff and vim / command line.

The bigger picture is that I have a tool too build, which offers some interactive interface. I'd like to make ~20 backend + frontend iterations per day, and with anything even as simple as jQuery I'd be limited to maybe three.

Also, the bigger picture is that reliable results matter more than engineering "beauty" in a moment.

The latter has effectively ensured frontend developers would be employed forever, even though not much has changed in the UI space in the past 20+ years.

The former ... well, Unix way rocks, and it has evolved organically from C to bash and pipes to grep/cut/awk, to JSON and jq, and now to super-lightweight and remarkably useful HTML-based interactive browsers.

If my Facebook (Facebook owns React, right) would actually deliver on the promise re. lightning fast interfaces (<0.1s to fold/unfold sections such as comments), I'll consider buying into your philosophy. For now, if you don't mind, I'll focus on getting some tangible things done.

sompylasar:
I'll just leave this here for your consideration: https://github.com/mbasso/asm-dom/tree/5f941c8055c63fcbebc4923c316798a2213a28aa/cpp

sompylasar:
https://github.com/mbasso/gccx 

Preview

sompylasar:
> The bigger picture is that I have a tool too build, which offers some interactive interface. I'd like to make ~20 backend + frontend iterations per day, and with anything even as simple as jQuery I'd be limited to maybe three.

In other news, try learning [create-react-app](https://github.com/facebook/create-react-app) – it'll give you far more iterations per hour than your homegrown HTML macros framework.

PR #811, opened 2018-03-26
Making `std::tuple<>`-s [almost] first-class citizens of Current's type system.

dkorolev:
CC @nyospe @grixa @sompylasar.

I didn't bother with `TypeID`-s (yet), mostly worked on JSON support and a few helper methods Max asked for.

Thanks,
Dima

### DIFF
--- a/bricks/template/tuple.h
+++ b/bricks/template/tuple.h
@@ -47,6 +47,44 @@ static_assert(is_std_tuple<std::tuple<int, std::string>>::value, "");
 static_assert(!is_std_tuple<int>::value, "");
 static_assert(!is_std_tuple<std::string>::value, "");
 
+template <typename T>
+struct wrapped_into_tuple_impl {
+  using type = std::tuple<T>;
+  static type Cast(const T& parameter) { return type(parameter); }
+};
+
+template <typename... TS>
+struct wrapped_into_tuple_impl<std::tuple<TS...>> {
+  using type = std::tuple<TS...>;
+  static const type& Cast(const type& parameter) { return parameter; }
+};
+
+template <typename T>
+using wrapped_into_tuple_t = typename wrapped_into_tuple_impl<T>::type;
+
+static_assert(std::is_same<std::tuple<>, wrapped_into_tuple_t<std::tuple<>>>::value, "");
+static_assert(std::is_same<std::tuple<int>, wrapped_into_tuple_t<int>>::value, "");
+static_assert(std::is_same<std::tuple<int>, wrapped_into_tuple_t<std::tuple<int>>>::value, "");
+static_assert(std::is_same<std::tuple<char, double>, wrapped_into_tuple_t<std::tuple<char, double>>>::value, "");
+
+template <class LHS, class RHS>
+struct tuple_cat_type_impl;
+
+template <typename... LHS_TS, typename... RHS_TS>
+struct tuple_cat_type_impl<std::tuple<LHS_TS...>, std::tuple<RHS_TS...>> {
+  using type = std::tuple<LHS_TS..., RHS_TS...>;
+};
+
+template <typename LHS, typename RHS>
+using tuple_cat_t = typename tuple_cat_type_impl<wrapped_into_tuple_t<LHS>, wrapped_into_tuple_t<RHS>>::type;
+
+static_assert(std::is_same<std::tuple<char, double>, tuple_cat_t<char, double>>::value, "");
+static_assert(std::is_same<std::tuple<char, double>, tuple_cat_t<std::tuple<char>, double>>::value, "");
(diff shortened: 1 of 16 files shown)


typesystem/serialization/test.cc
```diff
@@ -241,6 +247,7 @@ TEST(Serialization, Binary) {
 #endif
 
 TEST(JSONSerialization, CPPTypes) {
+  using namespace serialization_test;
```
sompylasar: This `using` looks unnecessary. Even if there was anything from that namespace, it would look more aligned with explicit references like in the test below: `serialization_test::Serializable`, `serialization_test::Enum`.

sompylasar:
If you want to eventually add TypeScript support, it'll need to use `iots.tuple` with arrays of types like here:
https://github.com/gcanti/io-ts/blob/0.6.x/src/index.ts#L577

i.e. instead of using [comma-separated generic type arguments](https://github.com/C5T/Current/blob/962fc820368ddeb03bf10737a37f3b4c0296feaa/typesystem/schema/c5t-current-schema-ts/src/index.ts#L67-L70) like `Pair<T_FIRST, T_SECOND>` does when the number of items in the tuple is known:
```ts
export const Pair_IO = <RT_FIRST extends iots.Any, RT_SECOND extends iots.Any>(FirstType_IO: RT_FIRST, SecondType_IO: RT_SECOND, name?: string) => (
  iots.tuple([ FirstType_IO, SecondType_IO ], name || ('C5TCurrent.Pair<' + FirstType_IO.name + ', ' + SecondType_IO.name + '>'))
);
export type Pair<T_FIRST, T_SECOND> = [ T_FIRST, T_SECOND ];
```

for `Tuple` we'll have to use an array of types `Tuple<[ T_FIRST, T_SECOND ]>`:
```ts
export const Tuple_IO = <RT_TUPLE_TYPES extends Array<iots.Any>>(TupleTypes_IO: RT_TUPLE_TYPES, name?: string) => (
  iots.tuple(TupleTypes_IO, name || ('C5TCurrent.Tuple<[' + TupleTypes_IO.map((rt) => (rt.name)).join(', ') + ']>'))
);
export type Tuple<T_TUPLE_TYPES extends Array<typeof iots.Any>> = T_TUPLE_TYPES;
```

This most probably won't let each of them typecheck independently.

To make each of the tuple items typecheck independently, the hardcoded signatures of 1 to whatever reasonable N number of type arguments can be added [like here](https://github.com/gcanti/io-ts/blob/0.6.x/src/index.ts#L560-L576), but not a generic solution until [variadic generics](https://github.com/Microsoft/TypeScript/issues/5453) are implemented into TypeScript.

bricks/template/tuple.h
```diff
@@ -47,6 +47,44 @@ static_assert(is_std_tuple<std::tuple<int, std::string>>::value, "");
 static_assert(!is_std_tuple<int>::value, "");
 static_assert(!is_std_tuple<std::string>::value, "");
 
+template <typename T>
+struct wrapped_into_tuple_impl {
+  using type = std::tuple<T>;
```
mzhurovich: `type` -> `tuple_t`?

typesystem/types.h
```diff
@@ -62,6 +62,18 @@ struct CurrentVariant : CurrentSuper {};
 
 #define IS_EMPTY_CURRENT_STRUCT(T) (::current::reflection::IsEmptyCurrentStruct<T, IS_CURRENT_STRUCT(T)>::value)
 
+// The `_CURRENT_` part of the name is only here to not pollute the "global macros namespace".
+// The macro just checks whether `T` is an `std::tuple<...>`.
+template <typename>
+struct CurrentTupleDetector { constexpr static bool value = false; };
+template <typename... TS>
+struct CurrentTupleDetector<std::tuple<TS...>> { constexpr static bool value = true; };
+#define IS_CURRENT_TUPLE(T) (::crnt::CurrentTupleDetector<::current::decay<T>>::value)
```
mzhurovich: There's no such thing as "Current tuple", right? Maybe just add `is_tuple` to `tuple.h`?

typesystem/serialization/test.cc
dkorolev: That's generally how we do it in the tests, to make the test body code as clean and uncluttered of an example code as possible.

dkorolev:
@mzhurovich Thx && PTAL.
@sompylasar Once we're on it, the `tuple`-s exported type will be fully specified. No time and/or need for it yet; tuples are not even supported in `TypeID` or schema exports now, to begin with.

mzhurovich:
Thank you @dkorolev!
Without TypeID support tuples look more like "third-world country citizens" :)
Anyway, it's helpful and sufficient for now.
Merging.

PR #815, opened 2018-04-10
Master flip over http (stream only).

grixa:
.


stream/replicator.h
```diff
@@ -403,15 +464,332 @@ struct StreamReplicatorImpl {
...
+    if (exposed_via_http_) {
+      CURRENT_THROW(StreamIsAlreadyExposedException());
+    }
+
+    exposed_via_http_ = std::make_unique<ExposedStreamState>(
+        port, route, std::move(restrictions), flip_started, flip_finished, flip_canceled);
+    exposed_via_http_->routes_scope_ +=
+        HTTP(port).Register(route, URLPathArgs::CountMask::None | URLPathArgs::CountMask::One, *Value(stream_)) +
+        HTTP(port).Register(
+            route + "/become/master", URLPathArgs::CountMask::None, [this](Request r) { MasterFlipRequest(r); });
```
dkorolev: Is the construction of the URL of two pieces simply due to the fact that the `URLPathArgs::CountMask::None | URLPathArgs::CountMask::One` mask for the original subscription URL would "intercept" any call with one URL parameter?

I can see the logic, but `"/become/master"` sort of invites other `"/become/*"`, while `"/become/follower"` doesn't make immediate sense to me (and neither do any other `"/become/..."`-s). Unless there's deeper logic here, perhaps the URL would be cleaner if it's some `"/control/flip_to_master"`?

stream/replicator.h
```diff
@@ -332,19 +377,35 @@ class SubscribableRemoteStream final {
...
+    return RemoteSubscriberScope<F>(stream_, subscriber, start_idx, from_us, subscription_mode, done_callback);
+  }
+
+  template <typename F, ReplicationMode RM>
+  void FlipToMaster(F& subscriber, head_optidxts_t head_idxts, uint64_t key, SubscriptionMode subscription_mode) const {
+    static_assert(current::ss::IsStreamSubscriber<F, entry_t>::value, "");
+    const auto response = HTTP(GET(stream_->GetFlipToMasterURL(head_idxts, key, subscription_mode)));
+    if (response.code == HTTPResponseCode.OK) {
+      const auto start_idx = Exists(head_idxts.idxts) ? Value(head_idxts.idxts).index + 1 : 0;
+      RemoteStreamSubscriber<F, entry_t, RM> remote_subscriber(stream_, subscriber, start_idx);
```
dkorolev: Nit: These two statements can be a one-liner, as all you need to do it call `.PassChunkToSubscriber()` once on a temporarily-created `RemoteStreamSubscriber<>`.

stream/replicator.h
```diff
@@ -80,36 +82,155 @@ class SubscribableRemoteStream final {
...
+    std::string carried_over_data_;
+
+   private:
+    template <ReplicationMode MODE = RM>
+    ENABLE_IF<MODE == ReplicationMode::Checked> PassEntryToSubscriber(const std::string& raw_log_line) {
+      const auto split = current::strings::Split(raw_log_line, '\t');
+      if (split.empty()) {
+        CURRENT_THROW(RemoteStreamMalformedChunkException());
+      }
+      const auto tsoptidx = ParseJSON<ts_optidx_t>(split[0]);
```
dkorolev: I understand it's copy-pasted (well, moved!), but, in `Checked` mode, shoudn't we also proxy a possible JSON parsing exception as an `RemoteStreamMalformedChunkException` one?

stream/replicator.h
```diff
@@ -80,36 +82,155 @@ class SubscribableRemoteStream final {
...
+        carried_over_data_ = chunk.substr(start_pos);
+      } else {
+        carried_over_data_.clear();
+      }
+    }
+
+   protected:
+    BorrowedWithCallback<RemoteStream> borrowed_remote_stream_;
+    F& subscriber_;
+    uint64_t index_;
```
dkorolev: Perhaps rename this `index_` into some `next_expected_index_from_remote_`?

stream/replicator.h
```diff
@@ -403,15 +464,332 @@ struct StreamReplicatorImpl {
...
+    if (client_head > head_idxts.head) {
+      r("", HTTPResponseCode.BadRequest);
+      return;
+    }
+    if (has_i && (!Exists(head_idxts.idxts) || next_index > Value(head_idxts.idxts).index + 1)) {
+      r("", HTTPResponseCode.BadRequest);
+      return;
+    }
+    if (Exists(head_idxts.idxts)) {
+      const auto next_index_by_timestamp =
```
dkorolev: This condition with `std::min` reads black-magic-ish to me. A comment would help.

stream/replicator.h
```diff
@@ -80,36 +82,155 @@ class SubscribableRemoteStream final {
...
+          CURRENT_THROW(StreamTerminatedBySubscriber());
+        }
+      } else {
+        if (split.size() != 1u || (from_us_.count() > 0 && tsoptidx.us < from_us_)) {
+          CURRENT_THROW(RemoteStreamMalformedChunkException());
+        }
+        if (subscriber_(tsoptidx.us) == ss::EntryResponse::Done) {
+          CURRENT_THROW(StreamTerminatedBySubscriber());
+        }
+        from_us_ = tsoptidx.us + std::chrono::microseconds(1);
```
dkorolev: Do we allow multiple `UpdateHead`-s with the same timestamp (and will they be persisted)? If yes, this may fail.

stream/replicator.h
```diff
@@ -80,36 +82,155 @@ class SubscribableRemoteStream final {
...
+        if (subscriber_(raw_log_line, index_++, unused_idxts_) == ss::EntryResponse::Done) {
+          CURRENT_THROW(StreamTerminatedBySubscriber());
+        }
+        from_us_ = std::chrono::microseconds(0);
+      } else {
+        const auto tsoptidx = ParseJSON<ts_only_t>(raw_log_line);
+        if (subscriber_(tsoptidx.us) == ss::EntryResponse::Done) {
+          CURRENT_THROW(StreamTerminatedBySubscriber());
+        }
+        from_us_ = tsoptidx.us + std::chrono::microseconds(1);
```
dkorolev: Do we allow multiple `UpdateHead`-s with the same timestamp (and will they be persisted)? If yes, this may fail.

stream/replicator.h
```diff
@@ -403,15 +464,332 @@ struct StreamReplicatorImpl {
...
+            route + "/become/master", URLPathArgs::CountMask::None, [this](Request r) { MasterFlipRequest(r); });
+    return exposed_via_http_->flip_key_;
+  }
+
+  void StopExposingViaHTTP(uint16_t port, const std::string& route) {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (exposed_via_http_ && exposed_via_http_->port_ == port && exposed_via_http_->route_ == route) {
+      exposed_via_http_ = nullptr;
+    } else {
+      CURRENT_THROW(StreamIsNotExposedException());
```
dkorolev: Note: The exception type is misleading, as the stream may well be exposed by on a different route/port.

stream/test.cc
```diff
@@ -1345,6 +1351,433 @@ TEST(Stream, ParseArbitrarilySplitChunks) {
...
+
+  current::stream::MasterFlipController<stream_t> stream2(stream_t::CreateStream(stream2_file_name));
+  // The second stream is master now, so it has no reason to flip.
+  ASSERT_THROW(stream2.FlipToMaster(flip_key1), current::stream::StreamIsAlreadyMasterException);
+  auto publisher = stream2.Stream().BecomeFollowingStream();
+  // Now the stream should be following.
+  EXPECT_FALSE(stream2.IsMasterStream());
+  // But it doesn't follow any remote stream, so it still can't perform the flip.
+  ASSERT_THROW(stream2.FlipToMaster(flip_key1), current::stream::StreamDoesNotFollowAnyoneException);
+  // Can't call the `FollowRemoteStream` now, cause it will hung in the `BecomeFollowingStream`.
```
dkorolev: _(Only oversaw it, not reading the test in details yet, the comment may be off.)_

Is `FollowRemoteStream` only used from the unit tests? If yes, it IMHO deserves a longer name highlighting this very fact, as well as a detailed comment.

stream/replicator.h
```diff
@@ -403,15 +464,332 @@ struct StreamReplicatorImpl {
...
+    if (!succeeded) {
+      // If something went wrong, restore the original state.
+      borrowed_publisher_ = nullptr;
+      stream_->BecomeMasterStream();
+      if (exposed_via_http_->flip_canceled_callback_) exposed_via_http_->flip_canceled_callback_();
+    } else if (exposed_via_http_->flip_finished_callback_) {
+      // Otherwise notify about flip completion from a separate thread
+      // to make it possible to unregister endpoints
+      // or even destroy the stream from inside the callback.
+      std::thread(exposed_via_http_->flip_finished_callback_).detach();
```
dkorolev: Likely the cleanest solution for now, but could you add some comment re. `cron` or some `NextTick()` logic?

If scheduled callbacks with a thread pool would become a norm in Current, this place is a prime candidate to save on spawning a new thread, IMHO. If that would be happening, I'll `grep -R` in Current for at least `cron`, `-i nexttick` and `-i next_tick` :-)

stream/replicator.h
```diff
@@ -403,15 +464,332 @@ struct StreamReplicatorImpl {
...
+    }
+
+    std::lock_guard<std::mutex> lock(mutex_);
+    const bool has_borrowed_publisher = Exists(borrowed_publisher_);
+    try {
+      remote_follower_ = std::make_unique<RemoteStreamFollower>(
+          has_borrowed_publisher ? std::move(Value(borrowed_publisher_)) : stream_->BecomeFollowingStream(),
+          url,
+          stream_->Data()->Size(),
+          stream_->Data()->CurrentHead() + std::chrono::microseconds(1),
```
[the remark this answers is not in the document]
grixa: Hm, we designed `UpdateHead` as an event with strictly monotone (increasing) timestamps. In other words, the persister treats it as an error if the timestamp in the current `UpdateHead` or the next entry is not greater than the previous head (or the timestamp of the last entry). Actually, I spent quite a lot of time debugging corresponding corner cases in combination with master-flollower flips.

stream/replicator.h
grixa: Yes, it was the only reason of this two-pieces URL construction. `"/control/flip_to_master"` sounds less confusing for me, too.

stream/replicator.h
grixa: Well, I think `next_expected_index_` should be enough, taking in account the fact that the class itself is called `RemoteStreamSubscriber`. BTW, what about `from_us_` member then? :)

stream/replicator.h
grixa: Actually, I think may be we don't need this "extra safety". Well, originally I thought we could extend the `MasterFlipController` to allow more than one `exposed_via_http_` port+route. If it's useless, let's just make `StopExposingViaHTTP` be with an empty arguments list.

stream/replicator.h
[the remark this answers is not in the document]
dkorolev: Got it, let's keep it this way then.

stream/replicator.h
dkorolev: Agree, let's remove this extra precaution.

dkorolev:
@grixa, thanks! Please resolve the conflict, there was literally one minor comment, and let's merge this one in!

Dima

dkorolev:
Thanks @grixa!

The tests after my minor commit have failed on Travis, but that seems intermittent. Just clicked "restart" on those two, fingers crossed ...

dkorolev:
Okay, restarting tests on Travis did the job. Merging. Thanks again @grixa.

PR #817, opened 2018-04-15
Refactored the names regexes library to eliminate possible construction of `std::string`-s.

dkorolev:
Hi @mzhurovich and @nyospe,

I've reworked the regexes library last week to eliminate room for possibly having an `std::string` constructed in the meantime. PTAL.

It also makes the API a bit bore verbose, hence I'd rather merge this in sooner, to make sure further usage of this code goes through the reworked API (although most usecases should be unaffected).

Thanks,
Dima

### DIFF
--- a/bricks/strings/chunk.h
+++ b/bricks/strings/chunk.h
@@ -88,6 +88,11 @@ class Chunk {
 
   const char* c_str() const { return S; }
 
+  // NOTE(dkorolev): We may need these later, but I'll comment them out for now,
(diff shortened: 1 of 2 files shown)


mzhurovich:
@dkorolev Travis says:
```
make[1]: *** [test] Segmentation fault (core dumped)
```
:-/

dkorolev:
Okay, disregard this one. The improvement was minor, and is no longer necessary for the external-to-Current logic I was using it from.

PR #818, opened 2018-05-04
Fixed Non-POD Optional assignment that caused broken references.

mzhurovich:
Our unit-tests were not good enough to catch this :(


dkorolev:
:+1:, and pls see https://github.com/C5T/Current/pull/819.

PR #821, opened 2018-05-06
Wrappers over cals to `::popen`, `::system`, and `::dl{open,sym,close}`.

dkorolev:
The `dlopen`-related are essential, but I've wrapped them all just in case.

Thanks,
Dima


bricks/system/syscalls.h
```diff
@@ -0,0 +1,214 @@
...
+
+class InputTextPipe final {
+ private:
+  const std::string command_;
+  std::vector<char> buffer_;
+  FILE* f_;
+
+ public:
+  template <typename S>
+  explicit InputTextPipe(S&& command, size_t max_line_length = 1024 * 1024)
```
mzhurovich: Using large buffers is usually a bad idea.
Typical buffer size for various file/network operations is 1-16K

dkorolev: True, but this is not just the buffer size, it's the maximum line length for `fread`; the output would make little sense if at least one line exceeds it.

Changed to 16K in a84c2d39 regardless. :)

bricks/system/syscalls.h
```diff
@@ -0,0 +1,214 @@
...
+
+struct PopenCallException final : SystemException {
+  using SystemException::SystemException;
+};
+
+struct SystemCallException final : SystemException {
+  using SystemException::SystemException;
+};
+
+class InputTextPipe final {
```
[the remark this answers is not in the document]
dkorolev: :+1:

How about `SystemCallReadPipe`? Ref. a84c2d39.

examples/dynamic_dlopen/impl.h
```diff
@@ -0,0 +1,90 @@
...
+
+#include "../iris/data/dataset.h"
+
+class DynamicDLOpenIrisExampleImpl {
+ private:
+  using F = void (*)(const Schema&, std::ostringstream&);
+
+  // clang-format off
+  const std::string prefix_ = current::Base64Decode("I2luY2x1ZGUgPHZlY3Rvcj4KI2luY2x1ZGUgPHNzdHJlYW0+CiNpbmNsdWRlIDxpb21hbmlwPgoKc3RydWN0IEJhc2UgewogIHZpcnR1YWwgfkJhc2UoKSB7fQp9OwoKc3RydWN0IEZsb3dlciA6IEJhc2UgewogIHZpcnR1YWwgfkZsb3dlcigpIHt9CiAgdW5pb24gewogICAgc3RydWN0IHsKICAgICAgZG91YmxlIHNsOwogICAgICBkb3VibGUgc3c7CiAgICAgIGRvdWJsZSBwbDsKICAgICAgZG91YmxlIHB3OwogICAgfTsKICAgIGRvdWJsZSB4WzRdOwogIH07CiAgc3RkOjpzdHJpbmcgbGFiZWw7Cn07CgpleHRlcm4gIkMiIHZvaWQgUnVuKGNvbnN0IHN0ZDo6dmVjdG9yPEZsb3dlcj4mIGZsb3dlcnMsIHN0ZDo6b3N0cmluZ3N0cmVhbSYgb3MpIHsK");
+  const std::string suffix_ = "\n}\n";
```
mzhurovich: Use `{prefix,suffix}.inl` instead of hard coding huge base64 strings?

bricks/system/syscalls.h
mzhurovich: The thing is that in this code it's both, max line length **and** the buffer size.
Proper buffering implementation may be added later though.

examples/dynamic_dlopen/impl.h
dkorolev: Then the code would become base-directory-dependent, or I'd have to incorporate the `argv[0]` logic, not to mention the binary would stop working by itself.

I could add a unit test to confirm these constants are exactly the base64 representations of the `.inl` files, what do you think about it?

bricks/system/syscalls.h
dkorolev: Yes, definitely not now.

examples/dynamic_dlopen/impl.h
dkorolev: @mzhurovich, a better idea could be to augment our default `Makefile[Impl]` with the logic to wrap up every `.inl` within the directory into some `.gitignore`-d header file containing their base64 representations prior to any compilation.

This way we get the best of all three worlds: The "constants" are automatically initialized with the up-to-date `.inl`-files, the `.inl` files can even be binary, and yet the resulting output binaries are self-contained, with no dependencies on any external files.

That's how it's done for static content in FD, and I've had no issues whatsoever, on both Linux and Mac.

Again, now doesn't seem the right time to be doing that. Just an option to keep in mind.

PR #823, opened 2018-05-12
DataFest 2018.

dkorolev:
Hi @mzhurovich,

4K lines sharp :-) These were a worthy three weekends.

Thanks,
Dima


dkorolev:
@mzhurovich this version should be good enough to merge, and it passes `make check`.

PR #828, opened 2018-06-23
Fixed several compilation warnings discovered by the llvm 9.1.0 clang.

grixa:
Hi @dkorolev and @mzhurovich . I've updated my xCode and got several new warnings. Most of them were about unused variables captured by lambdas. I removed them, but if their presence is important for some reasons, we can also declare them unused explicitly.


dkorolev:
:+1:

dkorolev:
Thanks @grixa!

PR #829, opened 2018-07-01
Fixing `make test`.

dkorolev:
Hi @mzhurovich,

Fixing `make test` and its coverage report (and also kicking off Semaphore, as it does weird things in a parallel PR).

Thanks,
Dima

### DIFF
--- a/bricks/dot/graphviz.h
+++ b/bricks/dot/graphviz.h
@@ -80,7 +80,7 @@ struct Node {
     return Set("shape", std::forward<T>(shape));
   }
 
-  Node& HTML() {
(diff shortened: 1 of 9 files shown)


dkorolev:
@mzhurovich -- `DLOpenExample.Smoke` still fails, and Semaphore still does something weird.

Please do not merge, I'll take a better look from where Internet actually works.

dkorolev:
`make test` is fixed in the previous previous commit, and Semaphore seems to be fixed by adding `sudo apt-get update` before the `g++-5` install script.

PR #830, opened 2018-07-03
Terminal colors and a progress tracker helper.

dkorolev:


blocks/xterm/example.cc
```diff
@@ -0,0 +1,104 @@
...
+      progress << u8"Йоу" << '!';
+      wait();
+      progress << u8"Йоу" << ", world!";
+      wait();
+    }
+    std::cout << "Done." << std::endl;
+  }
+  {
+    // Variable length characters.
+    // NOTE(dkorolev): This fails on my Linux. :/ Prints `Test3> tttesting ... OK`, then `Test3> ttDone.` at the end.
```
sompylasar: Have a look: https://github.com/sindresorhus/string-width/blob/master/index.js

blocks/xterm/vt100.h
```diff
@@ -0,0 +1,94 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*******************************************************************************/
+
+// Here's the best description I found: `https://misc.flogisoft.com/bash/tip_colors_and_formatting` -- D.K.
```
sompylasar: :+1:

blocks/xterm/example.cc
dkorolev: The issue is with Unicode in this case, not control characters. The latter are safely ignored by themselves. (And even UTF8 seems to work just fine with Cyrillic, just not with Japanese.)

dkorolev: To elaborate: JavaScript strings are "one index one Unicode character", while in Current it's UTF8.

mzhurovich:
Rainbow

dkorolev:
:trollface: 

@sompylasar would support me here. These colors will be put to the good use!

blocks/xterm/example.cc
sompylasar: I wonder why it won't work though. There's a notion of code points which are 1 to 2 UTF-16 characters, and 1 to 4 UTF-8 characters. Does Current consider that?

dkorolev: Current deals with UTF8, and this functionality (`UTF8StringLength`) is extensively tested and used in other places (FriendlyData).

dkorolev: I'm afraid this has to do with the behavior of the terminal (wrt how many `\b`-s does it need), not with how string length is computed.

Thus: a "frontend" issue which I could not care less about :)

dkorolev: > That's good. Tried Japanese in its tests? I think I only saw Cyrillic.

[duh](https://github.com/C5T/Current/pull/830/files#diff-31bf74dccc14f47f7f925c6ceb14aa14R79)

> Of course, you haven't unlearned how to behave like a d*ck. ;)

You'd like this book.

image

sompylasar: In the UTF8StringLength tests, not the terminal progress tests. There's just a smoke with latin and cyrillic, or I'm missing something.

dkorolev: Did you click the "duh" link above?

sompylasar: Yes I did, it points to a few lines below this comment:
screen shot 2018-07-03 at 10 47 17 am

And here's what I'm talking about:
https://github.com/C5T/Current/blob/d38e4af8c876d92be948a89c22310231981f14db/bricks/strings/test.cc#L947-L950
screen shot 2018-07-03 at 10 48 00 am

dkorolev: So, your question was "Are Japanese characters tested wrt the `UTF8StringLength()` code?". Now I get it.

Maybe not in this repo, but we use UTF8 in prod, and I've double-checked the logic of UTF8 symbol boundaries, which makes me rather confident that the issue is not here.

PR #831, opened 2018-07-05
Asynchronous subscriber destruction in Sherlock HTTP API.

mzhurovich:
Not very elegant, but quick solution - `SubscriberScope::~ SubscriberScope()` was never called causing the memory leak.


dkorolev:
LGTM, waiting for the CI to complete ...

dkorolev:
Ran the test manually on Semaphore, they passed.

PR #839, opened 2019-01-31
Caught those `warning: delete called on non-final RTTIDispatcherBase<…

dkorolev:
…>, has virtual functions but non-virtual destructor.

God bless `clang version 6.0.0-1ubuntu2`, `g++ (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0` doesn't catch them.


mzhurovich:
@dkorolev Yay!

PR #841, opened 2019-03-07
`CURRENT_T_EXTRACT_SUBTYPE`.

dkorolev:
Hi @grixa, CC @mzhurovich,

Grixa, here's `CURRENT_T_EXTRACT_SUBTYPE`, to use in nontrivial cases of `CURRENT_STRUCT_T`-s.

In this PR, I've also research on another question you've asked offline -- the `using SUPER::SUPER` one. The TL;DR is that the `CURRENT_CONSTRUCTOR` macros have the SFINAE to disable them in various structure compilation modes. So, the easiest solution, I'm afraid, is to duplicate the declaration of `CURRENT_CONSTRUCTOR(...)`, and then use `... : SUPER(...)` in its initialization list.

Would appreciate it if you could test the code on Windows as well.

Thanks,
Dima


grixa:
Checked compilation on Windows, see 8a69136
CURRENT_EXTRACT_T_SUBTYPE is amazing, thanks!

BTW, VS doesn't like that gtest uses std::tr1 namespace. May be it's time to update it to a newer version.

grixa (approved):
++LGTM!

dkorolev:
@grixa -- upgraded `googletest` in https://github.com/C5T/Current/pull/843, PTAL.

PR #846, opened 2019-03-17
CURRENT_USE_BASE_CONSTRUCTORS

grixa:
Hi @dkorolev, CC @mzhurovich 

I added new macro `CURRENT_USE_BASE_CONSTRUCTORS` to fix that `using SUPER::SUPER` problem we had discussed before. It works on both Windows and Linux, although I have several concerns:
* Naming, as usual. `CURRENT_EXPOSE_BASE_CONSTRUCTORS` or `CURRENT_FORWARD_BASE_CONSTRUCTORS` might be better.
* I'm not sure why `CURRENT_STRUCT` and `CURRENT_STRUCT_T` use different names for similar entities: `CSSH_##s` and `CURRENT_STRUCT_T_SUPER_HELPER_##s`. If there is no special meaning behind this difference, we could make them equal and get rid of the second `CURRENT_USE_T_BASE_CONSTRUCTORS`

I also fixed `CURRENT_EXTRACT_T_SUBTYPE`, cause it didn't work when being used multiple times.

### DIFF
--- a/typesystem/struct.h
+++ b/typesystem/struct.h
@@ -440,19 +440,33 @@ struct CurrentStructFieldsConsistency<T, 0u> {
   constexpr static const char* CURRENT_EXPORTED_STRUCT_NAME() { return #original_struct_name "_Z"; } \
   using template_inner_t_internal = std::true_type;                                                  \
   using template_inner_t_impl = original_template_inner_type
+
+#define CURRENT_USE_BASE_CONSTRUCTORS(s)                                                                        \
+  using CRNT_super_t =                                                                                          \
+      typename std::conditional<std::is_same<INSTANTIATION_TYPE, ::crnt::r::DF>::value, SUPER, CSSH_##s>::type; \
+  using CRNT_super_t::CRNT_super_t
+
+#define CURRENT_USE_T_BASE_CONSTRUCTORS(s)                                                               \
+  using CRNT_super_t = typename std::conditional<std::is_same<INSTANTIATION_TYPE, ::crnt::r::DF>::value, \
+                                                 SUPER,                                                  \
+                                                 CURRENT_STRUCT_T_SUPER_HELPER_##s>::type;               \
+  using CRNT_super_t::CRNT_super_t
 #else
 // I sure hope this is how it should be. -- D.K.
 #define CURRENT_EXPORTED_TEMPLATED_STRUCT(original_struct_name, original_template_inner_type)        \
   constexpr static const char* CURRENT_EXPORTED_STRUCT_NAME() { return #original_struct_name "_Z"; } \
   using template_inner_t = original_template_inner_type
+
+#define CURRENT_USE_BASE_CONSTRUCTORS(s) using SUPER::SUPER
+#define CURRENT_USE_T_BASE_CONSTRUCTORS(s) using SUPER::SUPER
 #endif
 
-#define CURRENT_EXTRACT_T_SUBTYPE(subtype, exported_subtype)             \
-  struct CRNT_##exported_subtype_##_helper_t { using subtype = int; };   \
-  using exported_subtype =                                               \
-    typename std::conditional<std::is_same<T, ::crnt::r::DummyT>::value, \
-                              CRNT_##exported_subtype_##_helper_t,       \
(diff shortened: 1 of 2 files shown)


typesystem/struct.h
```diff
@@ -440,19 +440,33 @@ struct CurrentStructFieldsConsistency<T, 0u> {
...
+
+#define CURRENT_USE_BASE_CONSTRUCTORS(s)                                                                        \
+  using CRNT_super_t =                                                                                          \
+      typename std::conditional<std::is_same<INSTANTIATION_TYPE, ::crnt::r::DF>::value, SUPER, CSSH_##s>::type; \
+  using CRNT_super_t::CRNT_super_t
+
+#define CURRENT_USE_T_BASE_CONSTRUCTORS(s)                                                               \
+  using CRNT_super_t = typename std::conditional<std::is_same<INSTANTIATION_TYPE, ::crnt::r::DF>::value, \
+                                                 SUPER,                                                  \
+                                                 CURRENT_STRUCT_T_SUPER_HELPER_##s>::type;               \
```
dkorolev, line 456: :+1:

typesystem/struct.h
```diff
@@ -440,19 +440,33 @@ struct CurrentStructFieldsConsistency<T, 0u> {
...
+  using CRNT_super_t::CRNT_super_t
 #else
 // I sure hope this is how it should be. -- D.K.
 #define CURRENT_EXPORTED_TEMPLATED_STRUCT(original_struct_name, original_template_inner_type)        \
   constexpr static const char* CURRENT_EXPORTED_STRUCT_NAME() { return #original_struct_name "_Z"; } \
   using template_inner_t = original_template_inner_type
+
+#define CURRENT_USE_BASE_CONSTRUCTORS(s) using SUPER::SUPER
```
dkorolev, line 466: A comment explaining why this "simple" solution doesn't work on other platforms may help.

dkorolev (approved):
Thanks @grixa, nicely done, and extra kudos for the tests!

dkorolev:
Also, @grixa, how do you feel about adding a one-parameter `CURRENT_EXTRACT_T_SUBTYPE(subtype)`?

So that instead of `CURRENT_EXTRACT_T_SUBTYPE(type_t, type_t);` one could just say `CURRENT_EXTRACT_T_SUBTYPE(type_t);`?

You would be the best person to make it happen since, unlike our CI machines, you have Visual Studio at hand ;-)

dkorolev:
And re. `CURRENT_STRUCT_T_SUPER_HELPER_##s` vs. `CSSH_##s` -- they are the artifacts remaining from the times we suffered from slow compilation times and larger compiler memory footprints, and thought that shorter symbol names could help. This is the reason behind both `CSSH_` instead of `CURRENT_STRUCT_SUPER_HELPER_` and behind `namespace ::crnt::r` as a shortcut for `namespace current::reflection`. I believe upon reworking the internals of the reflector this is no longer the issue (ah, memories, I was `objdump`-debugging those `.o`-s from Vancouver Island in Canada, that must was quite a few years ago!).

And whether `CURRENT_STRUCT_SUPER_HELPER_` and `CSSH_`/`CURRENT_STRUCT_T_SUPER_HELPER_` can be collapsed into a single piece is an open question. We just haven't touched that code because it does the job just fine. If there's an obvious improvement you could think of, by all means, don't hesitate.

---

As for the macro name for your new core, I'm cool with `CURRENT_USE_BASE_CONSTRUCTORS`. @mzhurovich?

grixa:
Added `CURRENT_EXTRACT_T_SUBTYPE(subtype)`, PTAL f6f19f9.
Regarding `CSSH_`/`CURRENT_STRUCT_T_SUPER_HELPER_` I can only tell that using the same name for both of them will allow to get rid of the second version of `CURRENT_USE_T_BASE_CONSTRUCTORS` with `_T_` in it. Plus it might slightly improve the `CURRENT_STRUCT` macros readability, that's it. I'm not sure these reasons are enough to do this.

dkorolev:
Thanks @grixa! Looks good to merge from my end with a small comment nit.

PR #851, opened 2019-05-09
`RoundDoubleToString` to support negative numbers and an unary plus.

dkorolev:


dkorolev:
@mzhurovich, I'll merge it into `stable`, because something closed-source needs it. Thx in advance && pls post-review.

bricks/strings/rounding.h
```diff
@@ -36,14 +36,64 @@ namespace current {
...
+  CURRENT_ASSERT(n_digits <= 15);
+
+  bool negative = false;
+  if (value < 0) {
+    value = -value;
+    negative = true;
+  }
+
+  if (value < 1e-16) {
+    return "0";
```
mzhurovich, line 50: @dkorolev Does it make sense to return `+0` in case of `plus_sign = true`?

dkorolev: Good idea. Shall we wait until if and when it becomes important? :-)

PR #855, opened 2019-09-01
Added inplace `Base64{Encode,Decode}`.

dkorolev:


dkorolev:
@mzhurovich -- renamed `ZeroCopy*` into `*Into`, PTAL.

PR #861, opened 2019-09-14
Streamed sockets

dkorolev:
Hi Max,

Shaped the streamed sockets benchmarking code. Here's the teaser:

image

---

I have also used it to confirm the per-instance EC2 AWS network capacity numbers are as advertised by Amazon. They match. The numbers are:

* 1.2 gigabytes per second (~10 gigabits per second) via a single TCP connection, and
* 3.1 gigabytes per second (~25 gigabits per second) across all TCP connections.

The best way to see these numbers for yourself would likely be to unpack this archive of screenshots and swipe through them with left and right arrows: [ec2_speed_test.zip](https://github.com/C5T/Current/files/3613088/ec2_speed_test.zip). My test scenario is:

* Start one connection between two hosts, on port 9001. Observe 1.2GB/s.
* Start another one, on port 9002. Observe both at 1.2GB/s.
* Add the 3rd, 4th, and 5th ones, on ports 9003 to 9005. Observe the connections getting throttled, to the max. total throughput of 3.1GB/s.
* Then, begin terminating the connections, bottom to top. Observe the throttling do its job, allowing for higher per-connection throughput as the number of connections gets fewer and fewer.
* Obviously, when back to two, and then one connection, they each will show 1.2GB/s.

---

Test setup.

Instance type: `m4.16xlarge`.

Kernel: `Amazon Linux 2 AMI (HVM), SSD Volume Type - ami-0b69ea66ff7391e80 (64-bit x86) / ami-09c61c4850b7465cb (64-bit Arm)`

Commands:
```
sudo yum install -y git make clang
git clone -b streamed_sockets https://github.com/dkorolev/current
cd current/examples/streamed_sockets/speedtest
NDEBUG=1 make
```

I have two machines as above. On one I copy the `./.current/sender` into `~`, on the other I copy `./.current/receiver` into `~`. Then I run the following "send" commands in my five terminals on the left:

```
PEM=...
SENDER_EXTERNAL_IP=...
INTERNAL_SENDTO=172.30.0.157
ssh -i $PEM ec2-user@$SENDER_EXTERNAL_IP ./sender --host $INTERNAL_SENDTO --port {9001..9005}
```

And the following "receive" commands in my five terminals on the right:

```
PEM=...
RECEIVER_INTERNAL_IP=...
ssh -i $PEM ec2-user@$RECEIVER_INTERNAL_IP ./receiver --port {9001..9005}
```

When terminating them bottom-to-top, it's sufficient to Ctrl+C either the sender or the receiver, so I alternated those in my series of screenshots.

The test uses the `3b9a2aeddaddd6ae7b9ff0a1ead2f6cfde0dadfa` commit.

### DIFF
--- a/examples/streamed_sockets/speedtest/receiver.cc
+++ b/examples/streamed_sockets/speedtest/receiver.cc
@@ -17,14 +17,24 @@ DEFINE_double(max_seconds_of_silence, 1.5, "Terminate the connection if it sends
 int main(int argc, char** argv) {
   ParseDFlags(&argc, &argv);
 
+  using namespace current::net;
+  using namespace current::vt100;
+
+#ifndef NDEBUG
+  {
+    std::cout << yellow << bold << "warning" << reset << ": unoptimized build";
+#if defined(CURRENT_POSIX) || defined(CURRENT_APPLE)
+    std::cout << ", run " << cyan << "NDEBUG=1 make clean all";
+#endif
+    std::cout << std::endl;
+  }
+#endif
+
   std::chrono::microseconds const t_window_size(static_cast<int64_t>(FLAGS_window_size_seconds * 1e6));
   size_t const window_size_bytes = static_cast<size_t>(FLAGS_window_size_gb * 1e9);
   std::chrono::microseconds const t_output_frequency(static_cast<int64_t>(FLAGS_output_frequency * 1e6));
 
   std::vector<uint8_t> buffer(static_cast<size_t>(1e9 * FLAGS_receive_buffer_gb));
-
-  using namespace current::net;
-  using namespace current::vt100;
 
   current::ProgressLine progress;
 
--- a/examples/streamed_sockets/speedtest/receiver.cc
+++ b/examples/streamed_sockets/speedtest/receiver.cc
@@ -1,0 +1,96 @@
+#include <deque>
+
+#include "../../../blocks/xterm/progress.h"
+#include "../../../blocks/xterm/vt100.h"
+#include "../../../bricks/dflags/dflags.h"
+#include "../../../bricks/net/tcp/tcp.h"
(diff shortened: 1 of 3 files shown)


examples/streamed_sockets/speedtest/receiver.cc
```diff
@@ -0,0 +1,131 @@
...
+  {
+    std::cout << yellow << bold << "warning" << reset << ": unoptimized build";
+#if defined(CURRENT_POSIX) || defined(CURRENT_APPLE)
+    std::cout << ", run " << cyan << "NDEBUG=1 make clean all";
+#endif
+    std::cout << std::endl;
+  }
+#endif
+
+  std::chrono::microseconds const t_window_size(static_cast<int64_t>(FLAGS_window_size_seconds * 1e6));
```
mzhurovich: As far as I remember, we agreed to keep `const`s on the left hand side?

mzhurovich:
FTR, it gives me more than 3Gb/s on OS X loopback interface :)

examples/streamed_sockets/speedtest/receiver.cc
dkorolev: Fixed in `6932cf72`.

dkorolev:
> FTR, it gives me more than 3Gb/s on OS X loopback interface :)

10GB/s on my Ubuntu :-)

mzhurovich:
> > FTR, it gives me more than 3Gb/s on OS X loopback interface :)
> 
> 10GB/s on my Ubuntu :-)

Yeah, standard OS X TCP has always been not super fast

dkorolev:
@mzhurovich -- merged `upstream/stable` into this PR, fingers crossed for the tests to turn green.

mzhurovich:
Merging on ✅

PR #862, opened 2019-09-15
Making Travis a bit happier.

dkorolev:


bricks/util/test.cc
```diff
@@ -281,8 +281,11 @@ TEST(Util, Base64) {
   EXPECT_EQ(all_chars, Base64Decode(Base64Encode(all_chars)));
   EXPECT_EQ(all_chars, Base64URLDecode(Base64URLEncode(all_chars)));
 
+#ifndef NDEBUG
+  // The very `Base64DecodeException` is debug mode only.
```
mzhurovich, line 285: Hmm. What's wrong with this test?

dkorolev: Nothing except it was failing because in `NDEBUG=1`, because the exception is only thrown in debug node.

PR #863, opened 2019-09-15
Attempt to fix `make check` issues with include search path

mzhurovich:
TL;DR there's an issue on OS X when system headers do smth like `#include <time.h>` and there's our own `time.h` in the current dir, which goes first in the include search path order because of `-I.`. `-idirafter` should make current dir the very last in search path order.


dkorolev:
@mzhurovich LGTM++, feel free to merge while I'm AFK for the rest of the day.

dkorolev (approved):

PR #874, opened 2019-10-14
C++17 for `strings::Join` and `current::decay_t`

mzhurovich:
More vanilla c++17

### DIFF
--- a/bricks/strings/join.h
+++ b/bricks/strings/join.h
@@ -61,16 +61,28 @@ inline size_t StringLengthOrOneForChar(T&& param) {
   return StringLengthOrOneForCharImpl<rmref<T>>::Length(param);
 }
 
-template <bool CAN_GET_LENGTH>
-struct OptionallyReserveOutputBufferImpl {
-  template <typename CONTAINER, typename SEPARATOR>
-  static void Impl(std::string&, const CONTAINER&, SEPARATOR&&) {}
-};
+namespace sfinae {
 
-template <>
-struct OptionallyReserveOutputBufferImpl<true> {
-  template <typename CONTAINER, typename SEPARATOR>
-  static void Impl(std::string& output, const CONTAINER& components, SEPARATOR&& separator) {
+template <typename T, typename = void>
+struct is_container : std::false_type {};
+
+template <typename T>
+struct is_container<T, std::void_t<decltype(std::declval<T>().begin()),
+                                   decltype(std::declval<T>().end()),
+                                   typename T::value_type>> : std::true_type {};
+template <typename T>
+constexpr bool is_container_v = is_container<T>::value;
+
+template <typename T>
+constexpr bool is_container_of_strings =
+    is_container_v<T> && std::is_same_v<typename T::value_type, std::string>;
+
+}  // namespace sfinae
+
+template <typename CONTAINER, typename SEPARATOR>
+void OptionallyReserveOutputBuffer(std::string& output, const CONTAINER& components, SEPARATOR&& separator) {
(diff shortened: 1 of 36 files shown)


bricks/template/test.cc
```diff
@@ -274,3 +276,89 @@ TEST(TemplateMetaprogrammingInternalTest, NonemptyConstructorForCombiner) {
   EXPECT_EQ("foo", joined("test"));
   EXPECT_EQ(42, joined(0));
 }
+
+// ********************************************************************************
+// * `current::decay_t` test, moved here from `decay.h`.
+// ********************************************************************************
+template <typename A, typename B>
+inline constexpr bool is_same_v = std::is_same_v<A, B>;
```
dkorolev: I remember! This is the line that did not compile in C++14 on Windows, and I just used `std::is_same_v`.

bricks/template/test.cc
```diff
@@ -274,3 +276,89 @@ TEST(TemplateMetaprogrammingInternalTest, NonemptyConstructorForCombiner) {
...
+
+using namespace current;
```
dkorolev: Maybe put this part into some `namespace test_current_decay`? The seemingly-global `using namespace current` looks scary, even in the test (keep in mind we have the top-level `make test` target).

bricks/strings/join.h
```diff
@@ -61,58 +61,35 @@ inline size_t StringLengthOrOneForChar(T&& param) {
...
-struct IsContainerOfStringsImpl<T, true> {
-  enum { value = std::is_same_v<std::string, current::decay<decltype(*std::declval<T>().begin())>> };
-};
+constexpr bool is_container_v = is_container<T>::value;
 
 template <typename T>
-struct IsContainerOfStrings {
-  enum { value = IsContainerOfStringsImpl<T, HasBegin<T>::CompileTimeCheck(0)>::value };
-};
+constexpr bool is_container_of_strings =
```
dkorolev: `is_container_of_strings_v`?

bricks/strings/join.h
```diff
@@ -61,58 +61,35 @@ inline size_t StringLengthOrOneForChar(T&& param) {
...
 void OptionallyReserveOutputBuffer(std::string& output, const CONTAINER& components, SEPARATOR&& separator) {
   // Note: this implementation does not do `reserve()` for chars, the length of which is always known to be 1.
-  OptionallyReserveOutputBufferImpl<sfinae::IsContainerOfStrings<CONTAINER>::value>::Impl(
-      output, components, separator);
+  if constexpr(sfinae::is_container_of_strings<CONTAINER>) {
+    size_t length = 0;
+    for (const auto& cit : components) {
+      length += cit.length();
+    }
+    length += impl::StringLengthOrOneForChar(separator) * (components.size() - 1);
```
dkorolev: Need a check for `!components.empty()` above, right?

mzhurovich: Well, it's checked in the `Join()` function

dkorolev: Call me paranoid, but let's add this check into here as well? At least in `#ifndef NDEBUG` mode, perhaps?

dkorolev: (Or rename the function by adding some `Unchecked` or `Internal` to its name, with a comment that it would crash on an empty input).

mzhurovich: https://github.com/C5T/Current/pull/874/commits/e9d44966ee29c24f2a314e54062eac3f4450c80f

dkorolev:
LGTM

PR #881, opened 2020-07-25
C++17 touches.

dkorolev:
Making our code more up to date.


typesystem/typesystem.vcxproj
```diff
@@ -0,0 +1,107 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
mzhurovich, line 1: @dkorolev Why do we need this here? :)

dkorolev: Analogous to many other `.vcxproj`-s -- my intuition is that `.sln`-s are redundant here, but being able to run tests from individual directories from within Visual Studio is still a useful feature.

Maybe clean them up (or migrate to the up to date version of VS) in bulk in another PR?

mzhurovich: +1 to cleaning up later

PR #883, opened 2020-12-22
`ChunkedPOST` in addition to `ChunkedGET`.

dkorolev:
Hi @mzhurovich -- this is missing tests, but may well be good to go "as is", as I've confirmed it does the job (with `HTTP` examples from the [Docker API](https://docs.docker.com/engine/api/sdk/examples/)).

### DIFF
--- a/blocks/http/types.h
+++ b/blocks/http/types.h
@@ -88,32 +88,51 @@ struct GET : HTTPRequestBase<GET> {
   explicit GET(const std::string& url) : HTTPRequestBase(url) {}
 };
 
-struct ChunkedGET {
+template <typename T>
+struct ChunkedBase {
   const std::string url;
   std::function<void(const std::string&, const std::string&)> header_callback;
   std::function<void(const std::string&)> chunk_callback;
   std::function<void()> done_callback;
-  explicit ChunkedGET(const std::string& url,
+  explicit ChunkedBase(std::string url) : url(std::move(url)) {}
+  explicit ChunkedBase(std::string url,
                       std::function<void(const std::string&, const std::string&)> header_callback,
                       std::function<void(const std::string&)> chunk_callback,
                       std::function<void()> done_callback = []() {})
-      : url(url), header_callback(header_callback), chunk_callback(chunk_callback), done_callback(done_callback) {}
+      : url(std::move(url)),
+        header_callback(header_callback),
+        chunk_callback(chunk_callback),
+        done_callback(done_callback) {}
+  T& OnHeader(std::function<void(const std::string&, const std::string&)> header_callback) {
+    this->header_callback = header_callback;
+    return static_cast<T&>(*this);
+  }
+  T& OnChunk(std::function<void(const std::string&)> chunk_callback) {
+    this->chunk_callback = chunk_callback;
+    return static_cast<T&>(*this);
+  }
+  T& OnDone(std::function<void()> done_callback) {
+    this->done_callback = done_callback;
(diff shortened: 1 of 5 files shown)


dkorolev:
Upd: Per-line-parsing added, tests added, `Chunked{GET,POST}`-s extended to support dot notation -- good to review from my end. Thanks @mzhurovich!

bricks/strings/group_by_lines.h
```diff
@@ -0,0 +1,71 @@
...
+ private:
+  F f_;
+  std::string residual_;
+
+ public:
+  explicit GenericStatefulGroupByLines(F&& f) : f_(std::move(f)) {}
+  void Feed(const std::string& s) {
+    Feed(s.c_str());
+  }
+  void Feed(char const* s) {
```
mzhurovich: `const char*` for consistency?

blocks/http/types.h
```diff
@@ -88,16 +88,97 @@ struct GET : HTTPRequestBase<GET> {
...
+    return static_cast<T&>(*this);
+  }
+  T& OnDone(std::function<void()> done_callback) {
+    this->done_callback_impl = done_callback;
+    group_by_lines_.clear();
+    return static_cast<T&>(*this);
+  }
+
+  // TODO(dkorolev): Move this to `Chunk`-s if performance becomes the bottleneck.
+  T& OnLine(std::function<void(const char*)> line_callback) {
```
mzhurovich: Hmm, not sure why the API for lines is `const char*`, while other functions are `const std::string&`

bricks/strings/test.cc
```diff
@@ -991,3 +992,34 @@ TEST(UTF8StringLength, Smoke) {
   EXPECT_EQ(4u, current::strings::UTF8StringLength("test"));
   EXPECT_EQ(4u, current::strings::UTF8StringLength("тест"));
 }
+
+TEST(StatefulGroupByLines, SmokeTrivial) {
+  std::vector<std::string> lines;
+  StatefulGroupByLines splitter([&lines](const char* line) { lines.push_back(std::move(line)); });
```
mzhurovich: `.emplace_back(line)`?

bricks/strings/group_by_lines.h
```diff
@@ -0,0 +1,71 @@
...
+        residual_.clear();
+        ++s;
+      } else {
+        return;
+      }
+    }
+  }
+  ~GenericStatefulGroupByLines() {
+    if (!residual_.empty()) {
+      f_(residual_.c_str());
```
mzhurovich, line 64: Maybe add a comment explaining the behavior when there's no newline at the end of the string?

mzhurovich (commented):
Thanks for adding this. Left a few minor comments/questions

bricks/strings/group_by_lines.h
dkorolev: :+1: c5ec7c52

blocks/http/types.h
dkorolev: Right, let's keep them `const std::string&`-s to keep things simple.

bricks/strings/group_by_lines.h
dkorolev: :+1: c5ec7c52

bricks/strings/test.cc
dkorolev: :+1: c5ec7c52

dkorolev:
TY @mzhurovich, PTAL.

PR #884, opened 2021-01-26
Fixes for Docker API

dkorolev:
Hi @mzhurovich -- this is my poor attempt to make it possible for the "user space" code to declare the HTTP response a chunked one based on the headers seen so far.

### DIFF
--- a/blocks/http/chunked_demo.cc
+++ b/blocks/http/chunked_demo.cc
@@ -149,8 +149,9 @@ CURRENT_STRUCT(ExampleMeta) {
 };
 
 // TODO(dkorolev): Finish multithreading. Need to notify active connections and wait for them to finish.
-int main() {
-  HTTP(FLAGS_port)
+int main(int argc, char** argv) {
+  ParseDFlags(&argc, &argv);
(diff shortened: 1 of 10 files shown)


dkorolev:
@mzhurovich -- okay, this proved to be more complicated.

I ended up adding (partial) support for "upgraded" connections, which effectively are a "light", binary version of the chunked ones: just read from the socket, in binary format, until forever from the point where the (potentially binary) body begins.

This can be triggered with `Connection: upgrade`, or with `Upgrade: <anything>` headers, or via a per-header callback which would call `response_kind.MarkAsUpgraded()`, so that the user can employ their own, custom logic.

Yes, this is a hack. Yes, I've tested it, including with a delay, to ensure the body is indeed streamed.

No, I don't quite see a usecase for this except making the Docker "HTTP" API work ;-)

Thanks for understanding!
Dima

dkorolev:
Ok, I'll close this PR. Will leave the `fixes_for_docker_api` branch here, just in case we do need these changes in the future.

CC @mzhurovich

PR #887, opened 2021-03-05
Plenty of fixes to have the Windows build and Travis succeed.

dkorolev:
- Building tests on new OS X on Travis, as the "original" OS X there doesn't support `invoke_result_t`.
- Split the `inividual_tests` target into three for, surprise, OS X on Travis.
- Got rid of all global `using namespace ...`, as they break Visual Studio builds (on reflection, there may be a fix, but certainly not now).
- Guarded plenty of code with `#ifndef CURRENT_WINDOWS` for Visual Studio to build everything.
- Removed `.template X(...)`, as it doesn't build under MSVS; left `.X()`.
- Made the destructor of `CurrentSuper` `{}` instead of `= default`, as the latter breaks some `noexcept` rule (will investigate deeper later).
- Added a bunch of `static_cast<>`-s to fix plenty of warnings. (Ignore longer lines as I plan to move to `clang-format-10` soon, which will alter, well, everything.)
- Broke the unit tests for `storage` into several separate `test.cc` files for them to run on Travis. (10mins++ compilation time kills it, and its CPU is low.)
- Other kinds of warning fixes.


bricks/util/test.cc
```diff
@@ -806,7 +806,7 @@ namespace object_with_interface_test_namespace {
 struct NamespacedObjectWithInterface {
   int calls = 0;
   int total_calls() { return ++calls; }
-  int add(int a, int b) const { return a + b; }  // `static` won't work with the inner `std::result_of_t<>`. -- D.K.
+  int add(int a, int b) { return a + b; }
```
mzhurovich, line 809: `static` then?

mzhurovich (approved):
@dkorolev Thanks for fixing so many issues! LGTM, just one minor comment about missing `static`.
There's too many `static_cast`s to my taste, which probably means we need to figure out the right usage of `uint_*`/`size_t` in our interfaces later.

bricks/util/test.cc
dkorolev: It works with statics too now! That's a nice side effect of changing `std::result_of_t` into `std::invoke_result_t`.

dkorolev:
> ... we need to figure out the right usage of uint_*/size_t in our interfaces later.

The original idea was that `size_t` is better to not be put into `CURRENT_STRUCT`-s, as they depend on 32- vs. 64-bit architecture.

One solution could be to hard-wire `uint64_t` to `size_t` as long as it's `static_assert()`-ed that the latter is 64 bits (so we consciously ignore the warnings on 32-bit architectures, as we ourselves are only building 64-bit binaries). Not sure what to do with it, let me sleep over it.

PR #889, opened 2021-03-07
Introduced `ReserveLocalPort()`, added a parallel test, refactored away `PickPortForUnitTest()`, fixed Travis (again).

dkorolev:
Max,

Sorry, couldn't stop: prototyping how modules could talk to one another, and realized "hard-coded" test ports are now on the critical path. So I fixed it. :-)

1. Fully retired the ugly `PickPortForUnitTest()`.

To accomplish this I've added a special type for the "reserved port", which holds the port. Sockets, as well the HTTP server, can be move-constructed from it. And there is a global function, `current::net::ReserveLocalPort()`, which finds a free local port, honds on to it, and returns it.

Tested:

```
# Build a debug test binary of `bricks/net/tcp`.
g++ -g -std=c++17 -W -Wall -Wno-strict-aliasing   -o ".current/test" "test.cc" -pthread -ldl

# Run in several terminals in parallel.
ulimit -c unlimited ; \
rm -f core ; \
while true ; do \
  ./.current/test \
     --gtest_throw_on_failure \
     --gtest_catch_exceptions=0 || break ;
done
```

The "canonical" test use syntax is:

```
// Reserve the port.
auto reserved_port = current::net::ReserveLocalPort();
// Extract it as `int` (`uint16_t` works too, but `int` is "%s"-frienldy w/o extra casts).
const int port = reserved_port;
// If the "legacy" code requires "just" a port (and "spawns" an HTTP server itself),
// then create the HTTP server by `std::move()`-ing this reserved port into `HTTP()`.
auto& http_server = HTTP(std::move(reserved_port));
static_cast<void>(http_server);
```
I've also made it such that in order to use a bare port for a socket it should be wrapped into `current::net::BarePort(desired_port)`. Originally, I thought it'd only be a change for during my development, but ultimately I decided to keep it this way. Hope you'll endorse it, open to suggestions regardless.

2. Added the `individual_tests_all_8` make target, which can be run with `make -j`.

Running tests in parallel is now possible, as they don't conflict on ports. On my laptop now:

```
$ time make -j individual_tests_all_8
real	1m50.030s
user	9m12.683s
sys	0m32.194s
```

I've also changed the runs for OS X on Travis accordingly.

As in the previous PR: pls don't worry about possible long lines, `clang-format-10` will change everything regardless =)

### DIFF
--- a/bricks/net/tcp/impl/posix.h
+++ b/bricks/net/tcp/impl/posix.h
@@ -80,7 +80,7 @@ namespace net {
 enum class NagleAlgorithm : bool { Disable, Keep };
 
 const size_t kMaxServerQueuedConnections = 1024;
-const NagleAlgorithm kDisableNagleAlgorithmByDefault = NagleAlgorithm::Keep;
+const NagleAlgorithm kDefaultNagleAlgorithmPolicy = NagleAlgorithm::Keep;
 
 struct SocketSystemInitializer {
 #ifdef CURRENT_WINDOWS
@@ -104,7 +104,7 @@ class SocketHandle : private SocketSystemInitializer {
  private:
   struct InternalInit final {};
   explicit SocketHandle(InternalInit,
-                        NagleAlgorithm nagle_algorithm_policy = kDisableNagleAlgorithmByDefault)
+                        NagleAlgorithm nagle_algorithm_policy = kDefaultNagleAlgorithmPolicy)
     : socket_(::socket(AF_INET, SOCK_STREAM, IPPROTO_TCP)) {
     if (socket_ < 0) {
       CURRENT_THROW(SocketCreateException());  // LCOV_EXCL_LINE -- Not covered by unit tests.
@@ -155,7 +155,7 @@ class SocketHandle : private SocketSystemInitializer {
   struct BindAndListen final {};
   explicit SocketHandle(BindAndListen,
                         BarePort bare_port,
-                        NagleAlgorithm nagle_algorithm_policy = kDisableNagleAlgorithmByDefault,
+                        NagleAlgorithm nagle_algorithm_policy = kDefaultNagleAlgorithmPolicy,
                         int max_connections = kMaxServerQueuedConnections)
       : SocketHandle(InternalInit(), nagle_algorithm_policy) {
     sockaddr_in addr_server;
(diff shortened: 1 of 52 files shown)


bricks/net/tcp/impl/posix.h
```diff
@@ -49,24 +60,27 @@ typedef int SOCKET;
...
+#include "../../exceptions.h"
+#include "../../debug_log.h"
+
+#include "../../../util/random.h"
+#include "../../../util/singleton.h"
+
 namespace current {
 namespace net {
 
+enum class NagleAlgorithm : bool { Disable, Keep };
```
mzhurovich, line 80: 👍

bricks/net/tcp/impl/posix.h
```diff
@@ -49,24 +60,27 @@ typedef int SOCKET;
...
+
 const size_t kMaxServerQueuedConnections = 1024;
-const bool kDisableNagleAlgorithmByDefault = false;
+const NagleAlgorithm kDisableNagleAlgorithmByDefault = NagleAlgorithm::Keep;
```
mzhurovich: nit: `s/kDisableNagleAlgorithmByDefault/kDefaultNagleAlgorithmPolicy/`?

bricks/net/tcp/impl/posix.h
```diff
@@ -199,24 +249,27 @@ class SocketHandle : private SocketSystemInitializer {
...
   SocketHandle() = delete;
   void operator=(const SocketHandle&) = delete;
-  void operator=(SocketHandle&&) = delete;
+  void operator=(SocketHandle&& rhs) {
+    socket_ = rhs.socket_;
+    rhs.socket_ = static_cast<SOCKET>(-1);
+  }
 
-#ifndef CURRENT_WINDOWS
+#if 1  // ndef CURRENT_WINDOWS
```
mzhurovich: Looks like it works on Windows well? Do we still need an else part?

dkorolev: Yes, I've checked this on Windows. Let me check again and remove the `#else` one =)

bricks/net/tcp/impl/posix.h
dkorolev: :+1: e4f8009b

bricks/net/tcp/impl/posix.h
dkorolev: Thanks for pointing this our, Max: This particular piece was now OK to remove, but the test that opens two sockets on one port -- another place I've tried again with Windows -- still fails, so I've changed that too.

Waiting for the last commit to (hopefully) turn green on Travis and pushing the update.

dkorolev: :+1: f9d4dcaa54b179c9995f2a1a7c38cd74496146af

Thanks for pointing this out, Max: This particular piece was now OK to remove, but the test that opens two sockets on one port -- another place I've tried again with Windows -- still fails, so I've changed that too.

mzhurovich (approved):
LGTM!

PR #896, opened 2021-04-12
Moving to `clang-format-10`.

dkorolev:
Hi @mzhurovich,

This PR moves Current to `clang-format-10`. I've also tweaked some `.clang-format` rules, see https://github.com/dkorolev/Current/commit/a290019130a852ae5da9f55b2f12cd9a68deb9ab for how it reformats the code.

Thinking of wrapping every top-line `#include "${relative_path}/port.h"` into `clang-format`-disabling guards, because we used to make it an important rule that `port.h` is `#include`-d first (i.e. `#define NOMINMAX` on Windows). Although we might well live without this change and see if all goes well, right?

Thanks,
Dima

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,12 @@
 BasedOnStyle: Google
 ColumnLimit: 120
+
 BinPackParameters: false  # Allow either all parameters on one line or one parameter per line, no packing.
 BinPackArguments: false   # The same for function calls.
+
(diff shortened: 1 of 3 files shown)


dkorolev:
@mzhurovich, I'll merge this one, as `clang-format-3.6` is quite obsolete anyway.

PR #903, opened 2021-06-09
Added support for hints to `ReserveLocalPort`.

dkorolev:


bricks/net/tcp/impl/posix.h
```diff
@@ -322,48 +323,78 @@ class ReserveLocalPortImpl final {
...
+[[nodiscard]] inline ReservedLocalPort ReserveLocalPort(
+    uint16_t hint,
+    NagleAlgorithm nagle_algorithm_policy = kDefaultNagleAlgorithmPolicy,
+    MaxServerQueuedConnectionsValue max_connections = kMaxServerQueuedConnections) {
+  return current::ThreadLocalSingleton<impl::ReserveLocalPortImpl>().DoIt(hint, nagle_algorithm_policy, max_connections);
+}
+
+[[nodiscard]] inline ReservedLocalPort ReserveLocalPort(
+    NagleAlgorithm nagle_algorithm_policy,
+    uint16_t hint,
```
mzhurovich: Why do we need two signatures with swapped arguments?

dkorolev: Yes, it's exactly the plan, for backwards compatibility (with the code that used the implementation from my dev branch that was there before).

They are all `enum class`-es, so looks safe to me.

mzhurovich: Compatibility with some dev branch is not a great justification for tweaking public APIs IMO

dkorolev: :+1: 0280c6c6

PR #908, opened 2022-02-01
Updates to TypeScript schema exporters.

dkorolev:
Hi @mzhurovich -- this is very hacky, but it does the job of exporting a relatively clean TypeScript schema, 2022 style.

Hi @sompylasar -- no point in looking at code, but you may find it aesthetically pleasing to see how does `typesystem/schema/golden/smoke_test_struct.ts` look now; and, of course, if something's wrong there, feedback would be more than welcome, so feel more than free to assign yourself as a reviewer and point out the wrongs!

Thanks,
Dima


dkorolev:
@mzhurovich, okay, as I'm in the cleanup mode for `Curent`, since this is a one-commit PR, and since it's used in another repo, I'll just merge this in. We won't lose any code, as the history is retained in the main branch of this repo anyway.

PR #915, opened 2022-03-05
And ran the top-level `clang-format-10`, let's see ...

dkorolev:
@mzhurovich, can we live with this?


mzhurovich:
I think we can. There's a number of weird changes like https://github.com/C5T/Current/pull/915/files#diff-bbd9b03a7984ea436c79dbf7d01a4acc4d04ac3867542c70e506de6425027b4dL144-R149 but in general it looks fine.

PR #941, opened 2023-08-12
Adding support for `std::array<T, N>` to the `typesystem/`.

dkorolev:
Incomplete wrt schema export and evolution, but should be good enough for now.


typesystem/schema/smoke_test_struct.h
```diff
@@ -85,6 +85,11 @@ CURRENT_STRUCT(FullTest) {
   CURRENT_FIELD_DESCRIPTION(primitives, "A structure with a lot of primitive types.");
   CURRENT_FIELD(v1, std::vector<std::string>);
   CURRENT_FIELD(v2, std::vector<Primitives>);
+#if 0
```
dkorolev, line 88: This is what I am not adding support for (yet), full schema export yields some `UNKNOWN` types.

PR #950, opened 2023-12-25
Re-enabled CI with proper tests run via Github actions.

dimacurrentai:


dimacurrentai:
@mzhurovich -- let's get Current back to how it should be WRT CI/CD?

dimacurrentai:
@mzhurovich -- let's roll into 2024 with this!

mzhurovich:
Nice :)

PR #951, opened 2023-12-29
Added `-I` for `current/` if it's present in `.`, `..`, or `../..`.

dimacurrentai:
Hi @mzhurovich,

In my `CMakeLists.txt` experiments, as well as with [my dotfiles](https://github.com/dkorolev/dotfiles/blob/main/.ycm_extra_conf.py#L7-L9), I am now used to "just" writing:

```
#include "bricks/strings/split.h"
```

This works right away, even though, obviously, the code I work on is not in the "root" directory of the `current` repo. By works I mean:

* This line is not highlighted as incorrect, with some "header file not found" error,
* This header file is parsed correctly, and
* Code completion / suggestions work.

Please see the footnotes for the illustration.

Now, for "toy" projects, I often just `ln -sf ../current/scripts/MakefileImpl Makefile`. This makes it easy to use `:mak`, as well as `:mak test` from Vim, in no time.

However, this very `:mak` does not work now with Current's default `Makefile`, since no `-I` dirs are set.

The proposal is to add a bit of logic into Current's default `Makefile`, so that if `current` is to be found in `.` or in `..` or in `../..`, this path is automatically added as `-I` when invoking `g++` / `clang++`.

What do you think?

If you like the idea but dislike making this the behavior default, I'm also open to guarding it by some environmental flag, which I'll set in my `.bashrc` / `.zshrc`.

Thanks!
Dima

Footnotes:

image

image

### DIFF
--- a/scripts/MakefileImpl
+++ b/scripts/MakefileImpl
@@ -26,6 +26,14 @@ CPLUSPLUS?=g++
 CPPFLAGS=-std=c++17 -W -Wall
 CPPFLAGS+= -Wno-strict-aliasing  # To suppress FnCAS tight struct packing warnings. -- D.K.
 
+ifneq ($(wildcard current/port.h),)
+  CPPFLAGS+= -Icurrent/
+else ifneq ($(wildcard ../current/port.h),)
+  CPPFLAGS+= -I../current/
+else ifneq ($(wildcard ../../current/port.h),)
+  CPPFLAGS+= -I../../current/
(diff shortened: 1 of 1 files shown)


mzhurovich:
I have mixed feelings about it tbh, this change feels like it goes in the opposite direction of the maximum cleanness we've been always striving for... But if it helps with usability out of the box, well, why not :)

PR #960, opened 2024-01-03
Made the links behind badges to point exactly to the right branches.

dimacurrentai:
CC @mzhurovich -- merging this in right away too.


dimacurrentai:
(Well, I am merging this one in right away -- the links work! -- but only after the badge turns green. So waiting for the tests to finish now.)

PR #967, opened 2024-03-12
`cmake` cleanup

dkorolev:
Back to using Current from the default `C5T/current`, plus a few more simplifications.


dkorolev:
@mzhurovich -- merging this with your post-review to keep things clean.

PR #970, opened 2024-03-12
Using the "new" `current_build_info.h`.

dimacurrentai:
Fingers crossed this passes the GitHub PR guard now.

If yes, I'll remove the "old" `current_build.h` target and merge.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -56,13 +56,6 @@ endfunction()
 UseGitOrClone(https://github.com/c5t/current)
 UseGitOrClone(https://github.com/c5t/googletest)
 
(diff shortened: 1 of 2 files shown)


dimacurrentai:
Okay, it did not "work" as intended right away, BUT: I've managed to re-build this target using `cmake` means only, 100% of the time.

AND: I have double-checked that only the targets that directly `#include "current_build_info.h"` do get re-build every time.

Which **is** the ideal outcome as far as I can tell: whatever `#include`-s `current_build_info.h` gets rebuilt every time with the freshly re-generated `current_build_info.h`, whereas whatever does not `#include` `current_build_info.h` is kept up to date.

Of course, I intend to only require linking whatever needs `current_build_info.h`, with the `.cc` units that `#include` `current_build_info.h` to be as compact as possible. Amen.

dimacurrentai:
Yay, all green, merging!

PR #977, opened 2024-03-27
Cleaner, more verbatim HTTP response method signatures.

dkorolev:


dkorolev:
@mzhurovich I'll merge this one too, as it's more of a follow-up to https://github.com/C5T/Current/pull/976 than the actual storage benchmark fix part.

PR #979, opened 2024-03-28
Next steps with `cmake`-based builds.

dkorolev:
Incremental improvements with `cmake`:

1. Added `C5T_NO_GTEST`, since it's silly to wait ~8 extra seconds to build "Hello, World!", and
2. Made `lib_*` libraries depend on each other, so that they can use each other's functions.

Re. (2) I think `dlib_*.cc` targets should also depend on all the libraries. But also I am jet lagged, so baby steps.

Thx,
Dima

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -65,7 +65,13 @@ function(UseGitOrCloneBranch remote branch)
 endfunction()
 
 UseGitOrClone(https://github.com/c5t/current)
-UseGitOrClone(https://github.com/c5t/googletest)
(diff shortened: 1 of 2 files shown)


dkorolev:
@mzhurovich I'll merge this in for your post-review, wanna clean up a few examples before going further.

dkorolev:
For instance, https://github.com/dkorolev/c5t_trivial_dep_example/commit/9746eef68fa5f66a70b88f01cbed0114a6426249 is now under a second for `make clean; make`.

PR #980, opened 2024-03-28
Introduced a clean way to catch that the HTTP port is taken.

dkorolev:
Hi @mzhurovich , @amdrozdov,

Jet lagged. Current Therapy for the win!

Thx,
Dima

### DIFF
--- a/bricks/net/tcp/impl/posix.h
+++ b/bricks/net/tcp/impl/posix.h
@@ -380,6 +380,25 @@ class ReserveLocalPortImpl final {
       hint, nagle_algorithm_policy, max_connections);
 }
 
+// Acquires the local port with the explicltly provided number, or throws the exception.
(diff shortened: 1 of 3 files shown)


bricks/net/tcp/impl/posix.h
```diff
@@ -380,6 +380,25 @@ class ReserveLocalPortImpl final {
       hint, nagle_algorithm_policy, max_connections);
 }
 
+// Acquires the local port with the explicltly provided number, or throws the exception.
```
mzhurovich, line 383: typo: `s/explicltly/explicitly/`

dkorolev: Fixed in https://github.com/C5T/Current/pull/990, merging as soon as the tests turn green.

PR #981, opened 2024-03-28
Extensions to `WaitableAtomic`, and the thread pool example.

dkorolev:
[ After https://github.com/C5T/Current/pull/980 ]

Hi @mzhurovich , @amdrozdov, here's my incremental work on `WaitableAtomic`. This PR should be good to review per commit, and I will be adding commits one by one. Will keep it open so that Github actions keep doing their job.

Will comment down there once it's ready to review & merge!

Thx,
Dima


dkorolev:
:point_up: good to review, and per-commit would do just fine. Thanks!

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+#include <mutex>
+
+#include "../../blocks/http/api.h"
+#include "../../bricks/dflags/dflags.h"
+#include "../../bricks/sync/waitable_atomic.h"
+
+DEFINE_uint16(port, 8080, "The local port to use.");
+DEFINE_uint32(n, 3, "Max number of audio streams");
+
+const std::string cmd_kill = "kill";
```
dimacurrentai: Why a string constant if you could use `enum class`, or even a special `[unordered]set` for the IDs to be killed?

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+
+struct SharedState final {
+  uint32_t active_channels = 0;
+  std::map<std::string, std::string> channel_control;
+  bool channel_exists(std::string channel_id) const{
```
dimacurrentai: `const std::string&` or `std::string const&`, since just `std::string` as an argument makes a redundant copy.

Btw, maybe your code is better off in a separate repo, using the `CMakeLists.txt` from `C5T/Current/cmake/`, such as https://github.com/dkorolev/c5t_trivial_dep_example ?

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+      }
+      // Send kill signal to the channel
+      safe_state.MutableUse([channel_id](SharedState& state) {
+        state.channel_control[channel_id] = cmd_kill;
+      });
+      r("channel killed\n");
+    });
+
+    std::cout << "listening  up to " << FLAGS_n << " streams on port " << FLAGS_port << std::endl;
+    http.Join();
```
dimacurrentai: `http.Join()` will block forever, and `joiner.join()` will never get called -- not sure though, as it may well be what you intended, since you only `kill` individual connections, not the service itself.

Personally, I'm a fan of graceful termination!

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+          std::cout << "[" << channel_id << "] worker tick" << std::endl;
+
+          std::this_thread::sleep_for(std::chrono::milliseconds(2000));
+        }
+      });
+      r("created\n");
+    });
+
+    scope += http.Register("/kill", [&safe_state](Request r) {
+      std::string channel_id = r.body;
```
dimacurrentai: Maybe `URLPathArgs::CountMask::One` and use `r.url_path_args[0]`, for `localhost:8080/kill/{channel_id}`?

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+      auto has_channel = safe_state.MutableUse([channel_id](SharedState& state) { return state.channel_exists(channel_id); });
```
dimacurrentai: Nit: This particular use case looks like `ImmutableUse`, although why lock twice if you can do it once?

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+        }
+        for(auto& to_drop: to_join){
+          threads.erase(to_drop);
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+      }
+    });
+
+    auto scope = http.Register("/", [&safe_state, &threads](Request r) {
+      std::string channel_id = r.body;
```
dimacurrentai: Maybe use `URLPathArgs::CountMask::One` and `r.url_path_args[0]`, for `localhost:8080/{channel_id}` URLs?

If you take this path though, `/kill` might potentially conflict with channel ID. The RESTful standard would be that `/` is just some top-level HATEOAS [1] endpoint (effectively returning the list of endpoints), where you then have `/channel/${id}`.

(This way btw, contradictory to my comment below about `/kill`, the canonical deletion would be `curl -X DELETE localhost:8080/channel/{id}`, and `if (r.method == "DELETE")` in the code. Overall, even though I use it a lot myself, `GET` requests are supposed to be legal for proxies to re-try, so for mutating operations such as killing something a non-`GET`, such as `DELETE` or `POST` of `PUT` or `PATCH`, is generally recommended.)

[1] https://www.geeksforgeeks.org/hateoas-and-why-its-needed-in-restful-api/#, see the `"links"` part -- it's a good idea for the API to return its top-level links for the root endpoint if the real biz logic is under some `/channel/{channel_id}`, etc. Also, to create channels where the ID is assigned by the system, just `POST /channel` is the standard way, but I'm sure you know this part very well.

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+      if(n_chans == FLAGS_n){
+        r("error: too many channels\n");
+        return;
+      }
+
+      // Create new channel
+      safe_state.MutableUse([channel_id](SharedState& state) {
+        state.channel_control[channel_id] = "";
+        state.active_channels++;
+      });
```
dimacurrentai: Here you lock the mutex three times, while one may well be enough.

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+      threads[channel_id] = std::thread([&safe_state, channel_id]() {
```
dimacurrentai: And this code, I think, relies on the fact that the HTTP handler for a particular port is single-threaded. Which is correct, but why not put the threads into the `WaitableAtomic<>` state too?

examples/audio_server/server.cc
```diff
@@ -0,0 +1,132 @@
...
+              },
+              // TODO: reduce this time (only for debug)
+              std::chrono::seconds(1));
+          if(control.stop) {
+            std::cout << "Channel '" << channel_id << "' has been stopped" << std::endl;
+            break;
+          }
+
+          if(!control.apply.empty()) {
+            std::cout << channel_id << "received signal " << control.apply << std::endl;
```
dimacurrentai: This looks like long polling to me. Realistically, if you already have a thread for the channel, why not `std::move()` the `Request` into some queue managed by this thread?

By not doing so you are blocking the HTTP thread in your `while (true)`, which I believe is not the behavior you intend to see =)

examples/audio_server/server.cc
dimacurrentai: Although I've contradicted this comment above, and think `if (r.method == "DELETE")` would be cleaner!

dimacurrentai (commented):
Sorry for my jet lag, and hope the comments are useful!

examples/audio_server/server.cc
dimacurrentai: Also, is `active_channels` needed as a separate variable, if `.channel_control.size()` can (and should?) mean the same thing?

examples/audio_server/server.cc
dimacurrentai: I.e. it is "correct" so that `cout` output will not get garbled, but overall we can do better =)

examples/audio_server/server.cc
amdrozdov: Yes, because it's just using current. Let's do the review and I can move it to the separate repo

examples/audio_server/server.cc
amdrozdov: Cool idea - just to have /< channel > url with post/delete methods

examples/audio_server/server.cc
amdrozdov: done

examples/audio_server/server.cc
amdrozdov: done

examples/audio_server/server.cc
amdrozdov: done

examples/audio_server/server.cc
amdrozdov: in progress

examples/audio_server/server.cc
amdrozdov: done

examples/audio_server/server.cc
amdrozdov: done, moved threads to WaitableAtomic

examples/audio_server/server.cc
amdrozdov: reverted, need to think (maybe in separate commit), we can ask only for destination change (which is just reconnect)

examples/audio_server/server.cc
```diff
@@ -30,101 +31,113 @@ int main(int argc, char** argv) {
...
-        for (auto& pair : threads) {
-          std::string channel_id = pair.first;
-          auto has_channel = safe_state.MutableUse([channel_id](SharedState& state) {
-            return state.channel_exists(channel_id);
-          });
-          if (!has_channel) {
-            to_join.push_back(pair.first);
-            pair.second.join();
+        auto die = safe_state.MutableUse([](SharedState& state){
+          if(state.die){
```
dimacurrentai: Just `return state.die`, or even `= safe.ImmutableScopedAccessor()->die`?

examples/audio_server/server.cc
```diff
@@ -30,101 +31,113 @@ int main(int argc, char** argv) {
...
           }
-        }
-        for(auto& to_drop: to_join){
-          threads.erase(to_drop);
+          std::vector<std::string> to_join;
+          for (auto& pair : state.threads) {
+            std::string channel_id = pair.first;
+            if (!state.channel_exists(channel_id)) {
+              to_join.push_back(pair.first);
+              pair.second.join();
```
dimacurrentai: I'd recommend `.join()`-ing these threads from outside the locked section.

examples/audio_server/server.cc
```diff
@@ -30,101 +31,113 @@ int main(int argc, char** argv) {
...
-      if(has_channel){
-        r("error: channel already exists\n");
-        return;
-      }
-      uint32_t n_chans = safe_state.MutableUse([](SharedState& state) { return state.active_channels; });
-      if(n_chans == FLAGS_n){
-        r("error: too many channels\n");
-        return;
-      }
+      struct channel_validation{
```
dimacurrentai: Btw, if this one is a `CURRENT_STRUCT` (declared outside any function), you could just respond with it from the HTTP handler, and the caller will get the JSON.

examples/audio_server/server.cc
```diff
@@ -30,101 +31,113 @@ int main(int argc, char** argv) {
...
-          // TODO: route the stream here
-          std::cout << "[" << channel_id << "] worker tick" << std::endl;
-
-          std::this_thread::sleep_for(std::chrono::milliseconds(2000));
-        }
+        });
+        return channel_validation{true, "created"};
       });
-      r("created\n");
+      r(node_state.msg, (node_state.is_valid ? HTTPResponseCode.OK : HTTPResponseCode.BadRequest));
```
dimacurrentai: Yep, just `r(node_state [, ...])` would return the JSON as long as it's a `CURRENT_STRUCT`, ref. https://github.com/C5T/Current/blob/stable/blocks/http/test.cc#L74-L82

examples/audio_server/server.cc
```diff
@@ -30,101 +31,113 @@ int main(int argc, char** argv) {
...
     });
 
     scope += http.Register("/kill", [&safe_state](Request r) {
       std::string channel_id = r.body;
-      auto has_channel = safe_state.MutableUse([channel_id](SharedState& state) { return state.channel_exists(channel_id); });
+      auto has_channel = safe_state.ImmutableUse([channel_id](const SharedState& state) {
```
dimacurrentai: This is also a good use case for `ImmutableScopedAccessor()->` imho.

dimacurrentai:
Replaced this PR by https://github.com/C5T/Current/pull/989.

examples/audio_server/server.cc
amdrozdov: Wow, that's cool!

dimacurrentai: > Wow, that's cool!

Enjoy! There's also `class Response` FTR, so that if you want to `return` it from some lambda, and want to have different HTTP response codes and/or headers or content type, etc.

PR #989, opened 2024-03-30
Extensions to `WaitableAtomic`, and the thread pool example. 

dimacurrentai:


dimacurrentai:
This PR replaces https://github.com/C5T/Current/pull/981.

PR #994, opened 2024-04-12
Checking whether `IntrusiveClient` for `WaitableAtomic` is in use at all except the test.

dkorolev:
@mzhurovich I'm strongly considering removing it, since `Owned<>`/`Borrowed<>` (or the good old `shared_ptr<>`) appear to offer a cleaner way than this confusing intrusive thingy.

### DIFF
--- a/bricks/sync/test.cc
+++ b/bricks/sync/test.cc
@@ -253,6 +253,7 @@ TEST(OwnedBorrowed, UseInternalIsDestructingGetter) {
   thread->join();
 }
 
+#if 0
 TEST(WaitableAtomic, Smoke) {
   using current::IntrusiveClient;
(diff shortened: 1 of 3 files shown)


dkorolev:
(Of course, if the tests pass, I'll clean up the smoke test to actually run the test, just in a non-intrusive way.)

dkorolev:
Okay, seems to work, minor tweaks now ...

dkorolev:
@mzhurovich -- good to review per commit!

mzhurovich:
Agree, I'm all for cleanness and simplicity, and intrusive pointers are pretty niche thing, so I don't think we lose anything by removing this implementation here. LGTM.

PR #995, opened 2024-04-12
Added notifications to `WaitableAtomic<>`.

dkorolev:


bricks/sync/waitable_atomic.h
```diff
@@ -316,10 +334,38 @@ class WaitableAtomic {
...
+    WaitableAtomicSubscriberRemoverImpl(WaitableAtomic& self, size_t id) : self_(self), id_(id) {}
+    void Remove() override {
+      // Okay to only lock the subscribers map, but not the data.
+      std::lock_guard lock(self_.subscribers_mutex_);
+      self_.subscribers_.erase(id_);
+    }
+  };
+
+  [[nodiscard]]
+  WaitableAtomicSubscriberScope Subscribe(std::function<void()> f) {
```
mzhurovich, line 349: nit: Could be a bit confusing why subscribers aren't really getting any data as a part of subscription, but I can't come up with some better (but still concise) naming to clarify it's subscription to state change signals.

dkorolev: gmta, leaving this in the back of my mind for some further PRs

(this code need a test too btw, which is not there yet!)

PR #997, opened 2024-04-15
Simple async replication based on vector clock

amdrozdov:
[WIP]
Notes:
1. Implemented minimalistic vector clock (sending and storing only 1 timestamp, not full vector) (is it correct?)
2. Binary serialization for replication delivery

Do not forget:
1. Need to validate reuse port (it doesn't work with 3+ nodes)
4. After the review we can move the container class to blocks
5. Need to add unit tests
6. Hope to add template in order to customize key/value types


examples/async_replication/async_replicated_container.h
```diff
@@ -0,0 +1,421 @@
...
+
+    while (true) {
+      is_stop = state.ImmutableUse(
+          [](SharedState const&state){ return state.die;});
+      if(is_stop){
+        break;
+      }
+      try {
+        current::net::Socket socket((current::net::BarePort(port)));
+        // FIXME? how to accept multiple connections on same port with SO_REUSEADDR/REUSEPORT
```
dkorolev: Just from another thread, after one connection is already established.

examples/async_replication/async_replicated_container.h
```diff
@@ -0,0 +1,421 @@
...
+        }
+      }
+      std::this_thread::sleep_for(std::chrono::milliseconds(delay));
+    }
+    // stopped
+  }
+
+  Relay recv_relay(current::net::Connection &c){
+    // Get buffer length
+    char* len_buf = (char*)malloc(sizeof(size_t));
```
dkorolev: A `vector<uint8_t>` (or of any other type, really) is far safer than `malloc`. Although if you just need a `size_t`, declare `size_t len` and `BlockingRead(&len, sizeof(len)` into it.

examples/async_replication/async_replicated_container.h
```diff
@@ -0,0 +1,421 @@
+#pragma once
+
+#include "../../bricks/sync/waitable_atomic.h"
+#include "../../blocks/http/api.h"
+#include "../../bricks/time/chrono.h"
+
+#include <sstream>
+#include <queue>
+
+#define htonll(x) ((1==htonl(1)) ? (x) : (((uint64_t)htonl((x) & 0xFFFFFFFFUL)) << 32) | htonl((uint32_t)((x) >> 32)))
```
dkorolev: This can definitely be `template`-d and `constexpr`-d, so that the check is not performed on every call to them, and the call is inlined too.

examples/async_replication/async_replicated_container.h
```diff
@@ -0,0 +1,421 @@
...
+        current::net::Connection connection(socket.Accept());
+
+        if(is_verbose) {
+          state.MutableUse([port](SharedState& state) {
+            std::cout << "Reader connected on port " << port << std::endl;
+          });
+        }
+
+        while (waits < max_waits) {
+          is_stop = state.ImmutableUse(
```
dkorolev: Why do you name this variable at all, just curious/

examples/async_replication/async_replicated_container.h
amdrozdov: it works for vector<uint8_t> but returns a compilation error for reading directly to int:
```c++
size_t len_buf;
auto size = c->BlockingRead(&len_buf, sizeof(len_buf));
if (!size) throw;
len_buf = be64toh(len_buf);
```
returns
```
error: no matching function for call to ‘current::net::Connection::BlockingRead(size_t*, long unsigned int)’
  178 |     auto size = c->BlockingRead(&len_buf, sizeof(len_buf));
```

examples/async_replication/async_replicated_container.h
amdrozdov: done

examples/async_replication/async_replicated_container.h
amdrozdov: done. Found `htobe64` and `be64toh` function, removed this code

examples/async_replication/async_replicated_container.h
amdrozdov: Currently replaced all `malloc()` calls with vector<uint8_t>

examples/async_replication/async_replicated_container.h
amdrozdov: done, it works properly with multiple nodes

amdrozdov:
TODO:
1. Replace single timestamp validation with real vector clock
2. Replace hardcoded value type with std::string/blob and do serialization internally
3. Test cases

amdrozdov:
Ready to review

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
+/*******************************************************************************
+The MIT License (MIT)
+
+Copyright (c) 2014 Dmitry "Dima" Korolev <dmitry.korolev@gmail.com>
```
dkorolev: You really want your name here =)

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
...
+
+TEST(VectorClock, SmokeTest) {
+  auto v = VectorClock();
+  v.step();
+
+  // test merge from future - should return false
+  auto data = DiscreteClocks();
+  data.reset(1);
+  EXPECT_EQ(false, v.merge(data));
+  EXPECT_EQ(v.state().size(), (size_t)(1));
```
dkorolev: We try to follow the [Google Style Guide](https://google.github.io/styleguide/cppguide.html), which suggests `static_cast<size_t>(-1)`, but these are details.

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
...
+
+  // test lte (t==t')
+  auto v2 = VectorClock(data, data.size());
+  EXPECT_EQ(true, v2.merge(data));
+}
```
dkorolev, line 48: This test can be cleaner, IMHO.

The constructor with two arguments could be a single-argument constructor, or some `auto v2 = VectorClock::FromVector(data);`

The `data.reset(1)` part above could also be simplified.

Plus, it would help the reader to a) show some `.ToString()`-ified representation (i.e. `EXPECT_EQ()` it), and b) show what minor changes to the input(s) result in the `true` becoming `false` and vice versa.

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
...
+
+TEST(VectorClock, Merge) {
+  auto base_time = current::time::Now();
```
dkorolev, line 51: There's `CURRENT_MOCK_TIME` and `SetNow()` FWIW.

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
...
+  // can't merge T > T'
+  EXPECT_EQ(false, v.merge(c2));
+  // invalid data, merged vector
+  EXPECT_EQ(v.state()[0], cur_state[0] + 1);
+  EXPECT_EQ(v.state()[1], cur_state[1]);
+
+  // Merge partially equals using lte validation
+  v = VectorClock(c1, 0);
+  c2 = {1, 3};
+  // 0 is equeal, 1 is greater - ok to merge
```
dkorolev: Typo.

Btw, `EXPECT_EQ()` et. al. "return" a stream, i.e. `EXPECT_EQ(4, 2*2) << "Two times two should be four.";` -- sometimes I find this better, since the string here is more of a piece of "code" than "just" a comment. Style nit though.

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
...
+  c2 = DiscreteClocks({1, 0});
+  EXPECT_EQ(false, v.merge(c2));
+  // invalid data, merged vector
+  EXPECT_EQ(v.state()[0], cur_state[0] + 1);
+  EXPECT_EQ(v.state()[1], cur_state[1]);
+}
+
+TEST(VectorClock, ContinuousTime) {
+  auto base_time = current::time::Now();
+  ContinuousClocks c1 = {base_time, base_time + std::chrono::microseconds(100)};
```
dkorolev: How do you feel about `+100us`?

https://en.cppreference.com/w/cpp/chrono/operator%22%22s

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
...
+  auto v = VectorClock<ContinuousClocks, MergeStrategy>(c1, 0);
+
+  // Merge correct update
+  ContinuousClocks c2 = {base_time + std::chrono::microseconds(200), base_time + std::chrono::microseconds(300)};
+  // each element is greater - ok to merge
+  EXPECT_EQ(true, v.merge(c2));
```
dkorolev: `EXPECT_TRUE`.

bricks/distributed/test.cc
```diff
@@ -0,0 +1,145 @@
...
+
+  // Merge incorrect
+  v = VectorClock<DiscreteClocks, StrictMergeStrategy>(c1, 0);
+  cur_state = v.state();
+  c2 = {0, 1};
+  EXPECT_EQ(false, v.merge(c2));
+  // internal state was not changed
+  EXPECT_EQ(v.state()[0], cur_state[0] + 1);
+  EXPECT_EQ(v.state()[1], cur_state[1]);
+}
```
dkorolev: Nit: No newline at end of file.

(Did you try `make indent` / `make fmt`, with `clang-format`?)

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
+#pragma once
+
+#include "../../bricks/time/chrono.h"
+
+class DiscreteClocks : public std::vector<uint64_t> {
```
dkorolev: FWIW, it's strictly discouraged to inherit from ~SQL~ STL containers.

The recommended solution is has-a, not is-a.

If you want it to be generic, might as well add some `: public ActsAsVector<uint64_t>`, which would contain all the necessary methods (`begin()`, `end()`, `size()`, `empty()`, and `operator[]` are enough, although you'd probably need `.resize()` and `.reserve()` too), and contain the very `vector`, so that there's a simple way to access it, perhaps even with a cast.

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+  void reset(uint32_t size) {
+    auto now = current::time::Now();
+    std::vector<std::chrono::microseconds>::resize(size, now);
+  }
+};
+
+template <class CLOCK = DiscreteClocks>
+class MergeStrategy {
+ public:
+  std::pair<bool, bool> merge(CLOCK &v1, CLOCK &v2, std::function<bool(CLOCK &v1, CLOCK &v2)> validator) {
```
dkorolev: Nit: I was a big fan of `pair<bool, bool>` until I began using more human-readable names. A two-fields `struct` with named fields is just better than `.first` / `.second`. And it's as trivial to construct as `return { lhs, rhs };` FWIW.

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+    // Basic merge strategy:
+    // 1. always merge (even in case of conflict)
+    // 2. default conflict function check that v1 <= v2
+    return std::pair<bool, bool>{!validator(v1, v2), true};
+  }
+  std::pair<bool, bool> merge(CLOCK &v1, CLOCK &v2) { return merge(v1, v2, is_conflicting); }
```
dkorolev: These `merge()` member functions should be `static`, imho, or at least `const`.

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+  static bool is_conflicting(CLOCK &v1, CLOCK &v2) {
+    // basic implementation
+    // returns true if there are no conflicts for merging
+    return !is_lte(v1, v2);
+  }
+
+  static bool is_same(CLOCK &v1, CLOCK &v2) {
+    // Happens on exactly same moment
```
dkorolev: s/on/at/

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+    for (size_t i = 0; i < v1.size(); i++) {
+      if (v1[i] != v2[i]) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  static bool is_lte(CLOCK &v1, CLOCK &v2) {
+    // Happens early or at the same time
```
dkorolev: Another nit: according to the Google Style Guide, comments are complete sentences, with periods at the end =)

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+  static bool is_conflicting(CLOCK &v1, CLOCK &v2) {
+    // Check if v1 is in sync with v2 and v1 is strictly early then v2
+    return !(!MergeStrategy<CLOCK>::is_parallel(v1, v2) && MergeStrategy<CLOCK>::is_early(v1, v2));
+  }
+  std::pair<bool, bool> merge(CLOCK &v1, CLOCK &v2) { return MergeStrategy<CLOCK>::merge(v1, v2, is_conflicting); }
+};
+
+template <class CLOCK = DiscreteClocks, template <typename> class STRATEGY = MergeStrategy>
+class VectorClock {
+ protected:
```
dkorolev, line 153: And another nit: we've converged to having plain `struct`-s with bare field names, or `class`-es, with `local_fields_` having underscores at the end (just like the G Guide suggests). So it reads odd to see a proper class but no `clock_` or `strategy_` in field naming.

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+  uint32_t local_id;
+
+ public:
+  explicit VectorClock(uint32_t size, uint32_t node_id) {
+    // Set local process id and cluster size
+    local_id = node_id;
+    clock.reset(size);
+  }
+
+  explicit VectorClock(CLOCK &v, uint32_t node_id) {
```
dkorolev: Is this `v` `const CLOCK&`?

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+    local_id = 0;
+    clock.reset(1);
+  }
+
+  void step() {
+    // T[i] = T[i] + 1 for logical step
+    clock.inc(local_id);
+  }
+
+  CLOCK &state() {
```
dkorolev: Add a `const` version too?

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,130 @@
...
+    // Returns current state for network transmission
+    return clock;
+  }
+
+  bool merge(CLOCK &to_compare) {
```
dkorolev: Not really clear from this code what's mutated and when -- I'd say either always mutate or always stay immutable, to have your users write `auto merged = merge(a, b);`, otherwise it's a potential source of bugs.

bricks/distributed/vector_clock.h
mzhurovich: @dkorolev SQL containers? 😁

dkorolev: LOL

No =)

amdrozdov:
Updated all mentioned moments, all test cases works correctly, but there is a compilation warning (for ActsAsVector):
```
vector_clock.h:62:7: warning: implicitly-declared ‘ActsAsVector<long unsigned int>& ActsAsVector<long unsigned int>::operator=(const ActsAsVector<long unsigned int>&)’
is deprecated [-Wdeprecated-copy]
```
Maybe I'm doing "naive" version of a vector template, should I change copy-constructor signature?

bricks/distributed/vector_clock.h
```diff
@@ -0,0 +1,208 @@
+/*******************************************************************************
+The MIT License (MIT)
+
+Copyright (c) 2014 Andrei Drozdov <sulverus@gmail.com>
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+                                                              copies of the Software, and to permit persons to whom the
```
mzhurovich: Oops, something went wrong here :)

bricks/distributed/vector_clock.h
```diff
@@ -3,24 +3,24 @@ The MIT License (MIT)
...
-    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-        SOFTWARE.
-            *******************************************************************************/
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+`copies of the Software, and to permit persons to whom the Software is
```
dkorolev: Nit: what's this tick quote?

bricks/distributed/vector_clock.h
```diff
@@ -34,16 +34,19 @@ class ActsAsVector {
  public:
   explicit ActsAsVector() { storage = std::vector<T>(); };
   explicit ActsAsVector(size_t count) { storage = std::vector<T>(count); };
-  explicit ActsAsVector(ActsAsVector &other) { storage = std::vector<T>(other); };
-  explicit ActsAsVector(const ActsAsVector &other) { storage = std::vector<T>(other.storage); };
+  explicit ActsAsVector(ActsAsVector& other) { storage = std::vector<T>(other); };
```
dkorolev: Nit: I believe these can be `= default;`

bricks/distributed/vector_clock.h
```diff
@@ -53,27 +56,27 @@ class ActsAsVector {
   bool empty() const { return storage.empty(); };
   size_t size() const { return storage.size(); };
   void resize(size_t count) { storage.resize(count); };
-  void resize(size_t count, const T &value) { storage.resize(count, value); };
+  void resize(size_t count, const T& value) { storage.resize(count, value); };
   void reserve(size_t new_cap) { storage.reserve(new_cap); };
-  void push_back(const T &value) { storage.push_back(value); };
-  void push_back(T &&value) { storage.push_back(value); };
+  void push_back(const T& value) { storage.push_back(value); };
+  void push_back(T&& value) { storage.push_back(value); };
```
dkorolev: Shouldn't this me `std::move(value)`?

dkorolev (approved):
LGTM!

bricks/distributed/vector_clock.h
amdrozdov: done

bricks/distributed/vector_clock.h
amdrozdov: done

bricks/distributed/vector_clock.h
amdrozdov: done

PR #999, opened 2024-04-20
Added `bool`-cast defaults for `Wait` and `WaitFor` in `WaitableAtomic`.

dkorolev:
So that if it's a `WaitableAtomic<bool>`, or `WaitableAtomic<T>` where `T` can be cast to `bool`, just `.Wait()`, or `.WaitFor(max_time_interval)` do the job.


dkorolev:
Well, I'll merge it for the action in `https://github.com/dkorolev/c5t_april2024_demo/actions` to pass =)

PR #1004, opened 2024-06-27
Added the `C5T_DLIB` symbol.

dimacurrentai:
Hi @mzhurovich, CC @yarmaksergey,

This `#define`-d symbol for `dlib` builds makes it possible to fix OS X builds.

Context: 

* There were some functions that were not supposed to be called from certain classes when those classes are used from `dlib`-based builds.
* For example, with `C5T_LOGGER`, it can only be initialized from the "main" process, not from the `dlib` itself, while the `dlib` should use the passed-in and injected instance of `C5T_LOGGER`.
* This worked on Linux so far "just" because that code path was never executed. (And executing it would have been a failure!)
* OS X is stricter, and this was resulting in linkage errors.
* With this fix, a clean solution is now possible, implemented, and tested.

I'll merge this one-liner PR in, and link to the respective fix in the other repo in the comment.

Thx,
Dima


dimacurrentai:
Ref. https://github.com/dkorolev/c5t_april2024_demo/pull/16, specifically the changes in [`lib_c5t_actor_model.h`](https://github.com/dkorolev/c5t_april2024_demo/pull/16/files#diff-ef5f9abf279babd6738ec3bd47b5f2bfe2f9a702fecd9739ed4e8b2744018d4e) and [`lib_c5t_logger.h`](https://github.com/dkorolev/c5t_april2024_demo/pull/16/files#diff-77911e25e5da40e8646efdb67a36c8b12c0ecedef90a925d29ddbd370135a205).

PR #1005, opened 2024-07-06
Supporting the `C5T_PROJECT_ROOT` symbol on Current's `CMakeLists.txt` level so that Current's `CMakeLists.txt` can be used as a nested, 2nd-level build script.

dkorolev:


dkorolev:
Proof that is does the job: https://github.com/dkorolev/cmake_in_cmake

Merging now.

